PR #13, opened 2014-12-05
Better proxy handling for Linux.

KamiIntelligence:


squarefractal:
Can this PR be closed?
(I don't really know what happened with this PR, but both participants removed their comments and there has been no follow up regarding the issue.)

wolfbeast:
Right.

PR #36, opened 2015-03-29
Create .mozconfig

### LINKED ISSUE #35: FTBFS: failure memory model cannot be stronger than success memory model for ‘__atomic_compare_exchange’
mach build failed with : 

 0:04.57 In file included from /usr/include/c++/4.9/atomic:41:0,
 0:04.57                  from ../../dist/include/mozilla/Atomics.h:176,
 0:04.57                  from /home/themusicgod1/palemoon/Pale-Moon/mfbt/tests/TestAtomics.cpp:6:
 0:04.57 /usr/include/c++/4.9/bits/atomic_base.h: In static member function ‘static bool mozilla::detail::IntrinsicMemoryOps<T, Order>::compareExchange(typename mozilla::detail::IntrinsicMemoryOps<T, Order>::Base::ValueType&, T, T) [with T = unsigned int; mozilla::MemoryOrdering Order = (mozilla::MemoryOrdering)1u; typename mozilla::detail::IntrinsicMemoryOps<T, Order>::Base::ValueType = std::atomic<unsigned int>]’:
 0:04.57 /usr/include/c++/4.9/bits/atomic_base.h:581:70: error: failure memory model cannot be stronger than success memory model for ‘__atomic_compare_exchange’
 
On ubuntu 15.04, gcc Ubuntu 4.9.2-10ubuntu12 
(it's been failing to build this way for at least a few months).

themusicgod1:
Tests failing? Screw the tests.

( resolves #35 as per @Axiomatic- 's suggestion )

### DIFF
--- a/.mozconfig
+++ b/.mozconfig
@@ -1,0 +1,4 @@
+# Do not bother to run tests, they will probably not work
+ac_add_options --disable-tests
+ac_add_options --disable-mochitests
+


wolfbeast:
Including a .mozconfig in the source dir is not a solution. .mozconfigs should be singular and not require overrides on a number of platforms, so this would be part of RelEng that needs to be done by whomever builds this.

PR #40, opened 2015-04-04
Added classic status bar

BitVapor:
I'm starting off simple with my first pull request.  One tweak, only affects OS X builds.
screen shot 2015-04-03 at 9 31 25 pm


wolfbeast:
Yup, looks good.

wolfbeast:
Tobin sent me a note about this:
> packager.mk for macos needs to be ifdef'd to MOZ_PHOENIX for statusbar inclusion
> in case someone builds something else than Pale Moon for mac..
> like FossaMail
> or xulrunner
So it needs the original line in there still, as an #else and the new in an #ifdef MOZ_PHOENIX block. See the other packager scripts for hints.

BitVapor:
Groovy.  Will investigate further.

mattatobin:
Fixed in 04b5f8c27c1e4972c87b5e9299a05593e2824e62

PR #44, opened 2015-04-08
Show quit warning when closing multiple tabs when set to "Never remember history" in prefs

trav90:
[Bug 952925] (https://bugzilla.mozilla.org/show_bug.cgi?id=952925).

Currently in Pale Moon 25.3.1, when the browser is set to "Never remember history" in the preferences, and the "Warn me when closing multiple tabs" pref is selected as well, the browser will ignore the warning, and just close with no message of any kind, causing the user to permanently lose any open tabs he/she may have had. This happens even if the correct options are set in about:config as well.

The Bugzilla page lists the platform as Mac, but I confirmed that it happened on both Windows and Linux as well. 

I also did a Linux build with the patch applied, and can confirm that it now shows a warning as intended.

### DIFF
--- a/browser/components/nsBrowserGlue.js
+++ b/browser/components/nsBrowserGlue.js
@@ -724,8 +724,6 @@ BrowserGlue.prototype = {
     if (!aQuitType)
       aQuitType = "quit";
 
-    var mostRecentBrowserWindow;
-
     // browser.warnOnQuit is a hidden global boolean to override all quit prompts
     // browser.showQuitWarning specifically covers quitting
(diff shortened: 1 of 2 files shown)


wolfbeast:
One problem: it adds a string. This won't work when language packs are installed, as a result.
I'm really hoping I can get a working l10n setup before the next release so i can stop having headaches about this :/

wolfbeast:
I'll just have to mangle l10n by somehow merging incomplete string in .properties files. I'll give that priority -- this bug needs to be fixed anyway.

PR #48, opened 2015-04-13
[Bug 680547] Allow Linux 64-bit to compile with -march=native.

Axiomatic-:
This only fixes one issue of compiling Pale Moon with -mavx, (this fixes the segfault) the another issue in which 'make package' seems to fail with -mavx (however it seems to be by, "random chance" for me at least). 

@trav90 if you have a AVX capable processor perhaps you can test if you can reproduce the same issue?

[Direct link to bug 680547.](https://bugzilla.mozilla.org/show_bug.cgi?id=680547)


trav90:
Sorry for my terribly late response to this.

Unfortunately, I do not have an AVX capable processor to be able to test.

PR #46, opened 2015-04-10
Make Ciphersuite and TLS Protocol visible

trav90:
User request: https://forum.palemoon.org/viewtopic.php?f=13&t=7843

Bug numbers:  [886752] (https://bugzilla.mozilla.org/show_bug.cgi?id=886752) & [947149] (https://bugzilla.mozilla.org/show_bug.cgi?id=947149)

PR #49, opened 2015-04-15
Add more SSL/TLS connection information to "Page Info"

trav90:
Original pull request: https://github.com/MoonchildProductions/Pale-Moon/pull/46

Forum thread: https://forum.palemoon.org/viewtopic.php?f=13&t=7843

Bug [886752] (https://bugzilla.mozilla.org/show_bug.cgi?id=886752)  

Pale Moon now displays the version of TLS in use in "Page Info". It still doesn't show as much info as Firefox does, but I'm going to attempt to incorporate some more of the patch from bugzilla. But for now, confirmed to show TLS version in x64 Linux trunk build. Note that the Cipherfox add-on (as mentioned in forum thread) still does not display any additional information.

Also note that unlike the first pull request, this one does not include the patch from bug 947149 (I will submit a new request for that later if this code looks good).

Screenshots:

Pale Moon 25.3.1:
pm25 3 1

Pale Moon Trunk with patch:
pmtrunk

Firefox 37.0.1:
ff37 0 1

### DIFF
--- a/browser/base/content/pageinfo/security.js
+++ b/browser/base/content/pageinfo/security.js
@@ -52,17 +52,35 @@ var security = {
         cAName : issuerName,
         encryptionAlgorithm : undefined,
         encryptionStrength : undefined,
+        version: undefined,
         isBroken : isBroken,
         isEV : isEV,
         cert : cert,
         fullLocation : gWindow.location
       };
 
+      var version;
       try {
         retval.encryptionAlgorithm = status.cipherName;
         retval.encryptionStrength = status.secretKeyLength;
+        version = status.protocolVersion;
       }
       catch (e) {
+      }
+
+      switch (version) {
+        case nsISSLStatus.SSL_VERSION_3:
+          retval.version = "SSL 3";
+          break;
+        case nsISSLStatus.TLS_VERSION_1:
+          retval.version = "TLS 1.0";
+          break;
+        case nsISSLStatus.TLS_VERSION_1_1:
+          retval.version = "TLS 1.1";
+          break;
+        case nsISSLStatus.TLS_VERSION_1_2:
+          retval.version = "TLS 1.2"
+          break;
       }
 
       return retval;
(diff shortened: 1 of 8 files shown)


wolfbeast:
It doesn't show the full cipher suite because I use the (originally intended) return of the symmetric cipher name, not the full suite, because that just looks damn ugly, IMHO, and shouldn't be presented to the user. See line 977 in nsNSSCallBacks.cpp where I commented a line that would return the full suite string instead.

I assume cipherfox parses this string.

wolfbeast:
Code looks good at first glance, I'll merge and do a win build with it.

trav90:
I agree, I think the full suite looks ugly as well and definitely prefer the looks of the PM Trunk over FF.

If all goes well with the Windows build, I'll look into 947149.

wolfbeast:
Windows build looks a-OK, all green.

trav90:
Great!

wolfbeast:
Thanks for taking the time to work through this :)

wolfbeast:
*If all goes well with the Windows build, I'll look into 947149.*
I think removing that distinction is bad. why not actually use strong/weak to indicate <128 bits (3DES) and RC4 since it's already in the code? I'll be poking at that.

wolfbeast:
Added Commit: 7ca6160abeabe9c1f6182b30de2aa62b589859ed to classify strong/weak encryptions - let me know what you think. if you have a better idea, I'm open for suggestions!

trav90:
*I think removing that distinction is bad.*
I had some time today and was looking more in depth at the mentioned bug, and I agree, removing the distinction is a bad idea.
The commit looks good to me! I was actually going to work on something along those lines today, but you beat me to it!

wolfbeast:
Currently in the pipe (not committed yet) is the folowup to that.
rockstarlowgrade1

trav90:
I like it! That is exactly the information I was going to have displayed in my patch (that I hadn't done yet :P).

PR #51, opened 2015-04-19
Bug 896287 - NVIDIA specific GLX server vendor check for pixmap depth and fbConfig depth

squarefractal:
https://bugzilla.mozilla.org/show_bug.cgi?id=896287

### DIFF
--- a/gfx/gl/GLContextProviderGLX.cpp
+++ b/gfx/gl/GLContextProviderGLX.cpp
@@ -260,6 +260,7 @@ GLXLibrary::EnsureInitialized(LibType libType)
     }
 
     mIsATI = serverVendor && DoesStringMatch(serverVendor, "ATI");
+    mIsNVIDIA = serverVendor && DoesStringMatch(serverVendor, "NVIDIA Corporation");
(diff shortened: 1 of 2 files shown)


wolfbeast:
Background: fbConfigs don't really have depths - they have a buffer_size.  From "GLXPixmaps are supported for visuals whose GLX_BUFFER_SIZE
is one of the pixmap depths supported by the X server" [2], the pixmap depths would be expected to match the buffer_size, but NVIDIA drivers do not provide any buffer_size = 24 fbconfigs for use with pixmaps without alpha bits (bug 779786 comment 3).

Visuals have a depth, but visuals don't usually include alpha bits in the depth.
Only the visual(s) from the Composite extension includes the alpha bits in the depth.

I did some debugging with gdb, and the crash is happening with the pixmap format->depth != depth for the visual. It happens for format->depth = 32 and depth = 24 in GLXLibrary::CreatePixmap(gfxASurface*). It looks like this check: if (depth != format->depth && depth != format->depth - alphaSize) doesn't work anymore with nvidia. I'm going to try dropping the 2nd test and seeing if the crash goes away.

PR #57, opened 2015-04-26
first draft of a Mac theme mimicking the Windows theme

BitVapor:
This is a first draft of the Pale Moon Windows theme for Mac which resembles the Windows theme.  There are a few known outstanding issues:

* no hover-over highlight for bookmarks toolbar
* 5 pixel white border on left and right side of content window
* full-screen: newly integrated fixed status bar doesn't disappear and switch to new style a la Windows
* full-screen: hovering to discover the menu also reveals an additional bar displaying the Mac close/minimize/maximize buttons.  Note that this is a Yosemite regression as it does not occur in my Mavericks build VM


wolfbeast:
Note that the fixed status bar is not visible in full-screen by default on purpose. This is a status bar options setting, and works the same on Windows (it will change to pop-up).

Otherwise thanks for your work so far! I wish I could help you with this but I don't have the required hardware.
One tip you can use is to use the unpackaged build, manually copy over the status bar code, and start the browser with --purge-caches (and set up in a way similar to what you'd need for add-on development so you never cache XUL etc.) You can then tinker with the theme components and run immediately with the changes.

PR #59, opened 2015-04-27
Remove Mozilla specific parameters and pretend to be Firefox for suggestions.

squarefractal:
The search engine XML for google.com contains Mozilla specific parameters that do not make any sense for PM.

However, the suggestions can only be retrieved if `client=firefox` and `output=firefox` are set. The domain has switched to `https://suggestqueries.google.com/`

This commit fixes the two issues.


wolfbeast:
This is actually not a simple decision for "making it work". Previously, with the search parameters in place combined with Mozilla's search contract with Google, any such queries would fatten Mozilla's pockets if they would lead to a sponsored search result.

Considering Mozilla's end-of-contract with Google in that respect, this is likely acceptable, however, can you verify that the result list and suggestion is NOT including any sponsored search results? If it is, then I cannot accept this patch.

squarefractal:
>any such queries would fatten Mozilla's pockets

Note that:
- The actual search (as would be done with my patch) does not have any parameters identifying itself as Firefox. So, a correlation between the search cannot be necessarily drawn.
- There is not much money to be made in a search suggestion as compared to an actual search, since only the latter can only be used as a revenue stream for Google (a share of which is given to Mozilla).

That being said, I tried a number of search terms, but did not find any Mozilla-sponsored results/ads.

wolfbeast:
It's not that simple - but in the end it doesn't matter as long as the search suggestions themselves do not include any sponsored results. There is a reason why it only works when you pretend to be Firefox, while for a search engine, it would not matter what the UA is of the browser used for searches. Think about that a moment.

PR #61, opened 2015-04-28
HSTS update

squarefractal:
Update the HSTS lists.

<p color="#acacac"><i>Aside: I'm a little disappointed that you aren't pulling this in, yourself.</i></p>


wolfbeast:
> Aside: I'm a little disappointed that you aren't pulling this in, yourself.

Really? Do a history search, man. I do this regularly!

wolfbeast:
Aside: Since you've made a point of adding a *lot* of entries to this list, please be aware that this is a *preload* list, and will override any actual HSTS checking. No entry = actually check HSTS on the server, so having this list in the first place is optional and a smaller list means fewer assumptions about the state of the web.
More entries means it requires more religious maintenance - are you going to keep an eye on these and submit pulls as-needed from this point forward?

squarefractal:
>are you going to keep an eye on these and submit pulls as-needed from this point forward?

Why not :)

wolfbeast:
Awesome!

wolfbeast:
Tag #62

PR #64, opened 2015-05-01
Linux: Use only GIO and remove GnomeVFS

trav90:
Basically a code cleanup, removing dead code (GnomeVFS) that is not used unless specifically enabled in mozconfig and also has been deprecated since April of 2008.

Bugs with relevant info/patches: [713802] (https://bugzilla.mozilla.org/show_bug.cgi?id=713802), [694870] (https://bugzilla.mozilla.org/show_bug.cgi?id=694870) and [1134537] (https://bugzilla.mozilla.org/show_bug.cgi?id=1134537).

Trunk with these commits applied builds successfully on CentOS with GCC 4.7.2.


wolfbeast:
Thanks Travis!

trav90:
You're welcome! :)

PR #65, opened 2015-05-01
Continuous integration

squarefractal:
Pale Moon currently does not have any system to check whether the commits/PRs made result in a "build-able" and stable trunk. As a result, any and all commits/PRs have to be checked manually.

This PR adds continuous integration support using Travis-CI. At present, it only solves the former problem ("build-able" state). I plan to look into downloadable "try-builds" (like Mozilla does) at some later point in time.

The build infrastructure uses:
- Ubuntu 12.04
- Clang instead of gcc-4.7 due to memory usage issues.

  I tried adding swap area to get around this problem, but (as I learned later) that does not work in OpenVZ containers.

The current build status may be reviewed here:
https://travis-ci.org/squarefractal/Pale-Moon

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,6 @@ sudo: required
 
 before_install:
   - bash build/travis_ci/travis_ci.sh deps
+  - bash build/travis_ci/travis_ci.sh swap
 script:
   - bash build/travis_ci/travis_ci.sh build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: cpp
+sudo: required
+
+before_install:
+  - bash build/travis_ci/travis_ci.sh deps
+script:
+  - bash build/travis_ci/travis_ci.sh build
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,5 @@ sudo: required
 
 before_install:
   - bash build/travis_ci/travis_ci.sh deps
-  - bash build/travis_ci/travis_ci.sh swap
 script:
   - bash build/travis_ci/travis_ci.sh build
--- a/build/travis_ci/travis_ci.sh
+++ b/build/travis_ci/travis_ci.sh
@@ -8,16 +8,15 @@ logfile="$srcdir/travis.log"
 install_deps () {
 	set -e
 
-	sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
 	sudo apt-get update -y --force-yes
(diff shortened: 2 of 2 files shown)


wolfbeast:
Sorry, but suddenly popping in unannounced and without discussing anything about integration of the Pale Moon source tree with a service I never even heard of is not something I can accept.

squarefractal:
>integration of the Pale Moon source tree with a service I never even heard of

Travis-CI is a popular service used across many Github repos. If you've ever seen those build|{failing/passing} icons on the readmes of a particular project, yes, Travis-CI is at work there.

Description of the patches has been given above and I think that's clear enough. If there's something I haven't clarified, feel free to ask :)

In addition, I got a successful build with Clang. See here:
https://travis-ci.org/squarefractal/Pale-Moon/builds/60852190

wolfbeast:
... Basic information about Travic-CI (which obviously is NOT well-known enough for me to know about) would have been nice. Please understand that just springing this on me without even pinging me beforehand is something that is an automatic refusal because it integrates with a third party without me knowing any details. Please try to understand the principle.

No, I've never seen any sort of "build failing/passing" indicators either. Nor do I think it would be useful to display something like that - not every commit is going to build on all platforms (and you're only testing one, here, anyway) because of the way I implement things, and as such having a build system try to is going to result in a lot of unnecessary red flags.

In addition, clang is not what's going to be used for production builds so "it builds with clang" is not going to be helpful at all, sorry.

squarefractal:
>Nor do I think it would be useful to display something like that

That is only an image that Travis-CI chooses to display, but the reports and the logs are useful. Of course for people who don't need to work with the code, such indicators are useless.

>not every commit is going to build on all platforms because of the way I implement things

For commits that you don't want should go through the process: just write `[ci skip]` or `[skip ci]` somewhere in the commit message!

Also, I think you are missing the point: say you have a component that you wish to remove, which has three subcomponents. Would you prefer to build it yourself three times for removing each of the components, or would you prefer to get some automatic report on whether it built or not?

>you're only testing one, here
Is it not better to have a partial implementation than no implementation at all?

>clang is not what's going to be used for production builds
gcc wastes a lot of memory and I've tried what I could (changing versions, using a swap area) but they have not worked. Also see above point.

wolfbeast:
You're missing the point, yourself.

1) Our main OS we build for is Windows. I doubt that's available (and certainly not for free).
2) On Linux, we use GCC, not clang. I already said this. That is what we use, and if you can't even get it to work then it's totally useless even for Linux. GCC does not "waste" a lot of memory. MSVC uses just as much memory to build Pale Moon - it is just how much memory is needed to compile 14 million lines of code. if clang uses significantly less then clang must be cutting corners.
3) Your "successful build" was completed in 15 minutes, which is impossible for a proper build - looks like a fudged incremental build. Incremental builds don't necessarily mean that the tree builds.
4) I am not going to dirty my tree with [ci skip] messages in everything but tagged betas.
5) it's better to have no implementation at all of something useless than to feign something useful.
6) We have our builds under control - the human factor is *important* for Pale Moon. We do not "ship when our automated tests say the tree is green".

PR #68, opened 2015-05-02
Mac theme fixes and cleanup

BitVapor:
This is a substantial cleanup of the Mac theme under development.  I removed all remaining Windows/Aero files and references within, enabled bookmark toolbar/library highlighting, fixed many missing images, and removed the 5 pixel border on the left and right side of the screen.


wolfbeast:
Thanks, BitVapor!
Since I'm unable to see the result, I'll just have to assume at this point that it all looks a-OK ;)

Issue #62, opened 2015-04-28
Keep a running update of HSTS preload lists


wolfbeast:
Pale Moon should keep updated with somewhat recent HSTS preload lists.

@squarefractal has offered to do this.

(responsibility tracking issue, leave open)

squarefractal:
Documentation:

HSTS preload lists have moved from `security/manager/boot/src` to `security/manager/ssl`:
Reference: https://bugzilla.mozilla.org/show_bug.cgi?id=1164714

wolfbeast:
I don't really care where Mozilla stores their HSTS preload lists in the tree ;)

squarefractal:
That was for documentation purposes, mostly for personal use.

BTW, is there some reason git is apparently doing the merge two times? It seems very odd and pollutes the commit history.

wolfbeast:
It happens when your PR merge and someone else's edits mid-air collide. e.g. I make a commit while you make a PR. I try to push the commit but it collides, so the branch on GitHub has to be pulled down before I can push. Anyone's local repo copy is an inherent separate branch from the remote that has to be re-merged if they are out of sync.

TPS:
#61, #69, #74, #82, #85, &c (just to pick 5 that auto-completed :wink:) were all not mentioned here....  If this issue isn't tracking, is there a reason to keep it open?

wolfbeast:
It's an on-going task.

wolfbeast:
Looks like we need a new person to pick this up and provide updates.

TPS:
> Looks like we need a new person to pick this up and provide updates.

@wolfbeast I suppose https://github.com/squarefractal/encryptedweb/issues/6#issuecomment-239602542 is a related problem. 😭

wolfbeast:
I don't mind if people get busy with other stuff but they should not let us hanging in limbo.
Communication is key.

trav90:
I looked at this tonight, and Mozilla has changed the format of the file containing the HSTS preload list (actually made it more efficient). But as a result we can no longer simply pull the updated files from Mozilla as-is (i.e., we'll need to go though the files and make any relevant changes manually).

This will not be an issue with Tycho (and I'll volunteer to keep this updated since no one else has), but it's going to take a _lot_ of time to update this for Pale Moon -release.

wolfbeast:
[...]

PR #69, opened 2015-05-02
HSTS preload update

squarefractal:
Just one of these before the 25.4.* release.


wolfbeast:
horseboners.xxx :D

wolfbeast:
Tag #62

PR #71, opened 2015-05-05
Removal of Mozilla-specific search parameters, part 2

squarefractal:
This PR removes Mozilla-specific search parameters from other search engines.

- _With this, will you add the other search engines to the browser via addition to list.txt?_
- _Is there a reason to notify DDG about the client name?_


wolfbeast:
Did you research requirements of search plugins for the different services? Do you know if these parameters are optional and if not, what is required to be passed? Etc.

squarefractal:
- Removal of `<MozParam>` elements should be self-explanatory. Some of them read from the prefs and pass them as part of the search query.
- answers.xml: `gwp=13`, *.xml: `sourceid=desktop-search` produces the same results with and without them.

wolfbeast:
Let me rephrase.
I know how the Mozilla code reads and passes the parameters to the search engine. That's not what I asked.

What do these parameters do in the search plugins? Have you researched this? e.g.: what does gwp=13 stand for in the answers plugin? what is its effect? what does passing MOZI to Bing,do, do we need a replacement or can it just be omitted? etc. etc. for the other search plugins.

squarefractal:
**TL;DR**: The removed parameters do not seem to have any effect on the searches.

Seems like you're looking for hard documentation, so here you go:

- [309379 – update answers.com plug-in](https://bugzilla.mozilla.org/show_bug.cgi?id=309379):

  >update the answers.com plugin to include tracking code

 Obviously not needed.

- [671307 – Add Twitter to the default search plugins (en-US only)](https://bugzilla.mozilla.org/show_bug.cgi?id=671307)

  > Updated twitter search plugin using https and corrected URL and params, per Twitter.

  The `desktop-search` param was added at the request of Twitter. I've tried making searches without this param, and I didn't see any effect, so I assume this is something Twitter wants for metrics.

- [Yahoo search not using custom search codes](https://bugzilla.mozilla.org/show_bug.cgi?id=942024)

  This one is not directly related, but will confirm the fact that such "search codes" are used for Bing, Yahoo and Google (already removed from PM with earlier PR).


- For the others, I haven't seen any effect by removing said parameters.

wolfbeast:
It's not so much "hard evidence" but rather going from the assumption that all parameters are there for a reason, and just blindly removing things without knowing this reason is not a good thing, Especially when dealing with (1) essential functionality that many users use and (2) third parties.

PR #72, opened 2015-05-06
[!] Minor bug fix.

Axiomatic-:


wolfbeast:
Nice catch for the few who actually use the js shell ;)

PR #75, opened 2015-05-10
Bug 1062355 - Crash in d3dcompiler, nested-functions-should-not-crash…

InternalError503:
….html test of WebGL Conformance Test

Crash during WebGL conformance test


wolfbeast:
Had to back this one out because it would prevent any shader compilation from occurring properly, resulting in general breakage of WebGL rendering.

Issue #16, opened 2015-01-30
Pale Moon fails to build with Clang.


Axiomatic-:
A possible bug, I have not had time to research into it, but it seems to be related to libpvx. (Tested on Linux, can anyone else confirm?) 

Can someone also please check, and see if Clang compiles Pale Moon on Windows?

wolfbeast:
Building with clang is completely untested (it's never used for production) It's very likely that clang on windows also fails on libvpx -- libvpx is a tricky library because of some very very odd ways of going about compiling it.

Self-Referential-Drama:
I see, well I am going to go ahead, and try to get Clang to work for compiling Pale Moon if that is alright with you.

wolfbeast:
Absolutely!

Axiomatic-:
Progress has been made, the issue is because libvpx uses in-line assembly code causing Clang to fail, using a compiler flag as a work around should fix that problem. (See commit: 5be82262925aac103cffd44837b01482b488c98e)

All that needs to be done is test if Clang fully builds Pale Moon, which I am planning to do soon. 

Upstream bug was labelled as WontFix. (https://code.google.com/p/webm/issues/detail?id=729)

Axiomatic-:
Clang version 3.5.1 fails to build Pale Moon, due to an upstream [bug](http://llvm.org/bugs/show_bug.cgi?id=20879).

I have yet to test Clang 3.4, or the SVN working commit, I will try to test either one later today.

wolfbeast:
Thanks for the updates!
Is commit you mentioned a working workaround, and only failing because of a different upstream bug, or is it actually not a working solution (in which case it would need something else)?

Axiomatic-:
No problem, the commit is a working solution for the libvpx inline assembly code issue with Clang, the current issue with Pale Moon not building with Clang is related to mozjemalloc. (However it is a Clang upstream bug as it seems.)

```
 3:43.87 jemalloc.c
 3:46.31 fatal error: error in backend: Cannot select: 0x2b8a4a0: i64 = X86ISD::WrapperRIP 0x2af2918 [ORD=12] [ID=12]
 3:46.31   0x2af2918: i64 = TargetGlobalTLSAddress<i1* @mmap_unaligned_tls> 0 [TF=10] [ORD=12] [ID=9]
 3:46.31 In function: chunk_alloc
 3:46.32 clang: error: clang frontend command failed with exit code 70 (use -v to see invocation)
[...]

PR #78, opened 2015-05-14
[!] Fix support for builds using newer freetype.

Axiomatic-:
This commit fixes the undefined reference for me in issue #16, caused by a change by the freetype upstream. (I did not remove the older references so this should still allows for building with older freetype packages as well.)

### DIFF
--- a/config/system-headers
+++ b/config/system-headers
@@ -411,6 +411,7 @@ freetype/ftlcdfil.h
 freetype/ftsizes.h
 freetype/ftadvanc.h
 freetype/ftbitmap.h
+freetype/ftfntfmt.h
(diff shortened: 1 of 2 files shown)


wolfbeast:
I never really looked into messing with the system header defs. Is this OK to do, i.e. won't it complain if it can't find the stated header files?

Axiomatic-:
It is okay (at least from my point of testing), I was able to compile the latest git build at the time of this pull request which included the older system header (ftxf86.h); everything compiled, and functioned as wanted.

wolfbeast:
All right - I'm sure it'll come out in the wash otherwise :)
Thanks!

PR #79, opened 2015-05-14
Miscellaneous fixes for issues found while building for Alpine Linux

idunham:
Here are a couple small fixes for some build breaks I stumbled across while building for Alpine Linux.

1- commit 1f31aa51, fix --with-system-pixman builds: This is a non-platform-specific bug that I happen to have encountered. By default the pixman-1 include directory is not added to ${INCLUDES}.
I'm not sure how this would affect Windows builds, though.

2- commit c065b28caf, make execinfo.h glibc-specific.
There are numerous Linux libc versions; only glibc and some builds of uclibc (which also defines __GLIBC__) provide execinfo.h. Android's bionic libc, musl, and a number of more obscure versions do not.
So far, the only Linux systems Pale Moon is officially built for are glibc-based distros and Android, but Alpine Linux uses musl.

I'm using the system library versions for most packages for several reasons; the main ones are that some packages have only recently been updated to compile in the build environment Alpine has, and that I'm trying a build for the first time on an Atom N270 netbook with 1 gig of ram.
I have not gotten a complete build yet, but these will probably be needed for it.

### DIFF
--- a/config/config.mk
+++ b/config/config.mk
@@ -380,6 +380,10 @@ JAVA_IFACES_PKG_NAME = org/mozilla/interfaces
 
 OS_INCLUDES += $(MOZ_JPEG_CFLAGS) $(MOZ_PNG_CFLAGS) $(MOZ_ZLIB_CFLAGS)
 
+ifndef MOZ_TREE_PIXMAN
+OS_INCLUDES += $(pkg-config --cflags pixman-1)
+endif
+
 # NSPR_CFLAGS and NSS_CFLAGS must appear ahead of OS_INCLUDES to avoid Linux
 # builds wrongly picking up system NSPR/NSS header files.
 INCLUDES = \
--- a/ipc/chromium/src/base/debug_util_posix.cc
+++ b/ipc/chromium/src/base/debug_util_posix.cc
@@ -5,7 +5,7 @@
 #include "build/build_config.h"
 #include "base/debug_util.h"
 
-#define MOZ_HAVE_EXECINFO_H (defined(OS_LINUX) && !defined(ANDROID))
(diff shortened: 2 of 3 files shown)


wolfbeast:
Thanks for your analysis and patches! I'll merge them in as they seem safe, and you can try a build against the current master.

idunham:
I've tried that, and it wasn't quite enough:
* where I wrote
OS_INCLUDES += $(pkgconfig ...)
it should have been $(shell pkgconfig ... || true)
* I needed to add -lfts to OS_LDFLAGS, but couldn't get that done properly (configure died if I ran it with LDFLAGS=-lfts).
Context: fts is part of all the BSD libcs and glibc, but neither musl nor bionic have it. It is possible to build it separate from libc, which Alpine Linux does--packaging the result in "fts" and "fts-dev"
I ended up hacking a workaround into config/config.mk, which is probably the wrong thing to do.
I suppose the right answers are to either add fts to the autoconf tests, or to define HAVE_FTS on glibc/bsd systems, and change the "ifndef ANDROID"  that makes fts conditional to "ifndef HAVE_FTS".
* Some remaining .py and .sh files needed to be chmod +x; I just used find -name -exec.

The result built and ran, but has a bug that I've encountered in Firefox on the same install: no text is visible, whether UI or text.
This bug seems to only affect xul/gecko based software: gtk2 and gtk3 programs work fine.
It does not seem to be a matter of fonts, as I was able to bind-mount /usr/share from a working Debian system onto the affected system without changing the results; it also does not seem to be X server-side, since I could not reproduce it when chrooting into debian and running iceweasel.
So I'm still wondering what else could be up.

wolfbeast:
I think the smart thing to do here is to create a specific dev branch for Alpine to try and work these things out. Maybe poke the Linux folks on the forum as well and see if there's anything specific to try.

PR #81, opened 2015-05-14
navigator.geolocation should never be null

trav90:
A patch for the issue described here: https://forum.palemoon.org/viewtopic.php?f=29&t=8219

Tested using the Live Result box on the geolocation MDM page here: https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/Using_geolocation

With geo.enabled set to true, test returns correct location.
When geo.enabled set to false, I receive a "Unable to retrieve your location" message (25.4.1 simply displays a never ending "loading..." message.

### DIFF
--- a/dom/base/Navigator.cpp
+++ b/dom/base/Navigator.cpp
@@ -1067,12 +1067,8 @@ NS_IMETHODIMP Navigator::GetGeolocation(nsIDOMGeoGeolocation** _retval)
     return NS_ERROR_FAILURE;
   }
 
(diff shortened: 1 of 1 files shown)


wolfbeast:
I previously aligned this with the spec in Nov 2014 according to data I had at the time. I guess the spec changed out from under me?

trav90:
I believe that it was a bug in the Mozilla implementation, not a change in spec (see top comment on above MDN page).

wolfbeast:
Fantastic. Thank you Mozilla. </sarcasm>

PR #83, opened 2015-05-22
Partial fix for devtools UA sniffing insanity in the browser.

squarefractal:
Devtools uses UA sniffing for some of its functions, which (as expected) breaks when one changes the user agent to something else.

This PR provides a partial fix for this.

A more serious approach towards this (involving nsIXULRuntime) has to be done, but I don't have a good command of Mozilla XUL development. You may reject this PR if you're planning on improving above discussed aspect.

### DIFF
--- a/browser/devtools/sourceeditor/orion/orion.js
+++ b/browser/devtools/sourceeditor/orion/orion.js
@@ -3420,9 +3420,6 @@ define("orion/textview/textView", ['orion/textview/textModel', 'orion/textview/k
 	}
 	var userAgent = navigator.userAgent;
 	var isIE;
-	if (document.selection && window.ActiveXObject && /MSIE/.test(userAgent)) {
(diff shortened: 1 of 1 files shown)


wolfbeast:
This is just exchanging one evil for another.

squarefractal:
The real question is: are there any plans to improve this?

wolfbeast:
Yes, there are.
The thing is, if you're using a global UA override that identifies Pale Moon as something not even closely related to the product... then what do you expect?

squarefractal:
Assuming that the last question was not rhetorical: internal functions should not break just because the UA is faked. I often change the UA value with an extension for some quick website testing and for occasional reverse engineering purposes.

wolfbeast:
I guess hard-coding is a band-aid until someone gets time to yank it all out properly...

PR #84, opened 2015-05-22
Update NSS to 3.19

squarefractal:
Just what it says.

Known to successfully build: https://travis-ci.org/squarefractal/Pale-Moon/builds/63672592


wolfbeast:
Before I merge this upstream pull:
- Have you taken into account the recent changes I made, e.g. to patch NSPR for IPv6 [e3afb8a] and prevent DNS ANY requests, etc.?
- Have you actually run and tested the resulting browser to see if secure connections and page information/certificates still work as they should?

mattatobin:
What are the over all benefits to the newest nspr besides it being newer? Our version has specific modifications.

mattatobin:
After going through this.. I don't think we should wholesale port from upstream. There is very little of functionality changes I could see that could not be specifically targeted and ported on a case-by-case basis from the upstream repo and it would (without additional work) obliterate our specific additions.

This upstream wholesale update would also deposit a load of junk in the form of "google tests".

My recommendation is to reject this pull request and select specific enhancements to our nspr.

"Because its newer and because it builds" is not a valid reason to do anything unless it can be specifically explained why we should update a component and what tangible benefit it has.

These kinds of wholesale updates must be evaluated VERY carefully when done as replacing massive chunks or whole components of the codebase could overwrite potentially years of work for something that could be inferior and have unforeseen repercussions to the sustainability of the codebase.

wolfbeast:
I've pulled down the patch file for this, and re-worked the code changes to re-implement PM-specific changes which I'll land tomorrow. This pull request as-is is not acceptable, but no further work should be done on this by anyone but me.

mattatobin:
In the future it may help to save time and effort to get the go ahead from Moonchild before putting work into something. He knows best what the current state score is.

squarefractal:
>In this case it makes further improvements possible like key pinning so is generally desirable, or I wouldn't have considered the update.

I don't understand, why are you opposed to updating something as important as crypto libs?

wolfbeast:
Excuse me? The crypto libs have been updated a few times and evaluations have been made of further point updates to 3.16.2.3, for example, that merely catered to SCSV which is a pointless exercise to begin with (see removal of insecure fallback @ Mozilla). 

You **MUST** understand that updating crypto libs cannot usually be done just "because they are updated" because it can and often will pull a working core set of functions out from under the browser. This is not resistance to updating crypto libs, it is resistance to pointlessly risking breakage and lack of research. It's important to know how the libs are called before you change them out from under an application.
Try updating openssl 0.9.8 to 1.0.1, for example, in existing programs. Good luck trying to sort that one out.
For example, I still need to investigate how cert signatures have changed and see if I need to update the PSM module accordingly. That is research that should be done *before* an update like this is pulled.

This is called knowing your code sensitivity. If you don't understand that then you're actually shaking **my** belief in your capabilities to contribute meaningfully. You obviously did not check out previous patches done to the same libs resulting in overwriting previous work with your pull request. You didn't even check back a few days to a recent NSS patch for Logjam.
If you think just blindly pulling upstream updates will "just work", then you've got another thing coming. You've been lucky in this case because the updated libs in question don't actually change any exposed functions/symbols that are in use by the browser, but even so, your patch was incomplete and would undo previous work. Having to go over your commits with a fine comb to double-check if you've done things right is not my idea of development.

wolfbeast:
Now look what you did, you made me look up the exact versions ;)

Firefox 24: NSS 3.15.1 crypto and (by the way that's not a crypto lib) NSPR 4.10.0
If I hadn't updated NSS, TLS 1.2 wouldn't even have been possible, either.

squarefractal:
>You obviously did not check out previous patches done to the same libs resulting in overwriting previous work with your pull request. You didn't even check back a few days to a recent NSS patch for Logjam.

I actually assumed they were backported upstream patches :P
But point taken, will be careful about other PRs.

Issue #94, opened 2015-06-09
Screensaver should be disabled when playing HTML5 video


wolfbeast:
Fairly simple: add a WakeLockListener to the <video> element.

Bugs that need porting:
https://bugzilla.mozilla.org/show_bug.cgi?id=968603 (add WakeLockListener)
https://bugzilla.mozilla.org/show_bug.cgi?id=1063995 (ignore HTML5 audio)
and perhaps for the nonstandard case of loading audio in a video element:
https://bugzilla.mozilla.org/show_bug.cgi?id=1022669

Feel free to take this bug and work on it.

trav90:
I volunteer to take this.

wolfbeast:
Travis' PR should fix this on the OSes where this matters most.

wolfbeast:
Re-landed with fixes as commit 085e1a6322d6cc5890078a6c3b296bc171b12ff2
Travis: the culprit was a missing MOZ_FINAL on the class; without it, the class is made abstract for inheriting it to child classes. While there I removed some double declarations of pointers and made them a single nsCOMPtr for the power management interface.

trav90:
OK, I see. Thanks for the info. Glad it's working!

trav90:
Should the fixes for Windows be included in Mac also?

wolfbeast:
I haven't looked at the Mac code. If it's stated similarly there then you probably do need to make the same changes.

wolfbeast:
Reopening this since it should probably be verified building on all platforms and verified working as intended ;)

trav90:
PR #98 Please double check make sure I didn't miss any changes.

trav90:
PR #101 Linux patch.

wolfbeast:
Now we need to build and test that it actually works ;)

trav90:
Sorry, should have mentioned that I tested it before submitting the PR (for Linux anyway). It builds successfully and works with the proper services enabled in CentOS. Icouldn't get it to work on my Manjaro box, but I believe it is related to the newer package versions Manjaro uses (see comments on the PR). I don't have a way to test on Mac. @BitVapor, could you take a look?

trav90:
The Linux patch is apparently still not fully complete per issue #109

wolfbeast:
We'll have some time to sort this out; if not for 25.6, then I'll back it out for the release.

squarefractal:
Will this backout be on the relbranch or master?

[...]

PR #95, opened 2015-06-10
Disable screensaver when playing HTML5 video

trav90:
This PR should resolve issue #94 on Windows & Mac. Unfortunately I am unable to test on either OS, I can only confirm that it doesn't break building on Linux.

Speaking of Linux, applying the Linux-specific patch DOES cause build bustage, so it was backed out for now and I will investigate further (although numerous comments on BMO seem to suggest that it only works on certain distros/package versions anyway so even if it does get applied successfully it still may not even work).


wolfbeast:
Thanks for your work on this, Travis!
What's the current state of this PR -- should it be OK to merge this in, in its current state, or is it still a WIP?

trav90:
Glad to do it!

The Linux commit causing the issues is reverted out. Should be good for Windows & Mac in it's current state.

wolfbeast:
I've had to back this out due to build failures on Windows.
Please check the PR for the log.

PR #99, opened 2015-06-13
Fix DNS prefetch.

Axiomatic-:
[!]  Fix DNS prefetch to skip requests that point to local resources.


wolfbeast:
Thanks - looks good, although DNS prefetching is not enabled by default in Pale Moon so it'd only impact people who explicitly enable it.

PR #101, opened 2015-06-14
Disable screensaver when playing HTML5 video on Linux

trav90:
Working PR for Linux to resolve issue #94.

Note that this patch uses the org.freedesktop.ScreenSaver and/or org.gnome.SessionManager D-BUS interfaces so it depends on such services being enabled to work.


trav90:
Also note that there are reports that it doesn't work on certain package combinations/versions. There is a followup bug created for this, but as of yet has been unresolved. I'll keep an eye on it.

wolfbeast:
If we can get the basics in this way then that's definitely good.

trav90:
Agreed.

PR #104, opened 2015-06-16
Display number of matches in Find bar

trav90:
Screenshots:

Pale Moon 25.5.0:
pm25 5 0-findbar

Pale Moon Trunk with patch:
pm-trunk-findbar

### DIFF
--- a/modules/libpref/src/init/all.js
+++ b/modules/libpref/src/init/all.js
@@ -391,6 +391,7 @@ pref("accessibility.typeaheadfind.enabletimeout", true);
 pref("accessibility.typeaheadfind.soundURL", "beep");
 pref("accessibility.typeaheadfind.enablesound", true);
 pref("accessibility.typeaheadfind.prefillwithselection", true);
+pref("accessibility.typeaheadfind.matchCountTimeout", 250);
 
 // use Mac OS X Appearance panel text smoothing setting when rendering text, disabled by default
 pref("gfx.use_text_smoothing_setting", false);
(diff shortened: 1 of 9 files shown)


wolfbeast:
I actually planned to also add the marker bar next to the scroll bar as in the extension on the forum...
Not entirely sure if this patch will allow for that addition later on or if it needs to be changed out in that case. This just seems to be an elaborate way of counting, and nothing else? Is this the Mozilla patch?

trav90:
OK, I wasn't real clear from the forum on which feature(s) you planned to integrate, but I thought I'd play with it a little. Feel free to reject if something better can be done (it didn't really take much time for me to do this).

And yes, this is the (pre-refactored) Mozilla patch ([Bug 257061] (https://bugzilla.mozilla.org/show_bug.cgi?id=257061))

wolfbeast:
I'll have another close look at the extension and integrating that functionality. if problematic I'll merge this pull request, but leaving this open for now.

wolfbeast:
I'm just merging this is. There's a lot of resistance to the scrollbar highlighting feature so it's probably better to leave that to extensions.

wolfbeast:
Question though, does this add language strings...?

wolfbeast:
Doing a final check for 25.6 I noticed that this still doesn't exactly work.
on actual web pages, counts aren't shown. In source view, they are.
It seems the code involved throws errors because aNode.ownerDocument is null, similar to the comment on the AMO bug this PR was based on:
https://bugzilla.mozilla.org/show_bug.cgi?id=257061#c87

Thankfully it doesn't exactly break finding, but it's also not working for many sites.

wolfbeast:
I think I have a workaround for this. I can assume that if a node doesn't have an owner document, that it's not going to be visible. Should be examined if this is always the case though.

trav90:
Sorry, my bad for missing that.

I can confirm that commit e2dcd2e works as intended on Linux.

PR #106, opened 2015-06-18
Fix Pale Moon issue with single word searches under proxies. 

Axiomatic-:
This stems from [here](https://forum.palemoon.org/viewtopic.php?f=37&t=8532&sid=3852a1c547313cf3f792828c669890d1), and this is the proper solution to the issue. This pull request also implements the URI typo fix. (e.g. ttp:// -> http://)

Compiles, and works on my machine. (Sorry for the messy commit log, hope it is not too bad.)


wolfbeast:
Looks good. thanks!

PR #107, opened 2015-06-20
[!] Apply one of Isaac's suggested fixes.

Axiomatic-:
[!] ... this specifically applies to builds that are configured to use --with-system-pixman

Suggestion came from pull request #79.


wolfbeast:
I'll have to defer to your judgement because I'm not too familiar with using system libs on *nix and why this change would help something.

squarefractal:
@Axiomatic- Doesn't this make the `pkg-config` statement entirely meaningless?

Axiomatic-:
@squarefractal not quite since ```||``` only executes the second command if the first command's exit status is zero. As for why this is needed, is because without the ```|| true``` part 'pixman.h' file not found errors will occurs on most *BSD systems, and a few Linux distros (the ones in my mind I can think of would be Gentoo Linux, and Arch Linux).

PR #110, opened 2015-06-22
Remove telemetry from devtools.

squarefractal:


wolfbeast:
Since this PR was against master and not the Code_Cleanup branch I use for this kind of work, I've manually applied it, instead. Closing this.

Issue #109, opened 2015-06-22
Linux builds of current master exits with segfault


squarefractal:
As described, Linux x86-64 builds of current master branch segfaults while exiting the browser. Builds are performed with the CentOS 6.5 and devtoolset-1.1.

The likely suspect here is the screensaver-related fixes made very recently.

Build of [a736ec4] exits with the following segfault:

```
*** Segmentation fault
Register dump:

 RAX: 0000000000000001   RBX: 00007fa3d9db82c0   RCX: 0000000000000006
 RDX: 00007fa3ead00048   RSI: 0000000000000000   RDI: 0000000000000000
 RBP: 00007fa3eacbe0a8   R8 : 00007fa3ead00048   R9 : 0000000000000cad
 R10: 00007fff262e5400   R11: 00007fa3e43ebe40   R12: 0000000000000038
 R13: 00007fa3eacbe1c8   R14: 00007fa3e7ec5a42   R15: 0000000000000000
 RSP: 00007fff262e5680

 RIP: 00007fa3e7b1a9ab   EFLAGS: 00010202

 CS: 0033   FS: 0000   GS: 0000

 Trap: 0000000e   Error: 00000004   OldMask: 00000000   CR2: 00000008

 FPUCW: 0000037f   FPUSW: 00000020   TAG: 00000000
 RIP: e83ff1ae   RDP: 262e3138

 ST(0) 0000 000000000000003c   ST(1) 0000 0000000000000004
 ST(2) 0000 0000000000000001   ST(3) ffff 9000000000000000
 ST(4) ffff c000000000000000   ST(5) ffff c000000000000000
 ST(6) ffff 970fe00000000000   ST(7) 970f 970fe00000000000
 mxcsr: 1fa1
 XMM0:  000000000000000000000000cf400000 XMM1:  000000000000000000000000cf400000
 XMM2:  000000000000000000000000cf400000 XMM3:  000000000000000000000000cf400000
 XMM4:  000000000000000000000000cf400000 XMM5:  000000000000000000000000cf400000
 XMM6:  000000000000000000000000cf400000 XMM7:  000000000000000000000000cf400000
 XMM8:  000000000000000000000000cf400000 XMM9:  000000000000000000000000cf400000
 XMM10: 000000000000000000000000cf400000 XMM11: 000000000000000000000000cf400000
 XMM12: 000000000000000000000000cf400000 XMM13: 000000000000000000000000cf400000
[...]

PR #113, opened 2015-06-25
Fix Linux segfault caused by commit a4e5d5c

trav90:
Resolves issues #109 and #94 for Linux

Confirm that it builds successfully and does not segfault on exit on both Manjaro with GCC 5.1 and CentOS with GCC 4.7.2


wolfbeast:
Fantastic! How did you figure this one out? Just curious :)

trav90:
When I re-landed the WakeLockListener for Linux, I used  the same changes you applied to Windows (i.e. nsCOMPtr<nsIPowerManagerService> sPowerManagerService = nullptr;). So I got to thinking about if that was the correct way to do it for Linux as well, so I reverted back to using "do_GetService" instead and that appears to have resolved it!

PR #116, opened 2015-06-27
HSTS preload update

squarefractal:
Differences from upstream:
* [Rejected possibly spurious "google" entry](https://hg.mozilla.org/mozilla-central/rev/e324d725b95d#l1.129)


wolfbeast:
Thanks!

PR #118, opened 2015-07-01
Remove Mozilla promotional snippets

squarefractal:
While already disabled by way of commenting out of the element in which the snippets are disabled, this PR fully removes the snippets backend code since this is of no use for Pale Moon.

Submitting to Code_Cleanup branch as requested, but git wants to push a lot more; only the last three commits are relevant to this PR.


wolfbeast:
In the future, if you want to make a PR against a master-equivalent feature branch that is behind, ping me so I can fast-forward it to keep the commits cleaner.
I don't see us using these snippets in the future, so yeah, might as well remove the redundant code from the browser ;)

wolfbeast:
Please add a cleanup for the abouthome.xhtml as well.

squarefractal:
_(Just in case if a notification was not generated)._

wolfbeast:
Code edits/updates to a pull request apparently do not generate notifications, so you'll have to add a comment to be noticed.

Issue #120, opened 2015-07-05
Add -moz-os-version support


Lootyhoof:
Using wholesale CSS media queries to determine the version of Windows we're using would make styling for them much easier, especially for Windows 8+, where selectors currently are several nested media queries.

BMO: [810399] (https://bugzilla.mozilla.org/show_bug.cgi?id=810399), [907373] (https://bugzilla.mozilla.org/show_bug.cgi?id=907373) & [1166867] (https://bugzilla.mozilla.org/show_bug.cgi?id=1166867).

wolfbeast:
Good call -- would actually simplify what I've done with the window title as well (using tricks to determine Win 8 with compositor etc.)

Lootyhoof:
Confirmed, it works! Thanks! (25.6b2)

wolfbeast:
Only one problem.. on Win 10, it apparently doesn't work right and it still thinks it's Win 8, causing the issue with toolbar borders (apparently).

> Mozilla/5.0 (Windows NT 10.0; WOW64; rv:25.6) Gecko/20150801 Firefox/31.9 PaleMoon/25.7.0a1

Means that the OS passes the right info to the browser (it's manifested correctly) but wilfully putting an error inside the CSS block that should be excluded in commit 7e09f7235d90d8065cf8bb693a9d9d96928e299e makes it show up in the error console. So, Win10 support still seems to be broken somehow, that it thinks it's win vista, 7 or (most likely) 8

wolfbeast:
Either that or I've just been dumb XD Sorry for the noise.

PR #121, opened 2015-07-07
Add -moz-os-version support

trav90:
Should resolve issue #120.

Possible followup bug: [1169981] (https://bugzilla.mozilla.org/show_bug.cgi?id=1169981)

### DIFF
--- a/content/base/src/nsGkAtomList.h
+++ b/content/base/src/nsGkAtomList.h
@@ -2024,6 +2024,7 @@ GK_ATOM(windows_version_xp, "windows-version-xp")
 GK_ATOM(windows_version_vista, "windows-version-vista")
 GK_ATOM(windows_version_win7, "windows-version-win7")
 GK_ATOM(windows_version_win8, "windows-version-win8")
+GK_ATOM(windows_version_win10, "windows-version-win10")
 
 // And the same again, as media query keywords.
 GK_ATOM(_moz_scrollbar_start_backward, "-moz-scrollbar-start-backward")
--- a/content/base/src/nsGkAtomList.h
+++ b/content/base/src/nsGkAtomList.h
@@ -2018,6 +2018,13 @@ GK_ATOM(windows_theme_royale, "windows-theme-royale")
 GK_ATOM(windows_theme_zune, "windows-theme-zune")
 GK_ATOM(windows_theme_generic, "windows-theme-generic")
 
+// windows version info selector metrics, helpful in removing ambiguity
+// in theme selectors.
+GK_ATOM(windows_version_xp, "windows-version-xp")
+GK_ATOM(windows_version_vista, "windows-version-vista")
+GK_ATOM(windows_version_win7, "windows-version-win7")
+GK_ATOM(windows_version_win8, "windows-version-win8")
+
 // And the same again, as media query keywords.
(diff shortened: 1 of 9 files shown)


wolfbeast:
Looks like the follow-up bug will be needed because of Win 10 MQ support

trav90:
I'll try to get another PR for that submitted this weekend.

trav90:
The above mentioned bug [1169981] (https://bugzilla.mozilla.org/show_bug.cgi?id=1169981) appears to be dependent on the changes made in the first patch of [1150006] (https://bugzilla.mozilla.org/show_bug.cgi?id=1150006).

wolfbeast:
I don't want bug 1150006 because it can and will break existing themes. Maybe you can discuss this with Ryan (Lootyhoof) to see that the impact of it would be, but it'd likely be better to just rewrite the 1169981 patch to apply to our current setup. If you prefer, I can look at that in a follow-up.

wolfbeast:
Actually... did you look at the patch for ESR? the main patch is for trunk but the ESR patch doesn't have the -aero yanked.

trav90:
Yes, I was looking at the ESR patch last night. I posted the comment mostly for my reference for when I work on this the next few days. :)

trav90:
```
"it'd likely be better to just rewrite the 1169981 patch to apply to our current setup."
```

I've gone over the patch(es) on BMO several times, but I'm not sure how to rewrite the patch correctly?

wolfbeast:
The ESR one might apply better, not needing a rewrite.

wolfbeast:
Build failure with this PR: 

 5:39.74 e:/mozdev/PaleMoon/widget/windows/nsLookAndFeel.cpp(402) : error C2039: 'WIN10_VERSION' : i
s not a member of 'mozilla::widget::WinUtils'
 5:39.74
 5:39.74         e:\mozdev\palemoon\widget\windows\WinUtils.h(41) : see declaration of 'mozilla::wid
get::WinUtils'
 5:39.74
 5:39.74 e:/mozdev/PaleMoon/widget/windows/nsLookAndFeel.cpp(402) : error C2065: 'WIN10_VERSION' : u
ndeclared identifier
 5:39.74
 5:39.74 e:/mozdev/PaleMoon/widget/windows/nsLookAndFeel.cpp(402) : error C2051: case expression not
 constant

wolfbeast:
Looks like it's just missing 0xa00 for Windows 10. Easy enough to add.
We'll have to see if it actually picks up the correct version with getVersionEx() because Microsoft has been deprecating it in Win 8.1

wolfbeast:
Commit 724c9ea should fix this

trav90:
Thanks.

trav90:
```
"The ESR one might apply better, not needing a rewrite."
```

The patch as-is on BMO puts it in the middle of a bunch of non-existent code in Pale Moon. I'm not sure where the correct location in the file to apply the patch.

Lootyhoof:
A bit late to be commenting on this, but 1150006 doesn't apply to us anyway; -aero suffixes were added in Fx40 with [70613] (https://bugzilla.mozilla.org/show_bug.cgi?id=706103).

wolfbeast:
We'll have to look at this a different way. Since we don't have the refactored -aero things, we simply need to make use of our new os switches and use those to make the win8 and win10 stuff flat and square.

Issue #122, opened 2015-07-08
Allow chrome://../skin/ overrides for chrome://../skin/ in theme's chrome.manifest


Lootyhoof:
Currently skin packages (themes) cannot override content that is in chrome://../skin/. If such feature were implemented, this would **significantly** improve efficiency when producing themes, as developers would only need to provide a simple toolbar.png replacement in order to create a theme, as opposed to having to copy the entire interface style in order to make simple changes. Among other things, this also allows for platform-specific overrides of particular files (for example, use a browser.css for Windows, Linux, ...), where currently this requires setting up a new skin package to implement.

BMO: [1170207] (https://bugzilla.mozilla.org/show_bug.cgi?id=1170207).

trav90:
Should be resolved by PR #123

wolfbeast:
If you're interested, I can send you a beta version once built to test this functionality (as well as OS selection) since it's all been merged in now. Let me know if so!

wolfbeast:
Verified by @Lootyhoof as working in practice. (25.6 b2)

trav90:
Great!

PR #123, opened 2015-07-09
Allow chrome://../skin/ overrides

trav90:
Issue #122 

Please see note below.


trav90:
I looked at and compared the code with Firefox, and I believe it is now correct.

wolfbeast:
Great! Merging in for the next beta. :)

PR #125, opened 2015-07-10
Unhighlight highlighted text when closing the findbar

trav90:
Here's another one! :)

Currently in Pale Moon if you search for text and click "Highlight All" and then close the findbar, the highlighted text remains highlighted. The only way to remove the highlight is to open the findbar again and reclick on "Highlight all" to remove the highlights. This PR automatically unhighlights the text on closing of the findbar.

Confirmed works as intended on top of current trunk (on Linux x64).


wolfbeast:
:+1: !

PR #130, opened 2015-07-20
[!] Support Link Time Optimization by newer GCC versions.

Axiomatic-:
I reverted the Makefile edit since it does not work, however it is not needed anyway if a person builds Pale Moon with the gold linker instead of the default GCC linker, the error on the elfhack tests do not appear.


wolfbeast:
Looks good, I'll merge it in -- Maybe you can discuss with @trav90 how best to verify these changes and if we want to use LTO for release builds, what needs to be changed in our build procedure, if anything?

trav90:
@Axiomatic- I sent you a PM on the forum regarding this for whenever you have a free moment.

PR #146, opened 2015-08-03
Add support for element.matches.

Axiomatic-:
Pale Moon had support for element.matches, however it was under
a non-standard name.
See: gorhill/ublock#546 as an example showing why adding support
under the standard name is good idea.

Side note, I have not verified this to be working, but I will get around to that at some point time (hopefully). On another note, not sure whether I should push this to Goanna, or the master branch. Sorry if I made a pull request to the wrong branch.

### DIFF
--- a/content/base/public/Element.h
+++ b/content/base/public/Element.h
@@ -591,6 +591,8 @@ public:
   }
   bool HasAttributeNS(const nsAString& aNamespaceURI,
                       const nsAString& aLocalName) const;
+  bool Matches(const nsAString& aSelector,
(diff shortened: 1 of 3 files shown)


wolfbeast:
Yeah, sorry about the branch confusion there.
Goanna is always good for the moment, if I want to take something on the current trunk from there, I can always cherry-pick it across. I'll be making a clarifying post today about the current tree state on the forum.

wolfbeast:
I've backed this out. Please investigate if the keyword is required and if so, how to implement it.

PR #167, opened 2015-09-10
Fix crash when GStreamer decodes certain audio files

trav90:
Builds successfully with CentOS 6 & GCC 4.9.1

Link to test (as reported on BMO): http://die-sondersendung.de/sz004-die-kleinen-werden-so-schnell-erwachsen#t=0:0

Works as intended on CentOS 6 x64.

Also works as intended on Manjaro x64, but the following message is displayed in the terminal:

CRITICAL **: connect_pad: assertion 'factories->n_values > 0' failed

Any pointers on how I can correct the assertion?


wolfbeast:
Sorry, no idea about the assertion off-hand. Can you pinpoint where it's thrown?

trav90:
It's thrown right as the audio starts loading. I originally thought that [this was it] (https://github.com/MoonchildProductions/Pale-Moon/pull/167/files#diff-7a30e6efee0dcdd9b0999d28f4367384R1126), but I did a full clobber build with those 2 lines removed and it still threw the assertion. I'll try to further pinpoint it.

wolfbeast:
No, the line you pointed out was just a nullcheck, not a value check. This is thrown somewhere else, possible in the calling module.

trav90:
I've been poking at this today, and after trying a few things and doing some more research, I'm beginning to think that it might be a GStreamer bug (please, correct me if I'm wrong!). I was looking through the comments on BMO again, and noticed [comment 6] (https://bugzilla.mozilla.org/show_bug.cgi?id=959007#c6):

> autoplug-sort could also be used instead of autoplug-select (as a workaround), but that would lead to an assertion and is not really nice (and will crash for people who build without assertions... don't do that!).

Older versions of GStreamer cause FF/PM to crash when using autoplug-select (which according to the rest of above comment, is the preferable solution), so autoplug-sort is used instead (as-is this PR will only use autoplug-select with GSteamer 1.x once it's supported), but using autoplug-sort causes an assertion with newer versions of GStreamer. :confounded:

Of note, Mozilla uses autoplug-sort exclusively. They also apparently don't care much about the assertion either, as I checked and it's still there in FF 40.0.3 (and this bug was fixed for FF 30).

Suggestions of where I should go from here?

wolfbeast:
I think you can leave this for now. An assertion in the console is better than a crash. If you say that once GStreamer 1.0 is supported, the code you have now will use -select, then that would solve the issue, right? It'd check for gstreamer 1.x and use autoplug-select if so, and will use -sort otherwise (that avoids the crash). Unless I'm missing something in what you said this is the right way to go -- just continue adding GStreamer 1.x support and the assertion should sort itself out once it's in.

trav90:
Exactly. Once GStreamer 1.x is supported, that should solve the issue. I'll try to have the gst1.x support added before Goanna is released.

Issue #133, opened 2015-07-22
Building for Android fails, likely NSPR pull caused


wolfbeast:
Looks like PR #84 broke building Pale Moon for Android completely.

Buildlog output:
```
configuring in nsprpub
running /bin/sh /home/moonchild/pmsrc/nsprpub/configure  --enable-application=mobile/android --target=arm-linux-androideabi --with-android-gnu-compiler-version=4.7 --with-android-version=14 --with-system-zlib --enable-optimize --enable-mobile-optimize --enable-official-branding --disable-webrtc --disable-updater --disable-tests --disable-debug --disable-mochitests --enable-strip --with-android-ndk=/home/moonchild/android-ndk-r8e --with-android-sdk=/home/moonchild/adt-bundle-linux/sdk/platforms/android-19 --with-ccache --with-dist-prefix=/mozdev/pm-android/dist --with-mozilla --disable-debug --enable-optimize --with-arm-kuser --with-android-toolchain=/home/moonchild/android-ndk-r8e/toolchains/arm-linux-androideabi-4.7/prebuilt/linux-x86 --with-android-version=14 --with-android-platform=/home/moonchild/android-ndk-r8e/platforms/android-14/arch-arm --with-arch=armv7-a --with-thumb=yes --with-thumb-interwork=toolchain-default --with-fpu=vfp --with-float-abi=softfp --with-soft-float=toolchain-default --cache-file=.././config.cache --srcdir=/home/moonchild/pmsrc/nsprpub
configure: WARNING: unrecognized options: --enable-application, --with-android-gnu-compiler-version, --with-system-zlib, --enable-mobile-optimize, --enable-official-branding, --disable-webrtc, --disable-updater, --disable-tests, --disable-mochitests, --with-android-sdk
configure: loading cache .././config.cache
configure: error: `host_alias' was set to `arm-linux-android' in the previous run
configure: error: `CFLAGS' has changed since the previous run:
configure:   former value:  `-Wall -Wpointer-arith -Wdeclaration-after-statement -Werror=return-type -Wtype-limits -Wempty-body -Wsign-compare -Wno-unused -mandroid -fno-short-enums -fno-exceptions -march=armv7-a -mthumb -mfpu=vfp -mfloat-abi=softfp -std=gnu99 -fgnu89-inline -fno-strict-aliasing -ffunction-sections -fdata-sections -pipe -g -std=c99'
configure:   current value: ` -fomit-frame-pointer'
[...]

PR #190, opened 2015-10-18
Android issue #133.

Axiomatic-:
@wolfbeast I reverted what you reverted, which reapplies the NSPR patches (at least from looking over the diff files), not sure if this is the correct way for it to be done, if it is not, and you want me to change to a different method let me know. Adding on, this pull request is untested as I cannot build for Android to test.


wolfbeast:
I'll see about making a try build tomorrow.

Issue #100, opened 2015-06-13
Update libvpx to v1.4.x


Axiomatic-:
Currently as I type this, libvpx has a v1.4.0 release. Which can be found [here](https://github.com/webmproject/libvpx/releases). This ideally should be done before Pale Moon hits the 26 milestone in my opinion.

squarefractal:
From the [release notes](https://chromium.googlesource.com/webm/libvpx/+/v1.4.0):

```
- Upgrading:
    This release is ABI incompatible with 1.3.0. It drops the compatibility
    layer, requiring VPX_IMG_FMT_* instead of IMG_FMT_*, and adds several codec
    controls for VP9.
```

wolfbeast:
Reference BMO bugs for this to catch snags:
https://bugzilla.mozilla.org/show_bug.cgi?id=1151175
https://bugzilla.mozilla.org/show_bug.cgi?id=1176730
People following this bug: Please be aware that our libvpx setup is completely different re: build files than what Mozilla uses. Please don't try to adopt it as-in the BMO bug.

wolfbeast:
@mattatobin  -- any progress on this?

mattatobin:
Bumped to the top of my list.. It is about half done.

wolfbeast:
Please make sure to update to a more recent Git version of libvpx, otherwise we'll be vulnerable to a sec issue introduced by it.
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1178215 for the update BMO bug.

wolfbeast:
"Top of your list" hasn't been addressed yet, I guess?

mattatobin:
Success Build on Windows

Linux needs to be tested but /should/ be fine - Tap Travis for try build
Android/ARM needs to be tested as it is the huge X factor - Tap Moonchild for try build

Please use https://github.com/binoc-software/palemoon-contrib/tree/libvpx14x-work

As for me, I am done for today. Will complete this when alt platform trys are attempted.

trav90:
Linux build failure:
```
make[5]: Entering directory `/home/palemoon/PaleMoon/pmbuild/media/libvpx'
[...]

PR #205, opened 2015-10-31
Pull Request to resolve Issue #100 - Update libvpx to 1.4.x (Goanna Edition)

mattatobin:
Just what it says on the tin


wolfbeast:
PR caused build bustage by basing on v25/Gecko build system state, not accounting for WebP changes.
Easy enough to fix, looks like, so I'll keep it merged in and apply a fix on top of it.

Issue #43, opened 2015-04-08
Add "containing folder" in library search results


wolfbeast:
Reference thread: https://forum.palemoon.org/viewtopic.php?f=13&t=7913

Bugzilla bugs:
https://bugzilla.mozilla.org/show_bug.cgi?id=469421
and to add context "go to parent"
https://bugzilla.mozilla.org/show_bug.cgi?id=469441

Extensions implementing this:
https://addons.mozilla.org/en-US/firefox/addon/show-parent-folder/
and
https://addons.mozilla.org/en-US/firefox/addon/go-parent-folder/

PR #211, opened 2015-11-03
Add "Parent Folder" columns to the Library

Antonius32:
In reply to issue #43, this pull request adds a "Parent Folder" column and a "Parent Folder Path" column to the Library. Tested and confirmed working on Linux x64 (thanks, @trav90!).


wolfbeast:
Thanks for this!

Antonius32:
Thanks for including it! I'll see if I can create a patch for the "Open parent folder" feature as well.

PR #218, opened 2015-11-15
Improve styling of Back and Forward buttons in the default theme for Windows

Antonius32:
With the current styling of the toolbar buttons in the default theme for Windows, I think it's rather difficult to see the difference between the idle (not disabled) and hover states of the Forward button. This pull request aims to improve that.

It also eliminates the transition delay for the hover state of the Back and Forward buttons, to make them "feel" more responsive and to make the hover response of these buttons consistent with the hover response of the other buttons on the Navigation toolbar.

Lastly, I have rearranged the code in the main style sheet in a few places, just to put some more logic to it and to "tidy up" a bit.

To demonstrate the effect, I have put together a theme that includes the relevant changes. I can't upload it here, but if you're interested, I can send you the theme by email or by PM (using the messaging system on the forum).

Before:
before

After:
after

### DIFF
--- a/browser/themes/windows/browser.css
+++ b/browser/themes/windows/browser.css
@@ -19,6 +19,7 @@
 %define navbarLargeIcons #navigator-toolbox[iconsize=large][mode=icons] > #nav-bar
 %define forwardTransitionLength 150ms
 %define conditionalForwardWithUrlbar window:not([chromehidden~=toolbar]) #navigator-toolbox[iconsize=large][mode=icons] > :-moz-any(#nav-bar[currentset*="unified-back-forward-button"],#nav-bar:not([currentset])) > #unified-back-forward-button
+%define conditionalForwardWithoutUrlbar window:not([chromehidden~=toolbar]) #navigator-toolbox[iconsize=large][mode=icons] > :-moz-any(#nav-bar:not([currentset*="unified-back-forward-button,urlbar-container"]),#nav-bar:not([currentset])) > #unified-back-forward-button
 %define conditionalForwardWithUrlbarWidth 27
 
 %ifdef MOZ_OFFICIAL_BRANDING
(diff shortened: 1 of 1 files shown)


wolfbeast:
Also, if you're going to remove things, then please remove them, don't just comment them out, please. we have a VCS for a reason ;) No need to keep commented code beyond WIP stage.

Antonius32:
> Have you tested this on other non-aero Windows operating systems as well?

No, I haven't. For what it's worth, I've tested it with Aero disabled, and all seems well.

> Also, if you're going to remove things, then please remove them, don't just comment them out, please.

Done!

Antonius32:
By the way, I can also restore the transition effects, and set all durations to 150ms, if you prefer.

Antonius32:
And another option would be to set the background image for the Back and Forward buttons to the exact same values as the background image for the other buttons on the Navigation toolbar. (I prefer to have the Back and Forward buttons look slightly different, though.)

Antonius32:
I have just made another commit (SHA: b18d321), that smoothens/rounds the top-right and bottom-right (top-left and bottom-left for rtl-oriented locales) corners of the Forward button when it's not merged with the URL bar, which makes it look more in-line with the other buttons on the Navigation toolbar.

Initially, I wanted to create a separate pull request for this commit, but I figured I might as well include it with this one.

wolfbeast:
Looks fine, code-wise. Can you verify that the rounded corners aren't interfering with the Win10 styling I introduced (that removes them from all buttons)?

Antonius32:
I have just made another commit (SHA: d01ae44) to take care of that.

wolfbeast:
Awesome - thanks!

Antonius32:
You're welcome, thanks for including this!

PR #220, opened 2015-11-16
Small correction in placesOverlay.xul

Antonius32:


wolfbeast:
Was there something wrong with your code?

Antonius32:
The commit merely rearranged the existing code somewhat, aiming to add logic, but I decided it wasn't necessary after all.

Maybe I will re-use this pull request for another commit later on, though.

PR #225, opened 2015-11-19
Make sure orientation-specific styling in the theme for Windows is applied properly

Antonius32:
This pull request makes sure that the styling of the Forward button, that was added recently by pull request #218, is applied properly both for ltr- and rtl-oriented locales.

When this patch is applied, I'm confident that the enhancements to the theme will work as intended.


wolfbeast:
For what few RTL users we have ;)

Antonius32:
Maybe their numbers will grow. At least we're prepared now!

Thanks for including this.

PR #229, opened 2015-11-21
Set ui.key.menuAccessKeyFocuses to "true" by default on Linux

Antonius32:
Currently on Linux, the preference ui.key.menuAccessKeyFocuses is set to "false" by default. I think it makes sense to set this preference to "true" by default, so when the Menu bar is hidden, pressing the Alt key will bring it up, just like on Windows.

### DIFF
--- a/modules/libpref/src/init/all.js
+++ b/modules/libpref/src/init/all.js
@@ -3647,6 +3647,8 @@ pref("ui.panel.default_level_parent", true);
 
 pref("mousewheel.system_scroll_override_on_root_content.enabled", false);
 
+pref("ui.key.menuAccessKeyFocuses", true);
+
 # XP_UNIX
 #endif
 #endif


wolfbeast:
Good catch -- Since I'm not a Linux user myself and if installed I always have the menu bar enabled by default anyway on it, I'd never even have thought about it.

Antonius32:
Seeing that "AppMenu mode" is the standard on Windows, I was thinking it might be a good idea to make it the standard on Windows as well. What are your thoughts on that?

Anyway, thanks for including this!

wolfbeast:
> What are your thoughts on that?

Not a good idea, because AppMenu mode on Linux in general conflicts with the standard window frame in WMs, and doesn't make it very OS-integrated.

Antonius32:
I see. I'm closing this branch, then.

PR #242, opened 2015-12-02
Add stdc++ compat hacks for newer GCC versions (and remove some old ones)

trav90:
Just what it says in the tin.

Confirmed working with GCC 4.9 on CentOS x64 and GCC 5.2 on Manjaro x64.

Will test with GCC 4.7 on CentOS 6 shortly (although it *should* be fine, since I only made changes for GCC 4.9+ and 4.2-)


wolfbeast:
I'll trust your expertise on this one :)

PR #245, opened 2015-12-04
Cleanup of dead GnomeVFS code.

Axiomatic-:
There was an earlier pull request from Travis that dealt with the main bulk; however, it missed a few things, this pull request cleans up those missed things.


trav90:
I had actually seen this code yesterday, but hadn't had a chance to remove it yet. Thanks for catching! :+1:

PR #216, opened 2015-11-14
Add "Open Parent Folder" item to the context menu for bookmarks search results

Antonius32:
This pull request adds an "Open Parent Folder" item to the context menu for bookmarks search results in the Library and in the Bookmarks Sidebar. Tested and confirmed working on Linux x64.

This is a follow-up on pull request #211, aiming to resolve issue #43 altogether.

@trav90: Thank you for helping me with this!

PR #247, opened 2015-12-04
In the Library, replace the term "Parent Folder" with "Containing Folder"

Antonius32:
I noticed that the context menu for downloads includes an "Open Containing Folder" item, and from a consistency point of view, I think it's best to use the term "Containing Folder" for the "Parent Folder" features of the Library as well.

This is a follow-up to PRs #211 and #216.


wolfbeast:
Good catch! Being consistent in your wording is a Very Good Thing™ :smile:

Antonius32:
:thumbsup:

PR #249, opened 2015-12-06
Allow use of a dedicated flag to not precompile the startup cache

trav90:
Commit afcb587 disabled the startup cache on BSD. This PR reverts that change, and allows the startup cache to be disabled with an option in .mozconfig instead.

Since the issue with failing to fill the startup cache has been reported for Linux as well, this PR provides a cross-platform  option to disable the startup cache on an as needed basis. This will allow BSD users and Linux users who have encountered this problem to still complete the packaging process without having to modify the source until we can find the root cause for the problem.

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -7382,12 +7382,12 @@ if test -n "$_ENABLE_CODESIGHS"; then
 fi
 
 dnl ========================================================
-dnl = Allow disabling the startup cache
+dnl = Don't precompile the startup cache
 dnl ========================================================
 MOZ_DISABLE_STARTUPCACHE=
 
 MOZ_ARG_DISABLE_BOOL(startupcache,
-[  --disable-startupcache          Disable the startup cache ],
+[  --disable-startupcache          Don't precompile the startup cache ],
     MOZ_DISABLE_STARTUPCACHE=1,
 	MOZ_DISABLE_STARTUPCACHE=)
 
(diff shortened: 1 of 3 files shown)


wolfbeast:
Otherwise this is a good idea, at least as a temporary measure until we can get to the bottom of this.

trav90:
Fixed. I also update the comments in configure.in to reflect this.

In addition, do you think it would be better to use a different flag such as --without-startupcache or similiar (since --disable-startupcache could be a little misleading), or will the comment in configure.in be enough?

wolfbeast:
Hmm.. good point. it should have a clear name to indicate exclusion of the precompiled startup cache.

--without-startupcache is.. OK I guess, better than --disable
maybe --no-precompiled-startupcache ? I dunno, kind of drawing a blank on a good name.

trav90:
Agreed, I'm not to fond of --without-startupcache cache either. I've been chewing on this, and what do you think about --disable-precompiled-startupcache? I know it's kinda long, but it is clear on what it does and it doesn't break form with most of the other options in configure.in

wolfbeast:
Sounds good. Probably as good as we're gonna get for something this specific ;)

trav90:
Agreed. I've added another commit that should take care of it.

trav90:
I am kicking off a build to ensure everything works as-intended. Will report back later today.

wolfbeast:
OK, I'll wait for that.

trav90:
For some reason, after changing it to --disable-precompiled-startupcache, it no longer works as intended. Going to investigate further.

trav90:
OK, this PR as-is is confirmed working as intended and should be good to merge now.

wolfbeast:
All right. Thanks!

PR #253, opened 2015-12-06
Added a new pref security.xssfilter.blockDynamic (default: true) to control whether to also monitor DOM-Based XSS vectors. Currently it only affects external scripts

BruceBerry:


wolfbeast:
Looks good. I'll merge it in and toss a new build in the oven ;)

PR #254, opened 2015-12-07
Added whitelist support. missing UI support for now

BruceBerry:


BruceBerry:
I suck at mozilla strings, so check nsXSSFilter::InitializeWhiteList

Also, i forgot a printf in there...

wolfbeast:
I'm not exactly fluent in nsAutoCString juggling either, but as far as I can tell you're doing things right, there. I'll yank out the printf after merging and will just see how it behaves or if there's issues.

wolfbeast:
I can confirm it works exactly as-intended. I ran a testset through with the known FP:

- (blank) hit
- example.com hit
- yimg.com pass
- yimg.com,example.com pass
- example.com,yimg.com pass
- example.com,yimg.com,example2.com pass
- example.com,example2.com hit

BruceBerry:
Can you take care of the infobar "add domain" button? It should be as simple as reading the pref, adding ",newdomain" and writing it back, which will automatically call `InizializeWhiteList`. I guess in the end it should also refresh the current URL. It should be in `browser.js`. I don't know if you want a similar button when block mode is active, that's up to you.

wolfbeast:
Yeah I can do the front-end stuff, no problem.

When block mode is active, I definitely don't want an easy way for a user to bypass it, so I don't think adding a button there is a good idea.

PR #263, opened 2015-12-17
Cleanup various GCC checks

trav90:
Confirmed works on Linux with GCC 4.7 & 4.9


wolfbeast:
Assuming the Linux-based OSes in question defining __domainname or nothing at all are obsolete.
Thanks for digging into this!

trav90:
You're welcome! :)

Issue #175, opened 2015-09-21
Add site indicator to downloads


wolfbeast:
The download managers currently list what sites a file was downloaded from after completion, but not on in-progress downloads.

See also: https://forum.palemoon.org/viewtopic.php?f=13&t=9620

This would probably be a simple yet nice enhancement to give people some more information about their running downloads. Could be as simple as adding an extra label after the file name for running downloads with already-known domain information.

Antonius32:
I can take a shot at this, if you want.

wolfbeast:
Go ahead!

wolfbeast:
Is this complete at this point or should I re-open this issue until it is?

Antonius32:
It's complete now, no need to re-open this issue.

wolfbeast:
Awsome then! Thanks :)

PR #265, opened 2015-12-17
Add a site indicator to in-progress downloads

Antonius32:
Aiming to resolve issue #175, this pull request introduces the following changes:

- In the Library and in the classic Download Manager, add the host display name to the display name (the main label) for in-progress downloads, and add the full host name to the display name tooltip
- In the Downloads panel, add the full host name to the display name tooltip for in-progress downloads (Since space is limited here, I decided not to add the host name to the display name itself.)

Library:

library

Classic Download Manager:

classic download manager

Downloads panel:
downloads panel


wolfbeast:
Looks great!

Antonius32:
Thanks, I'm glad you like it!

PR #273, opened 2015-12-19
Use built-in overflow-checked functions in supported compilers

trav90:
jsnum.h contains utility functions SafeAdd, SafeSub, and SafeMul for doing overflow-checked arithmetic. While it's theoretically possible for compilers to pattern-match this into something more efficient, it doesn't always happen.

Recent versions of GCC & Clang have built-in overflow-checking for these functions:

https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html#Integer-Overflow-Builtins
http://clang.llvm.org/docs/LanguageExtensions.html#checked-arithmetic-builtins

This PR will enable those overflow-checking functions when the compiler supports it.

### DIFF
--- a/js/src/jsnum.h
+++ b/js/src/jsnum.h
@@ -13,6 +13,21 @@
 
 #include "vm/NumericConversions.h"
 
+// This macro should be "1" if the compiler used supports built-in
+// functions like __builtin_sadd_overflow.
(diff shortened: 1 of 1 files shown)


wolfbeast:
LGTM! Thanks.

PR #276, opened 2015-12-21
Replace the History icon in the Aero theme for Windows

Antonius32:
Currently, the Aero theme for Windows comes with an Aero-specific version of the History icon for the Library (which is used in the various History menus as well), but I think the non-Aero version actually looks better. This pull request removes the Aero-specific version of the icon, so the non-Aero version is used instead.


wolfbeast:
Mixing and matching different looks of icons (from different sets) is usually not a good idea. The aero set keeps the same underlying design in all icons, so I'm not very inclined to take this change. Can you please look objectively at how the proposed non-aero icon would compare to the rest of the set?

Antonius32:
Sorry, I should have posted some screenshots when I opened this.

The Library folder tree:

library folder tree

And the History menu:

history menu

I think the non-Aero History icon fits in well with the Aero icons, and I think you could even consider replacing the Tags icon as well.

Let me know what you think!

win7-7:
Antonius32 This is your personal option of icons. You are trying to force your personal options to other users of Pale Moon browser and that is not good idea. I personally like Aero history icon more than non-Aero history icon and it looks better with Aero theme. I personally thing that currents icons are best fitting to Aero theme. This pull request should be rejected as its Antonius32 personal option how he thinks icons should look like instead of actually improving back-end underlying performance or adding features.

Antonius32:
@win7-7: Thank you for sharing your opinion on this. I don't think making a suggestion / creating a pull request is the same as forcing one's opinion on others, though. I've made my suggestion to make a change; you've made yours to keep things the way they are. No one is forcing their opinion on anyone, as far as I can tell.

wolfbeast:
I've looked at the differences and comparisons, also pinged some of my peers and we all agreed that the main history icon is a definite improvement over the aero one and fits well in the set, and the tags one being a toss-up, so best left as-is.

Thanks for the suggestion!

Antonius32:
You're welcome, and thank *you* for taking it into consideration!

PR #284, opened 2015-12-31
Bug 1112501

Axiomatic-:
ColumnSet, transforms, and fixed positions do not interact right this patch fixes that.

Reason: [Bug 1112501](https://bugzilla.mozilla.org/show_bug.cgi?id=1112501)

Tested, and confirmed working.


wolfbeast:
Sorry if I'm a little dense. What, *exactly*, does this fix?

Axiomatic-:
Currently there is a bug where the <span> element does not end up where it is supposed to be, when used in conjunction with column-count CSS.

You can produce the issue [here](https://hg.mozilla.org/integration/mozilla-inbound/raw-file/7cc47794c45d/layout/generic/crashtests/723108.html).

wolfbeast:
Still doesn't help me understand what the actual issue is, but I'll merge it in.

PR #288, opened 2016-01-03
Enable PulseAudio with libcubeb by default.

Axiomatic-:
Allows PulseAudio to be used even without compiling support (i.e. being determined at run time.).


Axiomatic-:
Oops, @wolfbeast your comment disappeared because I ```git reset``` the commit, then well yeah (oops). My mistake, sorry. 

I addressed two of the points you brought up (the GStreamer changes, and MOZ_PULSEAUDIO_CFLAGS), as for the third point, the one referring to MOZ_PULSEAUDIO_LIBS, libcubeb does not require it for PulseAudio integration (uses in-tree uses dlopen libpulse).

wolfbeast:
OK!

Issue #287, opened 2016-01-01
Remove "Connecting..." name from loaded sites with dynamic content


Lootyhoof:
On sites that load additional content the more you explore a particular page (e.g. Amazon, Facebook), the tab title changes to "Connecting..." as it would when initially loading a page. While this is technically expected behaviour, on sites that do this often it could be quite an annoyance if the tab title is constantly changing, plus it would make it less obvious what tab it is, especially if the user has a lot of them open.

Note that the below BMO only removes the word "Connecting..." from the tab title (if the site has already loaded and is only providing dynamic content); it doesn't remove the tab loading icon.

BMO: [1213438] (https://bugzilla.mozilla.org/show_bug.cgi?id=1213438).

wolfbeast:
Keeping the loading indicator is a good thing IMO, so you know which tabs are actually causing network traffic.

Lootyhoof:
True, though in this instance the above bug still keeps the loading icon itself, so in theory it should still be possible to know this from the tab icon.
I'm considering more such as that specified in [697725] (https://bugzilla.mozilla.org/show_bug.cgi?id=697725) (which, although unlinked to the above, is the same issue), where the tab is constantly switching between "Connecting..." and the tab title due to constantly having additional content.

wolfbeast:
I wasn't disagreeing with you XD
I was just saying that keeping the loading indicator (the spinny) is a good thing, so what the bug does is probably what we want.

PR #292, opened 2016-01-05
Don't set tab title to "Connecting..." when loading additional content

Antonius32:
Aiming to resolve issue #287, this pull request makes sure the tab title is not set to "Connecting..." when additional content is being loaded (after the initial page load).


wolfbeast:
LGTM! Thanks!

PR #296, opened 2016-01-09
Fixed minimum length, filter out matches without special characters

BruceBerry:


wolfbeast:
Looks good! I'll make a build and give it a go. Good times :)

PR #302, opened 2016-01-14
Disable -Wuninitialized warning for js/Vector.h, fixes #298

stevenschlansker:
Not sure that this is the best fix, but tested in Mac clang to dramatically reduce warning volume.

Issue #298, opened 2016-01-12
js/Vector.h produces dizzying number of "field 'storage' is uninitialized when used" warnings


stevenschlansker:
```
14:07.35 Warning: -Wuninitialized in /Users/pale/jenkins/workspace/palemoon/obj-x86_64-apple-darwin12.5.0/dist/include/js/Vector.h: field 'storage' is uninitialized when used here
14:07.35 ./../../dist/include/js/Vector.h:531:34: warning: field 'storage' is uninitialized when used here [-Wuninitialized]
14:07.37   : AllocPolicy(ap), mBegin((T *)storage.addr()), mLength(0),
14:07.37                                  ^
14:07.38 /Users/pale/jenkins/workspace/palemoon/js/src/jsgc.h:705:5: note: in instantiation of member function 'js::Vector<void **, 16, js::SystemAllocPolicy>::Vector' requested here
14:07.38     GCHelperThread(JSRuntime *rt)
```

A huge number (at least hundreds, maybe thousands?) of source files produce warnings much like the above -- I'm not sure whether it's harmful or not, but it seems like fixing it once in the header file could dramatically reduce the size of the build output (and probably therefore the speed of the build, as logging to stdout is a measurable amount of time in such a large build)

This is with:
```
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 5.1 (clang-503.0.40) (based on LLVM 3.4svn)
Target: x86_64-apple-darwin12.5.0
Thread model: posix
```

wolfbeast:
The code is from this snippet:
```` C++
/* Vector Implementation */

template <class T, size_t N, class AllocPolicy>
JS_ALWAYS_INLINE
Vector<T,N,AllocPolicy>::Vector(AllocPolicy ap)
  : AllocPolicy(ap), mBegin((T *)storage.addr()), mLength(0),
    mCapacity(sInlineCapacity)
#ifdef DEBUG
  , mReserved(sInlineCapacity), entered(false)
#endif
{}
````
This is perfectly fine, code-wise (it's implicitly initialized as null). Your compiler just doesn't seem to like it and is trying to be helpful.

stevenschlansker:
[...]

PR #305, opened 2016-01-16
Disable -Wuninitialized warnings for js/Vector.h

trav90:
Followup to PR #302

Since Clang can use most GCC pragma, this PR will silence the -Wuninitialized warnings discussed in Issue #298 with both GCC and Clang.

Tested and seems to work as intended with both GCC & Clang on my Manjaro Linux box.


wolfbeast:
Still needs a conditional, then, or you'll spit out warnings on MSVC about an incorrect `#pragma`, no?

trav90:
OK, I just pushed a new commit that I think solves it.

PR #311, opened 2016-01-24
Disable i386 build for Mac; remove MOZ_BUILD_PROJECTS because

stevenschlansker:
it seems to refer to missing $(ARCH)/build.mk files

I'll admit I don't fully understand this change, but setting this doesn't seem to
do anything useful except break the build.

( @mattatobin this replaces the `MOZ_BUILD_APP` change you didn't like, hopefully this makes more sense? )

### DIFF
--- a/build/macosx/universal/mozconfig
+++ b/build/macosx/universal/mozconfig
@@ -3,9 +3,6 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 # i386/x86-64 Universal Build mozconfig
-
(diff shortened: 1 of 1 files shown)


mattatobin:
You can't just arbitrarily do that... See our historical common ancestor for why. http://mxr.mozilla.org/mozilla-esr24/search?string=MOZ_BUILD_PROJECTS

wolfbeast:
Closing this. It's not an acceptable change.

PR #309, opened 2016-01-23
Three minor aesthetic improvements (default theme for Windows)

Antonius32:
This pull request introduces the following changes:

- Make the tab corners more distinct/clean-cut (and remove unneeded "background-repeat" declaration)
- Add a small visual gap between the tabs and the top edge of the screen when the tabs are on top and Pale Moon is maximized or in full-screen mode
- Make sure that when the tabs are *not* on top, the Navigation toolbar buttons are more or less vertically centered between the Bookmarks toolbar / the tabs on one side, and the AppMenu button on the other

This is a follow-up to commits 1204d1f and 09f2f5c.

Before:

screenshot-1

screenshot-2

screenshot-3

After:

screenshot-4

screenshot-5

screenshot-6

PR #312, opened 2016-01-24
Apply the same top margin to the Bookmarks toolbar and the tabs

Antonius32:
Striving for UI consistency, this pull request makes sure the same top margin is applied to the Bookmarks toolbar and the tabs when the tabs are *not* on top, so the distance between the Navigation toolbar buttons and the first element below the Navigation toolbar is always the same (whether the Bookmarks toolbar is enabled, or not).

This is a follow-up to PR #309 / commit fad8b25.

Before:

screenshot-1

After:

screenshot-2

### DIFF
--- a/browser/themes/windows/browser-aero.css
+++ b/browser/themes/windows/browser-aero.css
@@ -553,11 +553,11 @@
   }
 
   #navigator-toolbox[tabsontop=false] > #PersonalToolbar {
-    margin-top: 3px;
+    margin-top: 2px;
   }
   
   #navigator-toolbox[tabsontop=false] > #PersonalToolbar:not(:-moz-lwtheme) {
-    margin-top: 2px;
+    margin-top: 1px;
     border-top: 1px solid @toolbarShadowColor@;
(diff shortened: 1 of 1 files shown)


wolfbeast:
... is this a thing or not, then?

Antonius32:
Well, I'm not quite sure about it.

Arguments in favor of this change:

- When the tabs are *not* on top, the distance between the Navigation toolbar buttons and the first element below the Navigation toolbar is always the same.
- Looks better when the tabs are *not* on top, the Bookmarks toolbar is enabled and a lightweight theme is in use.

Arguments against this change:

- Keep things the way they are now (and have been in the latest release versions). If it ain't broke, don't fix it.
- The current styling looks better (IMO) when the tabs are *not* on top, the Bookmarks toolbar is enabled and the default theme is in use.

What do you think?

wolfbeast:
Current styling looks better in the default setup Pale Moon is released in, so .. I'd say don't touch it.

Antonius32:
I aggree; I won't reopen this pull request, then.

PR #314, opened 2016-01-26
Hack workaround for zero length ArrayLength call on unixThemeStrings:

stevenschlansker:
Building with Mac clang leads to:

```
/Users/pale/jenkins/workspace/palemoon/layout/style/nsMediaFeatures.cpp:387:32: error: no matching function for call to 'ArrayLength'
      for (size_t i = 0; i < ArrayLength(unixThemeStrings); ++i) {
                             ^~~~~~~~~~~
../../dist/include/mozilla/Util.h:303:1: note: candidate template ignored: substitution failure [with T = const UnixThemeName, N = 0]: zero-length arrays are not permitted in C++
ArrayLength(T (&arr)[N])
^                    ~
```
Looking at the Unix themes, they are QT4 and GTK2, neither of which makes sense on a Mac.  And they are already `#ifdef`'ed out on Windows, so let's just do that on Mac too?

Let me know if there's a cleaner fix!  I don't like the taste of using preprocessor directives for this, but it seems to be how it's done here.

### DIFF
--- a/layout/style/nsMediaFeatures.cpp
+++ b/layout/style/nsMediaFeatures.cpp
@@ -378,7 +378,7 @@ GetUnixTheme(nsPresContext* aPresContext, const nsMediaFeature* aFeature,
                 nsCSSValue& aResult)
 {
     aResult.Reset();
-#if !defined(XP_WIN)
+#if !defined(XP_WIN) && !defined(XP_MACOSX)
     int32_t metricResult;
     if (NS_SUCCEEDED(
         LookAndFeel::GetInt(LookAndFeel::eIntID_UnixThemeIdentifier,


stevenschlansker:
It might be nice to alternatively fix `ArrayLength` to handle 0-length arrays, but templates scare me...

wolfbeast:
Preprocessor directives are used all over the place. We have to, because it's a single code base for multiple platforms AND compilers. There's no other way to do this without making things very complex or duplicating tons of code.
The change here is probably exactly what we need, although I'm unsure why the array would come back as 0-length (and that is indeed illegal in any flavor of C++) I'll have to research a little where GetUnixTheme is exactly used.

stevenschlansker:
I'm fairly confident it comes back as zero-length because neither `MOZ_WIDGET_GTK2` nor `MOZ_WIDGET_QT` are defined on a Mac.

PR #316, opened 2016-01-26
Bring in upstream fixes to Mac makefiles

stevenschlansker:
Upstream FF has changed the Mac build infrastructure somewhat -- in particular it seems they are moving to simplify the "universal" build, which makes sense given that the Mac world has moved away from i386 and ppc entirely.

This ends up dropping support for ancient versions of Make, as well as bringing in a couple of trivial fixups.  They are not strictly necessary but I tried to bring them all in to keep the history as similar to upstream as possible, rather than dealing with conflicts picking out just the bits I need and then again if we cherry-pick later fixes.

Let me know if I should make a different patch with only the parts I care about (the ` build/macosx/universal/flight.mk` bits) at the expense of making future cherry-picks harder.

(How do I verify that these changes do not break Linux / Windows builds?  Is there a Pale Moon CI that will automatically build such versions?)

### DIFF
--- a/addon-sdk/Makefile.in
+++ b/addon-sdk/Makefile.in
@@ -30,7 +30,7 @@ TEST_FILES = \
 # Remove this once the test harness uses the APIs built into Firefox
 TEST_FILES += source/lib
 
-PKG_STAGE = $(DIST)/test-package-stage
+PKG_STAGE = $(DIST)/test-stage
 
 stage-tests-package:: $(TEST_FILES)
 	$(INSTALL) $^ $(PKG_STAGE)/jetpack
--- a/browser/app/profile/extensions/Makefile.in
+++ b/browser/app/profile/extensions/Makefile.in
@@ -8,7 +8,7 @@ topsrcdir  = @top_srcdir@
 srcdir     = @srcdir@
 VPATH      = @srcdir@
 
-DISTROEXT = $(call core_abspath,$(FINAL_TARGET))/distribution/extensions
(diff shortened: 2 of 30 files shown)


wolfbeast:
@mattatobin Maybe you can have a critical look at these build system changes? I get lost too quickly in Python.

mattatobin:
fdb5704 seems completely arbitrary. Building tests is barely supported anyway.. IMO the whole test system and files should be killed with fire.

For same and similar reasons above..  442ea5d likewise should not be accepted

As for a0a1457 .. I cannot effectively predict what will happen with this change on other platforms as well as comm.. Additionally, this patch is incomplete.

All in all, the contributor should not be overall reaching for Mozilla changes to the build system. We are not striving for build system parity with Mozilla. Also, these kinds of changes lack scope for the larger implications and impacts on other platforms. He should also make absolutely sure that ANY patches applied as-is from Mozilla sources are checked out regardless of if they apply cleanly or not.

A good way especially for build system is to check MXR at the mozilla-esr24 level.

wolfbeast:
Sounds like:

> Let me know if I should make a different patch with only the parts I care about (the `build/macosx/universal/flight.mk` bits) at the expense of making future cherry-picks harder.

is the better approach, here. I think what you should probably do is only pick up the parts that are required to have a working build on our build system (which is very much different from Mozilla's by now because they kept changing things around).

Are any of these patches actually required for us?

stevenschlansker:
I will make a smaller patch of only the parts of these patches that directly affect my work.

PR #317, opened 2016-01-26
Pale Moon 26 - patch for the default Windows theme

Antonius32:
This pull request introduces the following changes:

- Make sure that when the tabs are *not* on top, the Navigation toolbar buttons are vertically centered between the Bookmarks toolbar / the tabs on one side, and the AppMenu button (or the title bar) on the other
- Set margin on the Tab bar as a whole, instead of on the individual items it holds
- Make the tab corners more distinct/clean-cut
- Add a small visual gap between the tabs and the Menu bar / the top edge of the screen
- Make sure scrolling down to the last tab with the mousewheel disables the tab scrollbox's scroll-down arrow (for reference, see [this] (http://forum.palemoon.org/viewtopic.php?f=45&t=10828#p75823) post on the forum)

Additional change:

- Make sure that on Aero-enabled systems, the Tab bar is always transparent in full-screen mode

This is a revision of / follow-up to PRs #309, #310 and #313, which were reverted by commit f7c143e.

### DIFF
--- a/browser/themes/windows/browser-aero.css
+++ b/browser/themes/windows/browser-aero.css
@@ -553,11 +553,11 @@
   }
 
   #navigator-toolbox[tabsontop=false] > #PersonalToolbar {
-    margin-top: 1px;
+    margin-top: 3px;
   }
   
   #navigator-toolbox[tabsontop=false] > #PersonalToolbar:not(:-moz-lwtheme) {
-    margin-top: 0px;
+    margin-top: 2px;
     border-top: 1px solid @toolbarShadowColor@;
     background-image: linear-gradient(@toolbarHighlight@, rgba(255,255,255,0));
   }
--- a/browser/themes/windows/browser-aero.css
+++ b/browser/themes/windows/browser-aero.css
@@ -466,7 +466,7 @@
   #toolbar-menubar:not(:-moz-lwtheme),
   #TabsToolbar[tabsontop=true]:not(:-moz-lwtheme),
   #nav-bar[tabsontop=false]:not(:-moz-lwtheme),
-  #nav-bar + #customToolbars + #PersonalToolbar[collapsed=true] + #TabsToolbar[tabsontop=false]:last-child:not(:-moz-lwtheme) {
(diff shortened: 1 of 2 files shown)


Antonius32:
@wolfbeast Tested and confirmed working on Windows 7. Should work on Windows 8 and 10 as well.

wolfbeast:
I'll have a look at this tomorrow with a build and a test run on XP, 8 and 10

Antonius32:
All right, I hope it all works as intended now!

wolfbeast:
Getting there. Tested some on Win 7 and Win 8 until now while having a moment.

A few issues:

- All versions, fullscreen mode, tabs on bottom: the tab bar background is not glass/toolbar when the bookmarks toolbar is enabled (but not visible in fullscreen mode). I'm not sure if this is actually a different bug (might be) but you may be able to tweak it while you're in there?
- Win8 (and probably win 10) maximized+fullscreen, tabs on top: Tabs are rounded. Also, this seems to be a larger rounding than normal? Please keep things consistent. Define rounding once for each OS and don't override it later, to prevent this.

Borders seem to be drawn correctly :+1:

wolfbeast:
well that was based on the main commits, not the additions added more recently...

Antonius32:
I've just made a last-minute adjustment to make sure the patch works as intended for Windows 8.

wolfbeast:
Github was being odd - it showed 2 more entries with 2 files changed each, 22 hours ago...

Antonius32:
Well, I rebased to change some of the comments in the mean time, maybe that's why. I haven't changed any actual code until just now, though.

Antonius32:
Anyway, I think the rounding of the tabs should be OK now.

I'm not sure if the the Tab bar background in full-screen mode is a bug; I think that with the darker background, it would be hard to distinguish the All Tabs button. But for consistency, maybe the tabs should have the transparent border here as well. What do you think?

Antonius32:
Commit 3531cef makes sure that on Aero-enabled systems, the Tab bar is always transparent in full-screen mode.

As far as I'm concerned, the patch is good to go now.

Antonius32:
After a final revision of the first commit, the patch should be good to go now.

PR #318, opened 2016-01-26
Add OSXRunLoopSingleton from upstream mozilla to libcubeb

stevenschlansker:
I'm not sure why this goes into libcubeb, but cubeb_osx_run_loop.c
definitely links against it and won't compile without it.


Axiomatic-:
Just a note since I did some digging, the first time this pops in the Mozilla tree is [bug 1045018](https://bugzilla.mozilla.org/show_bug.cgi?id=1045018). I had a quick look, it has been refactored into another location since it's initial creation.

wolfbeast:
It's OK for now, but does need a note to take care if we upstream update libcubeb after this.

Issue #301, opened 2016-01-14
nsAppShell.mm: parameter type `nsIDOMMozWakeLockListener` is an abstract class


stevenschlansker:
I apologize for peppering the project with build issues, if it wasn't obvious already I am trying to fix the Mac build and I think that means I am exercising some code paths that are often not built currently.  Hopefully trying to fix the Mac build is a contribution that forgives my spam :smile: 

Again using `Apple LLVM version 5.1 (clang-503.0.40) (based on LLVM 3.4svn)` [the build fails on `nsAppShell.mm`](https://github.com/MoonchildProductions/Pale-Moon/blob/master/widget/cocoa/nsAppShell.mm#L305):

```
863:04.76 /Users/pale/jenkins/workspace/palemoon/widget/cocoa/nsAppShell.mm:305:65: error: parameter type 'nsIDOMMozWakeLockListener' is an abstract class
863:04.76 NS_IMPL_ISUPPORTS(MacWakeLockListener, nsIDOMMozWakeLockListener)
863:04.76                                                                 ^
863:04.77 ../../dist/include/nsISupportsBase.h:55:14: note: unimplemented pure virtual method 'QueryInterface' in 'nsIDOMMozWakeLockListener'
863:04.78   NS_IMETHOD QueryInterface(REFNSIID aIID, void** aInstancePtr) = 0;
863:04.79              ^
863:04.80 ../../dist/include/nsISupportsBase.h:63:25: note: unimplemented pure virtual method 'AddRef' in 'nsIDOMMozWakeLockListener'
863:04.80   NS_IMETHOD_(nsrefcnt) AddRef(void) = 0;
863:04.81                         ^
863:04.84 ../../dist/include/nsISupportsBase.h:72:25: note: unimplemented pure virtual method 'Release' in 'nsIDOMMozWakeLockListener'
863:04.85   NS_IMETHOD_(nsrefcnt) Release(void) = 0;
863:04.85                         ^
863:04.87 ../../dist/include/nsIDOMWakeLockListener.h:31:14: note: unimplemented pure virtual method 'Callback' in 'nsIDOMMozWakeLockListener'
863:04.89   NS_IMETHOD Callback(const nsAString & aTopic, const nsAString & aState) = 0;
863:04.89              ^
863:04.90 /Users/pale/jenkins/workspace/palemoon/widget/cocoa/nsAppShell.mm:305:1: error: C++ requires a type specifier for all declarations
[...]

PR #319, opened 2016-01-26
Revert Mac screensaver changes fixes -- they break the build

stevenschlansker:
Sounds like this is an important feature, but it breaks the build (see #301).  We should re-consider it once it compiles :smile: 

!!! before considering this PR, take a look at #321, which might be an actual fix rather than a revert!


stevenschlansker:
cc @trav90 @Axiomatic-

stevenschlansker:
Closed in favor of #321

PR #321, opened 2016-01-27
Try to fix Mac screensaver improvements, #301

stevenschlansker:
Should this be NS_IMPL_ISUPPORTS1?  Seems to compile now, at least

### DIFF
--- a/widget/cocoa/nsAppShell.mm
+++ b/widget/cocoa/nsAppShell.mm
@@ -302,7 +302,7 @@ nsAppShell::~nsAppShell()
   NS_OBJC_END_TRY_ABORT_BLOCK
 }
 
-NS_IMPL_ISUPPORTS(MacWakeLockListener, nsIDOMMozWakeLockListener)
+NS_IMPL_ISUPPORTS1(MacWakeLockListener, nsIDOMMozWakeLockListener)
 nsCOMPtr<nsIPowerManagerService> sPowerManagerService = nullptr;
 nsCOMPtr<nsIDOMMozWakeLockListener> sWakeLockListener = nullptr;
 


stevenschlansker:
If this PR is accepted, it obsoletes #319

stevenschlansker:
cc @trav90 @Axiomatic-

Axiomatic-:
Yeah, seems like this is the proper way (cannot give 100% assurance though) to my understanding, NS_IMPL_ISUPPORTSn, where n is the number of interfaces created. (This is a macro no?)

stevenschlansker:
I am making a build now and will be testing shortly.  I'm not sure exactly what to test, so as long as it compiles and doesn't crash quickly, I'm going to push for this instead of reverting.  Progress ;)

stevenschlansker:
Seems to work OK so far!

Axiomatic-:
The way to test this is set your screensaver to some low amount, then open an HTML5 video and play it. What should happen is the screensaver is inhibited, if the screensaver still shows up while the video is playing then it does not work.

wolfbeast:
Ping. 
I gather this works as-intended, then?

stevenschlansker:
The build works OK.  I have yet to test HTML5 video, because the browser crashes on youtube :(
So I'll follow up once I get that figured out.

trav90:
> The build works OK. I have yet to test HTML5 video, because the browser crashes on youtube :(

This is why the screensaver changes had to be backed out for Linux as well.

wolfbeast:
This didn't actually fix the crash issues, so closing this PR in lieu of a different solution or a backout PR.

PR #332, opened 2016-01-30
Fix version reading with new perl

cbehan:
Mozilla has started to adapt to this with bug 1231340.

### DIFF
--- a/config/milestone.pl
+++ b/config/milestone.pl
@@ -56,7 +56,7 @@ $MILESTONE_FILE  = "$TOPSRCDIR/config/milestone.txt";
 #
 my $milestone = Moz::Milestone::getOfficialMilestone($MILESTONE_FILE);
 
-if (defined(@TEMPLATE_FILE)) {
+if (@TEMPLATE_FILE) {
   my $TFILE;
 
   foreach $TFILE (@TEMPLATE_FILE) {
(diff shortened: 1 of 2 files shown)


cbehan:
The version that requires this is 5.22. But the patched file continues to work with old versions of perl.

wolfbeast:
What I meant to ask is how far back is this now-enforced syntax (in newer perl, although 22 seems to still only "deprecate" it according to the docs) available in perl? Some of the toolchains used for building Pale Moon use a relatively old version of perl and I'd rather not have to worry about forward compatibility resulting from having to upgrade at this point in time.

cbehan:
My 5.22.1 installation refuses to use the deprecated syntax. According to a blog (http://blogs.perl.org/users/rurban/2012/02/on-definedarray-and-definedhash.html), the syntax became equivalent in 5.16 p5p. But if you just use milestone.pl to print out the "2.0.0" string, a test I did with perl 5.08 still works.

wolfbeast:
Thanks for the article.
> defined(@array) and defined(%hash) have been deprecated in the docs since 5.6

This means it's not an issue since the oldest compatible toolchain uses 5.6.1, where it was already deprecated and thus the proposed syntax must have already been valid.

Issue #326, opened 2016-01-28
iMacros add-on broken. 


JHPArizona:
I have iMacros installed, version 8.8.2.1 which worked fine in Pale Moon 25.8.1 (x86) I have set up a lot of macros to handle repetitive steps. After upgrade to Pale Moon version 26, iMacros does not "Play" and every macro (I tried) produces error 
TypeError: Downloads.getList is not a function (Error code: -1001)

I have tried the macro with and without the XSS filter running. I tried upgrading to the lastest version of iMacros although I had read in another thread that 8.8.2.1 works versus anything higher is broken on version 25.8.1. The new version is 8.8.7.1. This version did not work either. When I get more time, I will try some of the other versions and see if I can make it work. NOTE: This is the free version from the company.

If you want any further testing, just let me know.
**Update 20160129:**
I have downgraded to 25.8.1 as I could not work as efficiently without iMacros. Reconfirm that it works fine in this version.

chaoskagami:
Leaving aside the fact that said addon is proprietary and I don't know how it ticks; I doubt it is working correctly in that version since we don't have an implementation of getList in Downloads.jsm and don't appear to EVER have had one. Firefox refactored Downloads.jsm and added that in the newer versions.

Aside from that; it's probably possible to make a shim for firefox compatibility. getList takes a single argument: Downloads.PUBLIC, Downloads.PRIVATE, or Downloads.ALL and returns the correct list for either normal, private, or both. We have different functions here, but a wrapper would be possible to implement in less than twenty lines.

@wolfbeast, would that kind of shim be accepted for compatibility with firefox? It should only require changes to Downloads.jsm and I can have it done rather quickly. Building it on the other hand is another issue, as well as testing it since both take forever on a dual core and I nuked my build directory to get back disk space.

chaoskagami:
I made a small changeset which partially imports a bit of Mozilla's Downloads.jsm, mostly the getList function, and redirected out our current functions to use what I imported in order not to break anything API-wise. After compiling Pale Moon functions as before this patch (at least on linux, but the changes are non-invasive anyways.)

https://github.com/chaoskagami/Pale-Moon/commit/365d1a116c12dbf997f8bd707bdbde6e524c3165

Submitting PR to master. Kind of code cleanup, and I'll probably want to port the summary stuff later since that also isn't terribly involved. This should avoid this issue, though I refuse to let iMacros touch my computer.

wolfbeast:
[...]

PR #335, opened 2016-01-31
Import some changes from Mozilla hg to Download.jsm in attempt to be compatible with newer API of firefox

chaoskagami:
In a nutshell; issue #326 is caused by lack of the newer Mozilla functionality in Downloads.jsm. There's no more separate functions to get different download lists there, but one function, getList, which fetches based on a parameter.

This PR partially imports the newer functionality from Mozilla's hg, the required jsm class to support it, and redirects the older names to what I added. I also added another function following our current naming convention; if you don't want that, I'll simply edit the PR (assuming you want it in the first place. ;P) It should change nothing functionality-wise, and everything still works correctly on my Linux box.

It should resolve the issue in #326 even though I'm not touching a proprietary addon with a 10-foot pole. I may import the other code (e.g. DownloadSummary) later but as we don't use it for anything in Pale Moon I didn't feel the need to import it yet.

edit: Actually had a mistake here, so I fixed it and squashed. Sorry.

### DIFF
--- a/toolkit/components/jsdownloads/src/DownloadList.jsm
+++ b/toolkit/components/jsdownloads/src/DownloadList.jsm
@@ -13,6 +13,7 @@
 
 this.EXPORTED_SYMBOLS = [
   "DownloadList",
+  "DownloadCombinedList",
 ];
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


tumpio:
This would also fix the Download Status Bar add-on. That although supports the old API is not properly checking Pale Moon 26 compatibility.

https://forum.palemoon.org/viewtopic.php?f=15&t=10861&p=75860

chaoskagami:
I'll test it once I rebuild PM again, if you'd like.

The only functionality truly added here is that one function, and what is required to support it. If the Download Status Bar addon uses anything besides that, yeah, it's probably still broken.

mattatobin:
I would be skeptical of any change here that breaks older functionality.

Mozilla style extensions are on their way out for pure Mozilla products. Potentally breaking stuff for OUR level of compatibility for the sake of changes made to a dead end code path is not the wisest thing in this case.

However, if these are additional to but do not affect the older and established functionality.. Then I am pretty ok with it.. provided they work.

wolfbeast:
This should keep full equivalent functionality for the shimmed functions, so it'd be OK to take. I'll have to be very careful what to allow because changing APIs around is tricky and will break internal functions in the browser if not done properly. We cannot lose any current functionality while adding more functions, and considering extensions for Firefox seems to be hurtling towards the abyss anyway, I'm not going to make too drastic changes just to satisfy a few extensions that don't have a Pale Moon equivalent (yet).

this will need thorough testing, regardless.

chaoskagami:
Agreed. I'd prefer it get well tested. The last thing I want is to break PM in pursuing addon compatibility. This shouldn't break or remove any existing functionality.

If I continued to develop a few more shims in the download API here, that wouldn't be a problem, correct?

wolfbeast:
> If I continued to develop a few more shims in the download API here, that wouldn't be a problem, correct?

No problem at all :)

chaoskagami:
So i just realized I made a mistake that I didn't catch because I'm an idiot.

@wolfbeast - The generator function in Downloads.jsm in getLine is throwing an error. We are working off ES6, right? Is that not enabled by default in internal jsm?

wolfbeast:
We don't have generator functions. Also, what you can use in jsms isn't necessarily the same as what you can use in page content (chrome code has its own implementation of Promises, for example).

I'll back this out.

PR #338, opened 2016-02-01
Add support for NTLMv2 (and use as default)

trav90:
This PR does the following:

Removes old LM code from the NTLM handler
Adds support for NTLMv2 and uses it by default
Adds a new pref ("network.ntlm.force-generic-ntlm-v1" defaulted off) for fallback to NTLMv1 if required

This resolves the issue mentioned on the forum, and has been confirmed working as intended by the OP:

https://forum.palemoon.org/viewtopic.php?f=37&t=10928

### DIFF
--- a/modules/libpref/src/init/all.js
+++ b/modules/libpref/src/init/all.js
@@ -1308,6 +1308,9 @@ pref("network.prefetch-next", false);
 //   [scheme "://"] [host [":" port]]
 // For example, "foo.com" would match "http://www.foo.com/bar", etc.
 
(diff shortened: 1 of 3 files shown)


wolfbeast:
Thanks! I already had part of the NTLM issues covered by disabling ntlmv1 in commit e4b5a4e8854891c99250ddc0e99a423ffbc1417a but adding formal support for NTLMv2 this way and restricting v1 further is very much appreciated :+1:

trav90:
You're welcome, glad to do it! :)

PR #349, opened 2016-02-05
Downloads API Shims, pt.2

chaoskagami:
This isn't every shim that can be possibly made for the newer API, but in it's current state I have tested this extremely well unlike last time. It unbreaks a portion of addons (not all, obviously.) It's in a state where I'm comfortable with merging since it doesn't break anything.

The major problem with last attempt was Mozilla's API has changed so severely I really can't import even tiny bits of their code without major problems. I have to cut corners, unless I want to import newer mozilla core JSM (which isn't going to happen for obvious reasons. ;P)

Commit message is pretty descriptive, and I've made an effort to document as well as possible inside the source itself. Most of the changes were implemented from scratch to mostly follow Mozilla's docs.

There are a few caveats, obviously. Downloads.ALL won't retrieve both public and private downloads. This requires importing DownloadCombinedList, which for multiple reasons can't possibly work as-is.

I'll still probably work on this further, but it should be safe to merge what I have done.

EDIT: https://addons.mozilla.org/en-US/firefox/addon/download-flash-and-video/ which was posted to #326 is made functional by this PR, for example - probably owing to it using only the common functions I implemented. It doesn't show up in the Downloads List for some reason, though.

### DIFF
--- a/toolkit/components/jsdownloads/src/DownloadCore.jsm
+++ b/toolkit/components/jsdownloads/src/DownloadCore.jsm
@@ -62,6 +62,8 @@ XPCOMUtils.defineLazyModuleGetter(this, "Promise",
                                   "resource://gre/modules/commonjs/sdk/core/promise.js");
 XPCOMUtils.defineLazyModuleGetter(this, "Task",
                                   "resource://gre/modules/Task.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "DownloadsCommon",
+                                  "resource://gre/modules/DownloadsCommon.jsm");
 
 const BackgroundFileSaverStreamListener = Components.Constructor(
       "@mozilla.org/network/background-file-saver;1?mode=streamlistener",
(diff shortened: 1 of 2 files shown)


wolfbeast:
So did you map downloads.ALL to be the public list then? that's what I would do ;)
I'll merge this in in a bit after I've looked over the code changes for a basic review for glaring mistakes you may potentially have missed.

Thanks very much for your work on this. Using shims for added compatibility but not trying to obtain parity or duplicate the moving Mozilla target is the best we can do. We have to be realistic in these endeavors in the fact that current "current" or "latest" Firefox extensions are on a dead-end road to be stopped when (not if) Mozilla drops current extensions (xul/bs/sdk) on the floor -- which they will. This is a realization extension developers also have to have, and aiming to actually become compatible with Pale Moon natively should be the path to go if they care about their work being used in the future instead of ending up in the dirt.

wolfbeast:
LGTM

Issue #27, opened 2015-02-14
Exclude pinned tabs from alltabs view


wolfbeast:
Should be behind a pref, disabled by default. browser.alltabs.hide_pinned or something
Filter on pinned state and exclude from tab list array. Add null check. Enh.: Switch to pinned tabs display if list empty (only pinned tabs present)

wolfbeast:
I've made a build with the current state of #362 and tested it. We're getting there ;)

Unfortunately this also causes issues:
- When you unpin a tab, the unpinned tab in the panel is put at the very end of the display while the actual tab is at the start of the tab strip (out of order)
- When you close the last "normal" tab, leaving you with just pinned ones, the panel doesn't open on first click (and there's an error in the error console) and the display of the tabs gets screwed up in the panel afterwards. I verified that this is not a separate bug; it's caused directly by these changes.
- When unpinning a tab after having just pinned tabs left, one of the pinned tabs vanishes (becomes hidden and can't be recovered). After this occurred I've also experienced a hard browser hang hitting ctrl+tab.

So, I think there's something that screws up the entire tab list here. Perhaps you need to throw the whole list away when pinning/unpinning and make it completely rebuild it, instead.

PR #362, opened 2016-02-13
Option for excluding pinned tabs from list

cbehan:
This attempts to fix issue #27. One bogus commit is before the real one because GitHub does not have a button to update what commit a fork is based on.


wolfbeast:
It's a smart bit of logic, but it doesn't seem to work. I applied this patch and it seems to have absolutely no effect -- pinned tabs are still listed in the pane (yes I flipped the pref).

wolfbeast:
It seems it needs a bit more work. If you pin a tab, it's just move to the first position in the tab previews panel, but isn't hidden. I tested this the following way:

1. pinned two tabs (a, b)
2. opened alltabs: they weren't hidden but positioned as the first two
3. exited the browser, restarted
4. opened alltabs: pinned tabs were now hidden
5. pinned another tab (c)
6. opened alltabs: a and b remained hidden, c was moved to the first position but visible
7. exited the browser, restarted
8. opened alltabs: a,b, and c were now properly hidden

So it seems it currently only works when the tab panel is first initialized with **existing** pinned tabs.

cbehan:
Thanks for testing. I just realized that this depends on browser.allTabs.previews. When this is true, I see behaviour just like what you described. I had it set to false before. And in that case everything seemed to work with the barebones menu listing tabs.

wolfbeast:
I've made a build and tested it. We're getting there ;)

Unfortunately this also causes issues:
- When you unpin a tab, the unpinned tab in the panel is put at the very end of the display while the actual tab is at the start of the tab strip (out of order)
- When you close the last "normal" tab, leaving you with just pinned ones, the panel doesn't open on first click (and there's an error in the error console) and the display of the tabs gets screwed up in the panel afterwards. I verified that this is not a separate bug; it's caused directly by these changes.
- When unpinning a tab after having just pinned tabs left, one of the pinned tabs vanishes (becomes hidden and can't be recovered). After this occurred I've also experienced a hard browser hang hitting ctrl+tab.

So, I think there's something that screws up the entire tab list here. Perhaps you need to throw the whole list away when pinning/unpinning and make it completely rebuild it, instead.

wolfbeast:
Added some comments on the actual issue instead of this PR.

wolfbeast:
Closing this PR since it hasn't seen any activity for close to 2 months and the code doesn't actually work as-intended.
Please create a new PR if you've had time and opportunity to work on this some more.

PR #365, opened 2016-02-16
Make sure the tab scrollbox arrows are only displayed when applicable

Antonius32:
Currently, when the tabs are on top, the tab scrollbox arrows appear as soon as the Tab bar is filled up with tabs, but there's no tab overflow at that point. This pull request makes sure that the tab scrollbox arrows are only displayed when they're applicable, i.e., when the tabs are actually overflowing.


wolfbeast:
Is this for the problem displayed here?
https://forum.palemoon.org/viewtopic.php?f=29&t=11223

Antonius32:
Yes, it should resolve that problem as well.

PR #372, opened 2016-02-19
Resolve issue #369

### LINKED ISSUE #369: 26.1.0 Downloads.createDownload crash
Since the 26.1.0 update, the `Downloads.createDownload` function from Downloads.jsm has begun failing.
This is because `_backendDownload` in Downloads.jsm has an argument named `aProperties`, but tries to access it as `aOptions`; additionally, it assumes `aProperties`\\`aOptions` is not `null`, when in this case it is.
It seems to be a simple fix, renaming `aProperties` to `aOptions` and setting it to an empty object if it's `null`.

Additionally, while this seems to be a fix, I'll mention it anyway. When passing a `nsIFile` to `createDownload` as the target, you previously had to pass `target: { file: nsiFile }`. Now it is simply `target: nsiFile`. The latter seems to be correct according to [Mozilla](https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Downloads.jsm#createDownload%28%29), but it is a breaking change.

chaoskagami:
Okay. This should fix #369's crash, and also fix the accidental API breakage, which was completely unintentional.

This'll teach me to copy code from other functions that have different parameter names. :/

### DIFF
--- a/toolkit/components/jsdownloads/src/Downloads.jsm
+++ b/toolkit/components/jsdownloads/src/Downloads.jsm
@@ -95,26 +95,45 @@ this.Downloads = {
       download.source = new DownloadSource();
 
       if (aSource instanceof Ci.nsIURI) {
-        source.uri = aSource;
+        // The source is an nsIURI, directly assign it (FF 26+)
+        download.source.uri = aSource;
       } else if (typeof aSource == "string" ||
                  (typeof aSource == "object" && "charAt" in aSource)) {
+        // The above check is an implementation of isString which isn't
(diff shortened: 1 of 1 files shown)


chaoskagami:
I'm now pretty much done with identifying every possible error condition. It *should* be good for merge now, unless you or anyone spots anything immediately off.

wolfbeast:
Thanks, merging this in -- let's hope this is now complete!

chaoskagami:
Let's hope. if not, I'm around to fix it up again (but let's hope that no more weird conditions can be hit.)

:D

PR #380, opened 2016-03-05
branding/official: add a .desktop entry file

MingcongBai:
Here is a simple .desktop entry file for Pale Moon, could be useful for packagers.


wolfbeast:
Could be useful, with one nit: opening a new tab should not default to about:newtab, but to whatever the user has set to be the new tab setting. If that can't be done (by leaving out the URL or similar) then it should default to `http://start.palemoon.org/`

MingcongBai:
Leaving it blank will result in making a new window instead of a new tab. I will point that to ```http://start.palemoon.org/```.

wolfbeast:
Okay! - Thanks.

MingcongBai:
Thanks mate.

LimboSlam:
One question, why does the new tab page start with http instead of https? Yeah sometime ago start.me defaulted to https in the beginning of their URL. So @wolfbeast why isn't this the default for http://start.palemoon.org/? 

Just a question.

MingcongBai:
Hmm. I did not notice that...
2016年3月6日 12:08，"LimboSlam" <notifications@github.com>写道：

LimboSlam:
@MingcongBai: Yeah it automatically redirects to https, as you see here: https://palemoon.start.me/start.

MingcongBai:
Let me start a new PR...

wolfbeast:
It's HTTP because it is forwarded, and not having to negotiate a TLS connection before forwarding is a (very) noticeable win in speed. In addition, https://start.palemoon.org was not a valid address re: presented certificates (I've just made that possible with a new cert).

If you're planning on making a PR pointing directly to the start.me domain, please don't - I'll reject it because the initial link from browser default settings should _not_ point to an external party directly that is out of my control. Start.me has been a good service provider so far but we have a flexible agreement that can be terminated by either side for disagreement reasons.

PR #382, opened 2016-03-06
branding/official: use https for new-tab

MingcongBai:
As suggested by @LimboSlam.


wolfbeast:
Please see my comment about about speed of redirect in the other PR.
Also, the http redirect is done at the cloud edge, and https will make an origin server connection before redirecting.

wolfbeast:
I've made some changes at the CDN and origin server to make this work. Please check whether https is still causing a noticeable delay or not.

Issue #367, opened 2016-02-17
26.1.0 tab bar scrolling buttons displayed even if unncecessary


wolfbeast:
> I have set browser.tabs.tabMinWidth to 1 and browser.tabs.tabMaxWidth to 180.
This means that tabs are never scrolled, and scrolling buttons are never needed.
Yet they are still displayed (and non-functional) in PM 26.1.0

![pm3](https://cloud.githubusercontent.com/assets/3359132/13109077/6c16f904-d577-11e5-9dcd-926367989f36.png)

wolfbeast:
This isn't yet solved, despite #365 - one the tab bar overflows in maximized mode, the arrow scroll boxes remain in place.

Antonius32:
> This isn't yet solved, despite #365 - once the tab bar overflows in maximized mode, the arrow scroll boxes remain in place.

I'm not sure what you mean, can you clarify? The patch seems to work for me..

wolfbeast:
If I maximize a window, with tabs on top, and add tabs until the tab bar overflows (scroll arrows appear) then closing tabs will not remove them again (it never triggers an underflow).

wolfbeast:
Correction, it does trigger an underflow, but only if it reaches max tab width.

wolfbeast:
I think I know what's going on here -- not necessarily a styling issue.

wolfbeast:
Yup, figured it out. 
If set to use resizing immediately, it displays this behavior -- which makes sense for the logic involved.

Antonius32:
Ah, I see what you mean, now.

I've done some testing, and I am positive that this is in fact a styling issue; it is caused by the negative side margin on the tabs that was introduced by PR #317 (commit a6c16f0). And I've found that another issue, regarding the tabs' resizing behavior when browser.tabs.resize_immediately is set to "true", is caused by this negative side margin as well. For reference, see [this](http://forum.palemoon.org/viewtopic.php?f=3&t=11419) topic on the forum.

I think we may end up having to remove the transparent outer border of the tabs, after all. In that case, we'll just have to live with the fact that when the tabs are on top, they are "glued" right up against the top edge of the screen when the window is maximized or in fullscreen mode...

Antonius32:
> And I've found that another issue, regarding the tabs' resizing behavior when browser.tabs.resize_immediately is set to "false", is caused by this negative side margin as well.

[...]

PR #383, opened 2016-03-06
Simplify CSS for the tabs and the New Tab button (default Windows theme)

Antonius32:
This pull request significantly simplifies the CSS for the tabs and the New Tab button in the default theme for Windows. As an additional benefit, the new code allows the user to customize the width of the tabs with the aid of the Tab Utilities extension, even when the tabs are on top (with the current code, that's not possible). Tested and confirmed working on Windows 7.

This is a follow-up to PR #317 (commit a6c16f0).


wolfbeast:
Please retain the Fitts' Law remark in the appropriate place and any other applicable comments.

wolfbeast:
I'm merging this in and will correct the comments myself, trying to fix #367 is better done with this simplification in place.

PR #401, opened 2016-03-19
Add GStreamer 1.x Support

trav90:
This PR will add GStreamer 1.x support when Pale Moon is compiled with the -enable-gstreamer=1.0 option in mozconfig.

I've been doing some testing, and it all seems to be working. We've also received reports on the forum as well that it is working.

Additionally, I've been testing gst0.10 support, and have not seen any regressions.


wolfbeast:
Fantastic! Hats off to you, Travis :)

trav90:
Thanks!

MingcongBai:
Awesome! Will this make into the 26.2 release? If so, when?

trav90:
Depends on when 26.2 is ready. There's a few more issues I am looking into to see if they affect our code or not, and then we need to get a working build in our CentOS 6 environment before this will be shipped by default.

PR #406, opened 2016-03-22
Stop GStreamer from treating an EOS of one stream as EOS of all streams.

trav90:
Tested and seems to work as-intended with both 0.10 and 1.x.

This patch also appears to resolve the issue that was reported here: https://forum.palemoon.org/viewtopic.php?f=37&t=10074


wolfbeast:
Thanks, can you nudge the forum about this?

trav90:
Will do.

PR #408, opened 2016-03-26
libvorbis: svn r18077 -> r19464

Axiomatic-:
Tested on: http://hpr.dogphilosophy.net/test/index.php and working correctly. (Linux)

This also brings along the includes needed for Vorbis encoding, but not actual Vorbis encoding support, which is [BMO 883749](https://bugzilla.mozilla.org/show_bug.cgi?id=883749).


wolfbeast:
Verified this still works on Windows after the update as well. Merging in.

PR #421, opened 2016-04-17
Set execute attributes on .sh files in tree

trav90:
Setting the correct attributes on all *.sh files in the tree has been reported to fix some build issues (including jemalloc on Gentoo).


wolfbeast:
Fallout from this originally being a Windows-only project...

PR #435, opened 2016-05-16
Colorize the Win10 titlebar

Lootyhoof:
This removes the `background-color` we currently use on Win10, and exposes the system's accent colour for the background. It also removes the browser-drawn caption buttons when not in lwtheme mode (note that the close button is still drawn when hovered). It introduces a more pleasant hover animation for the caption buttons, too. There are no longer any styles to differentiate between `default-theme` and non-default.

Note that the styles that are no longer used have simply been commented out, for future use.

### DIFF
--- a/browser/themes/windows/browser-aero.css
+++ b/browser/themes/windows/browser-aero.css
@@ -160,6 +160,7 @@
 
 @media (-moz-windows-compositor) {
   #main-window {
+    background-color: transparent;
     -moz-appearance: -moz-win-glass;
   }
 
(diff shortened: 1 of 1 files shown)


wolfbeast:
Thanks!
I'll probably just remove the commented-out lines while I fix a few typoes I noticed in surrounding code; we have the commits to reverse if needed in the future (when Microsoft decides they want to change it up yet again) and just removing the lines will keep the overall code cleaner and easier to follow.

PR #436, opened 2016-05-17
Add a global class for close icons

### LINKED ISSUE #350: Provide global .close-icon class for close buttons
In an effort to clean up themes, as well as make it easier to make changes for close buttons (for example, if you wanted to provide various versions, such as HiDPI versions), I propose all close buttons (i.e. those which currently reference `close.png`) should have an additional `.close-icon` class added to them. This way, we can move all the current close buttons styles (of which are all identical, save the element they are styling) into `global.css` and simply style the `.close-icon` class.

Theoretically, this could also be of benefit to extensions too should they wish it, as they could include some `.close-icon` elements in their extension, with the theme providing native styling.

BMO: [851001](https://bugzilla.mozilla.org/show_bug.cgi?id=851001)

Lootyhoof:
This adds the XUL class `close-icon` to close buttons across the browser, allowing for them to be styled only once, in `global.css`. As a result, individual stylings have, for the most part, been removed to save space. 

For the Linux theme, to try to keep things as uniform as possible while still using GTK icons, all but the close button on the alert window (`toolkit/themes/linux/global/alerts/alerts.css`) are moved to `global.css` like the Windows and OS X themes.

This will, in future, potentially allow for much easier application of different close images (see OS X theme, which has a HiDPI version present). Note that inverted close icons are still styled in their original respective stylesheets, as not all close icons can be inverted this way for lwthemes.

Fixes #350.

### DIFF
--- a/browser/base/content/browser.xul
+++ b/browser/base/content/browser.xul
@@ -264,7 +264,7 @@
       <stack id="allTabs-stack">
         <vbox id="allTabs-container"><hbox/></vbox>
         <toolbarbutton id="allTabs-tab-close-button"
-                       class="tabs-closebutton"
+                       class="tabs-closebutton close-icon"
                        oncommand="allTabs.closeTab(event);"
                        tooltiptext="&closeCmd.label;"
                        style="visibility:hidden"/>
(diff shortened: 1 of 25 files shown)


wolfbeast:
LGTM! Thanks :)

Issue #219, opened 2015-11-15
Detect whether the current Windows theme is dark


Lootyhoof:
On Windows 8 (and now Windows 10), the system colours can get quite dark. This can mean that dark icons are placed on a dark window frame, making things fairly hard to see. If we had a way to detect via CSS whether the window frame is dark, we could style things accordingly. This would also work for lwthemes.

BMO: [940393] (https://bugzilla.mozilla.org/show_bug.cgi?id=940393), [988191] (https://bugzilla.mozilla.org/show_bug.cgi?id=988191), [1012629] (https://bugzilla.mozilla.org/show_bug.cgi?id=1012629), [1008134] (https://bugzilla.mozilla.org/show_bug.cgi?id=1008134), [1061947] (https://bugzilla.mozilla.org/show_bug.cgi?id=1061947).

wolfbeast:
940393 is the main implementation so that's what needs focus (specifically the part setting darkwindowframe). Immediately weave in the algorithm from 988191 when doing so.
Theme changes can be built for our specific theme on top of that, and small bugfixes implemented from the other related bugs, after that.

chaoskagami:
This issue also applies on Linux when a dark gtk theme is used, fyi. I've had to resort to a modified chrome/userContent.js for a veeeeery long time to fix this one. If there's a way to do this cross platform, I'd be quite happy.

wolfbeast:
This should be given some priority so setting this for the 26.1 milestone.

Lootyhoof:
Note that the implementation linked above is Windows 8-specific, and will not work correctly on Windows 10. The only Mozilla-side bug tracking this is [1192622](https://bugzilla.mozilla.org/show_bug.cgi?id=1196266), and it's been labelled as low-priority with no work having been done to it yet.

It should however be noted that, in particular, the Windows 8 implementation looks for the `ColorizationColor` registry key, in `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\DWM`, however Windows 10 also appears to have a `AccentColor` key in this location. Both appear to change when the accent colour changes, though have different values.

wolfbeast:
Apart from actually using this in the theme for e.g. inverted icons, the detection routine is in and seems to work as-intended with commit 07e3c01f5011974fdbd2f59cd37930d062b18ed3

wolfbeast:
[...]

PR #437, opened 2016-05-18
Infer from each toolbar's text color whether we should use inverted icons

Lootyhoof:
This commit adds the `brighttext` attribute to `toolbar` when the text colour is bright (such as when a lwtheme is applied). As such, this is a way to detect dark themes cross-platform, as well as reducing the amount of styles needed to style for lwthemes and Windows-specific styling. 

Addendum to #219.

### DIFF
--- a/browser/base/content/browser-fullScreen.js
+++ b/browser/base/content/browser-fullScreen.js
@@ -589,6 +589,8 @@ var FullScreen = {
       navbar.appendChild(fullscreenctls);
     }
     fullscreenctls.hidden = aShow;
+	
+    ToolbarIconColor.inferFromText();
   }
 };
 XPCOMUtils.defineLazyGetter(FullScreen, "useLionFullScreen", function() {
--- a/browser/base/content/browser.js
+++ b/browser/base/content/browser.js
@@ -980,6 +980,7 @@ var gBrowserInit = {
     gPrivateBrowsingUI.init();
     TabsInTitlebar.init();
     retrieveToolbarIconsizesFromTheme();
+	ToolbarIconColor.init();
 
 #ifdef XP_WIN
     if ((window.matchMedia("(-moz-os-version: windows-win8)").matches ||
@@ -1399,6 +1400,8 @@ var gBrowserInit = {
     TabsOnTop.uninit();
 
     TabsInTitlebar.uninit();
+	
+	ToolbarIconColor.uninit();
 
     var enumerator = Services.wm.getEnumerator(null);
     enumerator.getNext();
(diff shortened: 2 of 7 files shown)


wolfbeast:
Looks good, but one question: You're calculating the luminance independently of what was recently added for the dark window frame - does the test have the same flip point?

Lootyhoof:
As far as I'm aware, this algorithm is pulled directly from `LightweightThemeConsumer.jsm`, so anything that applies there will also apply here.

wolfbeast:
What threw me off was that you said it was an addendum to #219 -- but actually it isn't, since this just looks at the existing text color from LWTs...

Also, I don't see how this reduces the amount of styles needed -- it just exchanges `:-moz-lwtheme-brighttext` suffixes with a `toolbar` element preselector...? if anything it makes it more ambiguous what people should use...?

Or stated another way: why do we actually need this?

Lootyhoof:
>What threw me off was that you said it was an addendum to #219 -- but actually it isn't, since this just looks at the existing text color from LWTs...

Please understand that while this uses a similar algorithm to lwthemes, it is NOT the same thing, primarily due to the method of execution. While the current algorithm decides whether we are using `-moz-lwtheme-brighttext` when a lwtheme is applied, this will instead check if the **system theme itself** is dark, by checking the text colour **at all times**. This is different to (and complements nicely) `darkwindowframe`, as that is Windows specific, and still useful for titlebar elements that don't exist within a `toolbar`.

>Also, I don't see how this reduces the amount of styles needed

This was my bad, sorry about that - I did make some large changes but changed them at the last minute as they were a little too heavy-handed.

>if anything it makes it more ambiguous what people should use...?

Think of it this way:

When styling **chrome**, use `-moz-lwtheme`, as we surely don't need different toolbar stylings for a dark theme but may be desirable for a dark lwtheme.
When styling **icons**, use `toolbar[brighttext]`. If a dark theme is present we might lose legibility of the icons, so we can replace them with inverted (or otherwise brighter) icons in this instance. This also includes dark lwthemes.

wolfbeast:
I think I get it, but won't this cause issues with LWTs then? I mean, you'd be looking at 2 different ways to determine what icons should be used. This means that this should *only* be used is you're not using a lightweight theme, i.e.: when the default theme is in use, otherwise you can get inverted icons on a light LWT just because the chose system theme is dark, and the other way around; it would override LWT icon styling...

Lootyhoof:
This applies to **both** dark system themes and dark lwthemes. The distinction is that, while `toolbar[brighttext]` can work for both, `-moz-lwtheme-brighttext` only works for dark lwthemes. Both of these have their uses.

Note that when a **light** lwtheme is applied, `toolbar` doesn't report as having `brighttext`, so this is a non-issue. It only reports `brighttext` when the text itself on the toolbar in question is bright.

wolfbeast:
Oh, right, I missed the lwt text observer that was added. My bad. :)

PR #438, opened 2016-05-21
Use correct close-inverted.png on OS X

Lootyhoof:
Fixing some issues with the OS X theme that I hadn't noticed I made in #436, specifically:

- make sure there's a close-inverted for normal DPI
- use the HiDPI close-inverted icons for HiDPI

Sorry about that!


wolfbeast:
Good catch!

PR #439, opened 2016-06-03
Various theme-related improvements

Lootyhoof:
This PR contains the following:

- The window title on Windows 8/10 for dark window frames was initially set whenever `darkwindowframe` is active, however it is still active when the window is inactive, making the title difficult to see. I've set it to only be white when the window is active.
- The previous `darkwindowframe` style for making the text white applied to only the menubar and tabbar; I've made it also enable when `tabsontop=false`, as there is still glass in the `#nav-bar` too.
- Corrected the mappings of `close-icon` for non-selected tabs.

### DIFF
--- a/browser/base/content/browser-title.css
+++ b/browser/base/content/browser-title.css
@@ -189,7 +189,7 @@
       top: -5px;
     }
     
-    #main-window[darkwindowframe="true"]::after {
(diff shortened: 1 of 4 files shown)


wolfbeast:
Sorry for letting this PR hanging, been focused on trying to fix some major issues in Tycho.

PR #441, opened 2016-06-12
Windows 10 theme changes

Lootyhoof:
In an effort to make the Windows 10 styling look more system-native, the following changes were made:

- Window title on dark windows frames is now not used in lwtheme mode. It was ignoring the styles the lwthemes wanted.
- Altered the close button animation. The most noticeable change here will be the opacity change when clicked - this is to mirror behavior in Windows 10.
- Bright caption buttons (for dark window frames) had the exact opposite issue of the window title above: it was working for lwthemes, but not inactive windows. I also lowered the brightness slightly to match the style used in the system.

### DIFF
--- a/browser/base/content/browser-title.css
+++ b/browser/base/content/browser-title.css
@@ -189,7 +189,7 @@
       top: -5px;
     }
     
-    #main-window[darkwindowframe="true"]:not(:-moz-window-inactive)::after {
+    #main-window[darkwindowframe="true"]:not(:-moz-window-inactive):not(:-moz-lwtheme)::after {
       /* Dark window frame/accent color on Win 8/10 */
       color: white;
     }
--- a/browser/themes/windows/browser-aero.css
+++ b/browser/themes/windows/browser-aero.css
@@ -174,7 +174,7 @@
 /* ==== Windows 10 styling ==== */
 
   @media (-moz-os-version: windows-win10) {
-    /* Draw XUL caption buttons Win10 in lwthemes */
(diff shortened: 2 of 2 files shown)


wolfbeast:
Native windows 10 has a brief transition on hover. This is why I made a transparent element as a click target with opacity change for hover with a short transition period to mirror native behavior. You're now doing away with all that?...

Lootyhoof:
In the current implementation, the opacity set to 0 makes it misbehave with lwthemes.

Caption buttons

The implementation in this PR retains the desired animation, while requiring less styles to do so.

With regards to things not looking right in certain configurations, this was bound to happen regardless of whether the background turns transparent on click anyway. If there's a misalignment, it'll be quite obvious with the other caption buttons too.

Misalignment

wolfbeast:
What I eant with misalignment is the fact tha the X is actually drawn with our SVG any background transparency if the system "X" is misaligned will cause blurring or ghosting.

Also, previously, I'm quite sure the LWT remained untouched; i guess with the compacting of style rules something specific got dropped or overridden in previous patches then 

I'll merge this in and have another look myself

wolfbeast:
Your PR doesn't fix the LWT issue.
Also, the close button lost its fade like I said; that's not good. it also flickers because of the on-the-fly loading of an image (which is also more expensive to do than just opacity compositor changes).

wolfbeast:
I've compared the close button and it indeed goes transparent now (didn't do so in previous versions of Win10, I think) so your transparency is OK. I just hope it won't misalign on user's systems.

PR #449, opened 2016-06-27
Add VPX_ prefix where it's missing to fix build

Pulfer:
In some places of code IMG_FMT_I420 wasn't changed to VPX_IMG_FMT_I420, PLANE_V to VPX_PLANE_V etc.

This commit alone doesn't fix building Pale Moon with WebRTC support, there are other build issues (missing "-I/usr/include/pixman-1" flag etc).


wolfbeast:
@mattatobin -- You pretty much own the VPX stuff, can you look at this?

mattatobin:
WebRTC is very bitrotted and is likely not compatible anymore with the newer libVPX code. Also, our WebRTC is not spec compliant by any means.

Recommending RESOLVED WONTFIX

wolfbeast:
Agreed. closing.
WebRTC isn't going to ever be usable in our current iteration, and any patches against it would be a waste of time. Please don't put any more time into this.

PR #453, opened 2016-06-29
Add user agent overrides for base.consultant.ru (and other UA overrides)

Pulfer:
1. docs.google.com complains about old browser being used (at least when opening a spreadsheet).

2. base.consultant.ru (one of the most popular Russian laws and regulations databases) rejects to show any documents. It suggests to use "modern" browser instead.

### DIFF
--- a/browser/branding/official/pref/firefox-branding.js
+++ b/browser/branding/official/pref/firefox-branding.js
@@ -91,7 +91,7 @@ pref("general.useragent.override.icloud.com","Mozilla/5.0 (Windows NT 6.1; WOW64
 
 // UA-sniffing domains that are "app/vendor-specific" and don't like Pale Moon
 pref("general.useragent.override.web.whatsapp.com","Mozilla/5.0 (Windows NT 6.3; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/37.0.2049.0 Safari/537.36");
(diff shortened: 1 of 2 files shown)


wolfbeast:
We already have an override for google.com (which should include any subdomains automatically) and that has a version 45.9 which is more modern than your override for docs.* -- So this makes very little sense? I'd expect 38.9 to throw up a "too old" complaint because it's no longer the current Firefox ESR.

Also, the consultant.ru override should not be in the section you placed it -- that is for app/vendor specific sites that require a completely different (non-mozilla) UA.

Pulfer:
Looks like the issue with docs.google.com was caused by unofficial branding having UA rv:31.9 for google.com.

Pulfer:
BTW, is there a way to set UA overrides for all google.* domains other than listing them all? google.com redirects to country-specific domains like google.ru and shows infamous "Black bar" interface.

wolfbeast:
> BTW, is there a way to set UA overrides for all google.* domains other than listing them all? google.com redirects to country-specific domains like google.ru and shows infamous "Black bar" interface.

Nope, it's not possible to use wildcards for the TLD -- this is top-down functionality. The only way this can be done for all google domains is to either have google stop doing it wrong, or make entries for each and every TLD they use -- which I don't want to do, because the SSUAO list should remain short (it's a performance drain if it gets too long)

PR #457, opened 2016-07-09
Change default MOZ_OPTIMIZE_FLAGS for GCC to -O2

trav90:
https://forum.palemoon.org/viewtopic.php?f=37&p=88000

I'm currently re-building Pale Moon 26.3.3 + this patch and will post results later, but should be safe to merge in.


trav90:
The tarball for PM 26.3.3 without -Os is only 2.6MB larger. Build appears to be stable.

PR #463, opened 2016-07-24
Enable Pale Moon to be built with GCC 6.x

trav90:
This PR enables GCC 6 to build Pale Moon. Detailed changes:

Our STL wrappers includes mozalloc.h, which includes stdlib.h, which in libstdc++ >= 6 is now itself a wrapper around cstdlib, which circles back to our STL wrapper. With this PR mozalloc.h is no longer included in the cstdlib, allowing GCC 6 to compile Pale Moon successfully (although the browser segfaults immediately after opening).

GCC 6 also fails to build Graphite, but using cmath as opposed to math.h in Collider.cpp resolves the issue (this has been [fixed upstream](https://github.com/silnrsi/graphite/commit/fc73a64f5f86bb00ae082a7991497cbc392bf3fd) as well).

GCC 6 made several breaking changes to the way it optimizes some code. By passing the extra flags in this PR when using GCC 6, it resolves the immediate segfaulting (see above).

While far from "supporting" GCC 6 (for what it's worth Mozilla does not fully support it either), this will at least allow it to compile a working browser. I can confirm it works as intended, and does not break GCC 4.9 builds.

### DIFF
--- a/config/gcc-stl-wrapper.template.h
+++ b/config/gcc-stl-wrapper.template.h
@@ -22,6 +22,11 @@
 #define NOMINMAX 1
 #endif
 
+// Don't include mozalloc for cstdlib. See bug 1245076.
+#ifndef moz_dont_include_mozalloc_for_cstdlib
+#  define moz_dont_include_mozalloc_for_cstdlib
+#endif
+#ifndef moz_dont_include_mozalloc_for_${HEADER}
 // mozalloc.h wants <new>; break the cycle by always explicitly
 // including <new> here.  NB: this is a tad sneaky.  Sez the gcc docs:
 //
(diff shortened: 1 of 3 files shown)


wolfbeast:
There's a problem with replacing `math.h` with `cmath` -- one of the things that always needs doing in updating graphite is replacing it in the other direction or it breaks (at least Windows) builds (see readme.mcp in the graphite folder). Now I haven't attempted it with the current code but I'm pretty sure it doesn't build on VS2012.

Possible workarounds:
- `#ifdef` the included library, or (better)
- Upstream pull 1.3.8 (or maybe 1.3.9 which seems release-ready) since it has collider changes...

trav90:
Looks like updating Graphite to v1.3.9 (PR #464) solved the issue. GCC 6 builds Pale Moon successfully with graphite updated and without using cmath.

wolfbeast:
Great! Thanks, Travis :)
I'll make sure to fire off a build to check the Windows side on both repos.

wolfbeast:
Confirmed building and working, including graphite font test.

trav90:
Great! Unfortunately turns out we aren't quite there yet on Linux. For an unknown reason, GCC 6 can only build use math.h and compile successfully when using Link Time Optimization. Should be a simple fix, and will have another PR shortly.

PR #465, opened 2016-07-24
Force Link Time Optimization to be used with GCC 6

trav90:
Followup to PR #463 and #464.

GCC 6 successfully compiles Pale Moon, but only when Link Time Optimization is used.

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -1463,10 +1463,10 @@ else
     _DEFINES_CXXFLAGS='-DMOZILLA_CLIENT -D_MOZILLA_CONFIG_H_ $(ACDEFINES)'
 fi
 
-#FIXME: Work around breaking optimizations performed by GCC 6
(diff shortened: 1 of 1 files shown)


wolfbeast:
**Dangerous waters here.** Forcing an optimization to solve build issues means that the compiler has a bug. These kinds of optimizations normally depend VERY heavily on the state of the code and _unrelated_ changes may cause breakage.

Because this only applies to GCC 6, the added comments and FIXME: state are important, as is the fact that this should only ever apply to GCC 6 and nothing else.

trav90:
Agreed. GCC 6 is still not fully supported, and should be a use at your own risk/your mileage may vary option. But these fixes will improve (_but not guarantee!_) the chances of getting a working browser if you do decide to use GCC 6.

Also, once Tycho becomes our primary tree, LTO will no longer be required (so this is also more of a stop-gap, temporary measure).

PR #468, opened 2016-07-26
Backport some BSD fixes

jbeich:
A minimal set of fixes after which FreeBSD builds fine with empty `.mozconfig` using GCC. DragonFly needs `--with-system-nspr` and `--with-system-nss` or FreeBSD configs. Runtime tested on x86_64 while watching YouTube music videos (HTML5 H.264) with configured PulseAudio. NetBSD and OpenBSD are probably still broken.

All patches are originally from Firefox.


wolfbeast:
Ow, quite a bit of work there. Thanks for the effort! However, I'm not entirely sure if this is going to help in the near future since we're in the middle of re-basing and this needs to be re-ported to Tycho for BSD to work. I'll merge this in and do some try builds on windows to see if things are still working as intended (maybe more of a risk for Linux, @trav90 ?).

trav90:
Yeah, Linux is likely at a higher risk of bustage. I'll fire off a build shortly.

mattatobin:
I really don't like this use of case in configure.in..

I also don't like that these files
 +        #include <sys/types.h>
+        #include <netinet/in.h>
+        #include <arpa/nameser.h>
seem not to be inside the BSD ifdef

Using System NSS and NSPR are also not supported and superseded our own changes to NSS and NSPR

I also don't care for the change of MOZ_ENABLE_DBUS to NECKO_WIFI_DBUS..

trav90:
Doesn't look like this caused any issues as-is with Linux builds.

jbeich:
> I really don't like this use of case in configure.in..

Which one? I'd prefer to not copy paste my rationale from bugzilla.mozilla.org.

> I also don't like that these headers in configure.in
> sys/types.h
> netinet/in.h
> arpa/nameser.h
> seem not to be inside the BSD ifdef Are these headers applicable for non BSD?

According to Mozilla bug it's applicable to Linux, OS X, Solaris, QNX, IBM V5R1 and some but not all BSDs. Where res_ninit() doesn't exist (e.g. on Windows) the change is nop.

> Using System NSS and NSPR are also not supported and superseded our own changes to NSS and NSPR

NSPR/NSS upstream is slow in accepting Tier3 patches, so DragonFly is stuck in limbo for years. This is mostly because new platforms have to be vetted by maintainers instead of autodetected via autoconf.

> I also don't care for the change of MOZ_ENABLE_DBUS to NECKO_WIFI_DBUS..

Without it building on any BSD hits AC_MSG_ERROR() unless `--disable-necko-wifi` is added to `.mozconfig`.

mattatobin:
We have our own NSPR/NSS in tree that we have made our own modifications to. You may submit patches to us. Using system libs for NSPR and NSS negates our modifications.

NECKO_WIFI_DBUS breaks the general naming convention of MOZ_ENABLE/DISABLE_COMPONENT That is my only problem with it is..

wolfbeast:
Um.. yeah I guess I missed that. Using system NSPR/NSS will break some things in the security manager because we expect e.g. Camellia-GCM to be present in NSS and a system one won't have that. You have to use the in-tree one.

jbeich:
> configure.in lines now 5557 5589 and 8149 ..

Please, use inline comments instead.

> I would prefer if you used 'if test' logic instead of case..

I think, long `if test` conditionals are more error-prone. Tycho already inherited the change. Here's what you're proposing:

```sh
if test "$OS_TARGET" != "WINNT" -a "$OS_TARGET" != "Darwin" \
     -a "$OS_TARGET" != "Android" -a "$OS_TARGET" != "OpenBSD" \
     -a -n "$MOZ_CUBEB" -a -z "$MOZ_B2G"; then
```
or vet every new platform
```sh
if test \( "$OS_TARGET" = "Linux" -o "$OS_TARGET" = "GNU/kFreeBSD" \
     -o "$OS_TARGET" = "Bitrig" -o "$OS_TARGET" = "DragonFly" \
     -o "$OS_TARGET" = "FreeBSD" -o "$OS_TARGET" = "NetBSD" \
     -o "$OS_TARGET" = "OpenBSD" -o "$OS_TARGET" = "SunOS" \) \
     -a -n "$MOZ_CUBEB" -a -z "$MOZ_B2G"; then
```

> We have our own NSPR/NSS in tree that we have made our own modifications to. You may submit patches to us. Using system libs for NSPR and NSS negates our modifications. Which means inconsistent results for users when it comes to cert store and potential behavior.

Add more `AC_MSG_ERROR()` like you did in edd6e62. For now, `--with-system-nspr` works fine.

> NECKO_WIFI_DBUS breaks the general naming convention of MOZ_ENABLE/DISABLE_COMPONENT That is my only problem with it..

Originally, I've landed without `MOZ_` prefix to be consistent with `NECKO_WIFI`, the main option to disable Wifi geolocation. Tycho already inherited the change.

wolfbeast:
So you can't take responsibility by following instructions and everything has to be a hard error? o.O

I'm backing this work out of our stable tree considering it touches many things that may have subtler consequences we can't see right now with a few builds. It's best if you externalize BSD code changes as indicated in the Tycho issue that has the jemalloc discussion.

PR #469, opened 2016-07-28
Fix MOZ_NATIVE_JEMALLOC

jbeich:


wolfbeast:
Sorry, no. I can't accept this.
Old versions of system jemalloc are not supported. Using jemalloc3 implies using the in-tree version.

jbeich:
How old? FreeBSD 11.0 has jemalloc 4.2.1. This PR doesn't enable system jemalloc by default.

wolfbeast:
Then please provide a description what you're doing here and why.

jbeich:
If not obvious from reading commit messages and maybe following bug references:
- c9eccd1e4 fixes 4d9fe80 regression
- cece2ee fixes 9df3bfd regression

wolfbeast:
Considering ragequit and deleted repo, reopening is not possible.
Don't let the door hit you on the way out.

PR #475, opened 2016-08-06
Use -O1 optimization for kiss_fft lib (32bit builds)

trav90:
Followup to commit 96301f23 to work around OOM errors when compiling with 32bit GCC.


trav90:
Thanks. This workaround shouldn't be needed for Tycho, since it supports GCC 5.x much better than our current codebase does, and will likely be used for final release builds provided by me, thereby bypassing the problem.

If this does become an issue with Tycho for other people building from source (although I'd suspect most will be building 64bit and using something newer than CentOS 6), we can re-evaluate it at that time.

wolfbeast:
Sounds good! Thanks for the research :)

PR #483, opened 2016-08-19
Don't force Link Time Optimization with GCC 6.x

trav90:
White LTO was needed to get a successful build with GCC 6.0, it apparently breaks building with GCC 6.1.  :confounded:


wolfbeast:
IOW: GCC 6.* is unreliable.

PR #484, opened 2016-08-21
Build GStreamer 0.10 support by default again

trav90:
GStreamer 1.x support is causing way too many segfaults at present, and needs some serious TLC before we start using it again by default.

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -5584,7 +5584,7 @@ dnl = Enable GStreamer
 dnl ========================================================
 if test "$OS_TARGET" = "Linux"; then
   MOZ_GSTREAMER=1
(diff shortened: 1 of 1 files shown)


wolfbeast:
Do you think this needs a point release or not?

trav90:
Probably is a good idea. I've got one more incoming PR that I'm going to submit first.

wolfbeast:
As discussed with Tobin this point release should best be 26.4.0.1 instead of 26.4.1 to keep with the numbering scheme we've intended.

trav90:
Agreed. Final release builds will be 26.4.0.1

PR #485, opened 2016-08-21
Implement GStreamer blacklist (& crash fixes)

trav90:
This implements a blacklist for GStreamer (behind a pref). Older versions of the flump3dec plugin are known to be crashy, and the h264parser could be a potential sec issue with some GStreamer 0.10 versions, so are blacklisted.

This PR also fixes a couple other GStreamer 0.10 crashes


trav90:
I'm going to kick off some builds (26.4.1) with the patches in this PR and throw them to the beta testers.

PR #487, opened 2016-08-30
HSTS preload update

trav90:
This PR updates the URL to download an updated HSTS preload list from upstream, removes a couple ff references, and updates the HSTS preload list to current.


trav90:
Tag #62

PR #526, opened 2016-09-17
Remove Google as a search provider

trav90:
Noticed while packaging a build this morning that Google's search plugin is back after the Tycho merge.

This PR is simply a re-land of commit 82a5f08 after the Tycho merge.


wolfbeast:
Thanks!

PR #527, opened 2016-09-17
Change WIN_VER and _WIN32_WINNT to 0x0600 where applicable

Gpower2:
This time, I included a fix in util.h of nICEr module, for not breaking x86 builds.


media/mtransport/third_party/nrappkit/src/util/util.h
```diff
@@ -66,8 +66,12 @@ int nr_reg_uint8_fetch_and_check(NR_registry key, UINT8 min, UINT8 max, int log_
 
 #ifdef WIN32
 int snprintf(char *buffer, size_t n, const char *format, ...);
+// When using Windows Vista+ header files, inet_ntop is already defined 
+// with a different type modifier, so we don't redefine it
+#if defined(WIN_VER) && WIN_VER <0x0600
```
wolfbeast: I don't think this identifier is correct -- it's either WINVER or _WIN32_WINNT...

Gpower2: Oh shoot... You are correct! I fixed it, re-tested both builds and pushed the fix.

PR #530, opened 2016-09-18
Cleanup of Statusbar component

Lootyhoof:
This removes several redundant Australis-focused parts of code related to the statusbar component, ensuring we only have the code we need. It merges contents from `legacy.xul` and `prefsLegacy.xul` into their main components to reduce redundancy. It also moves button stylings for statusbar-related buttons into `browser.css` for each OS, both to appease `customizeToolbar.xul` and to keep button stylings in one place.


wolfbeast:
Thanks for tackling this! LGTM so I'll just merge this in.

mattatobin:
I am not quite happy with the intermingling of a component's style into browser.css in 1e31244a38539e4217a7cd50598fa1e9f8c9454d

The rest looks fine.. But I specifically kept everything from mixing in for the sake of ease. There is still a lot of future work that should go into the statusbar code and hiding away the css in the massive browser.css will simply make it harder to find later.

Is there some advantage to having this in browser.css that I am unaware of?

wolfbeast:
Actually, I had a look and because you're moving things from shared to individual skins it's actually adding redundancy. I'll revert the style changes for now, they are better separated if you're going to parse the statusbar/*.css anyway. There's no real win there.

Lootyhoof:
This was done because (from what I could see, anyway) `customizeToolbar.xul` wasn't able to see the icon stylings, so wasn't displaying icons for them. This was a (temporary?) hack to make them viewable. If you're able to view them without this, then fair enough.

wolfbeast:
How would I verify if this is the case?

Maybe it's an idea to keep the actual rules separated in the source but import the stylesheet into browser.css...? I'm fine either way with it; if it does cause missing icons if separated out then there's a good reason to adopt it into the main css file.

Lootyhoof:
You can verify by opening the Customize Toolbar window on a build after the statusbar merge. It certainly would be possible to import the CSS into `browser.css`; it's multiple means to an end, really.

mattatobin:
Well.. if you think it should be in browser.css.. Why don't we just remove the references for the css in jar.mn and stylesheet tags from the xul and just use the preprocessor to

`#include file`

If this is what you want to be done let me know and I will make the modifications.

PR #531, opened 2016-09-19
Media updates

trav90:
A pretty substantial patch, with the following highlights (see individual commits for more detailed info):

1. Updated the Media Decoder, the Media Decoder State Machine, AudioOffloadPlayer and Metadata decoder to use Promises.
2. Allow more things in MediaDecoderStateMachine to run on the state machine thread.
3. Removed some code no longer in use after switching to Promises.
4. A few crash fixes.
5. Create an abstract base class for a track demuxer (for MP4).
6. A few other dep patches to make all this buildable.

Reason for this: Our future ffmpeg code will depend on and/or be tied into the majority of the code added by this PR.

I've tested this with YouTube, Dailymotion, and various MP4's from around the internet, and everything appears to be working as-intended.


wolfbeast:
Damn, you've been busy, Travis ;)

Let me apply this as a working copy and build/test on Windows before I merge this in.

wolfbeast:
Working copy builds, plays both mediasource and standard containers without issues, plays h.264, webm and ogg/theora -- I'd say this is good to go!

Thanks for all the work :)

trav90:
Awesome! :tada:

PR #532, opened 2016-09-21
Provide an indication for when there are no active downloads

Lootyhoof:
This adds an additional string, `downloadsPanelEmpty.label`, when no active downloads are taking place.

This has been added due to the current ambiguity of the panel, as there is potentially an implication that active downloads might be _somewhere_, just not living in the panel at that time. With this, it is made clear that _nothing_ is currently downloading, or has downloaded in the period of the browser session.

This also keeps the styling of the footer consistent, as otherwise it would change depending whether there was a download present in the panel or not.

Some before/after images (on Windows 10, at least; footer styles will remain as they should on Vista/7/etc):

Before After


wolfbeast:
Good idea, thanks for that!

PR #534, opened 2016-09-23
Enable parallel state machine execution

trav90:
The new code introduced in PR #531 allows us to run state machines concurrently. This PR takes advantage of this and creates one unified thread pool for media code and runs the MDSM task queues on it.

This PR also fixes a couple potential crash points.


wolfbeast:
Completely deferring to you on this one ;) I know very little about the involved code.

PR #539, opened 2016-09-27
Extend non-default autocomplete shading to Win8/10

Lootyhoof:
Following up on Awesomebar changes for Win8/10, this makes the URL text be white and Switch to Tab icon use the light version, when the relevant rows are selected. This styling already existed for non-default themes (aka Classic), and is now available to Win8/10 default themes.

### DIFF
--- a/browser/themes/windows/browser.css
+++ b/browser/themes/windows/browser.css
@@ -1647,15 +1647,19 @@ richlistitem[type~="action"][actiontype="switchtab"] > .ac-url-box > .ac-action-
   padding: 0 3px;
 }
 
-@media not all and (-moz-windows-default-theme) {
-  richlistitem[type~="action"][actiontype="switchtab"][selected="true"] > .ac-url-box > .ac-action-icon {
-    -moz-image-region: rect(11px, 16px, 22px, 0);
-  }
+@media not all and (-moz-os-version: windows-vista),
+       not all and (-moz-windows-default-theme) {
+  @media not all and (-moz-os-version: windows-win7),
+         not all and (-moz-windows-default-theme) {
+    richlistitem[type~="action"][actiontype$="tab"][selected="true"] > .ac-url-box > .ac-action-icon {
+      -moz-image-region: rect(11px, 16px, 22px, 0);
+    }
 
-  .ac-comment[selected="true"],
-  .ac-url-text[selected="true"],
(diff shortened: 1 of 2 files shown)


browser/themes/windows/browser.css
```diff
@@ -1647,15 +1647,19 @@ richlistitem[type~="action"][actiontype="switchtab"] > .ac-url-box > .ac-action-
   padding: 0 3px;
 }
 
-@media not all and (-moz-windows-default-theme) {
-  richlistitem[type~="action"][actiontype="switchtab"][selected="true"] > .ac-url-box > .ac-action-icon {
-    -moz-image-region: rect(11px, 16px, 22px, 0);
-  }
+@media not all and (-moz-os-version: windows-vista),
+       not all and (-moz-windows-default-theme) {
+  @media not all and (-moz-os-version: windows-win7),
```
wolfbeast: You're nesting these -- that won't work.

Lootyhoof:
Nested queries are certainly possible, do you have a specific issue with them?

This is as simple as possible, really - we _don't_ want it on Vista/Win7, but _only_ the default theme (as Classic and user-installed system themes use this), while we _do_ want the Win8/Win10 default themes.

See below for the change this makes (apologies for not showing these sooner) - before/after:

Before After

wolfbeast:
I know you can nest queries, but the way you're nesting them seems wrong...
````
if (not vista) or (not default theme) {
  if (not 7) or (not default theme) {
    do something
  }
}
````
That doesn't seem to be what you describe you want?

Don't you want:
````
if (not default theme) {
  if (not vista) and (not 7) {
    do something
  }
}
````

... I guess writing it out I realize it does do exactly the same. Stupid inverse logic. Never mind. :smile:

However... typed text is highlighted with white text on blue, on a blue background, that should be fixed since it's not really visible? The grey on white default isn't the best but this is worse :P

Lootyhoof:
Agreed - I've been playing around with some values and come up with this. It's doing a similar job as this current commit - uses the non-default styling on Win8/10 too. Thoughts? I can place it on this PR if it suits.

Highlight

wolfbeast:
Yeah that looks fine - just add to this PR and I'll merge the total in.

Lootyhoof:
Added.

PR #540, opened 2016-09-28
A few more media updates

trav90:
Should be final followup taking advantage of code introduced in PR #531.

Makes a couple Media Decoder State Machine functions happen on the state machine thread, and dispatch AudioSink notifications async. Also adds a few more assertions.

Lightly tested here on Linux and no bustage encountered. Appears to be working as intended.


wolfbeast:
Merging this. @mattatobin can you be tryserver for these changes on Windows? I have a publish today and no time.

mattatobin:
@trav90 

Windows is green.
Tested with YouTube video

Issue #529, opened 2016-09-18
Remove D3Dcompiler_43 usage


wolfbeast:
For WebGL, we currently pull in D3DCompiler_43 (DX9) and D3DCompiler_47 (DX10/11).
The former is (I think) only needed for Windows XP support, and our sole dependency for having to install the DirectX SDK for d3dx9 headers (the latter lib is pulled from the Win 8.1 SDK).

We should be able to cut this away and have WebGL on supported operating systems through DirectX by way of the latter lib since we're no longer going to support XP.

mattatobin:
I guess.. How happy is gl and angle gonna be?

Gpower2:
I have a development setup where I haven't installed the DirectX 9 SDK, could I give this issye a try, or is someone else working on it?

wolfbeast:
Nobody else is working on this right now. If you want to give it a shot, I'll label it as assigned.

Gpower2:
Perfect! I'll get into it this weekend.

Gpower2:
Unfortunately the task is so much more than it looks. Whole WMF backend in dom/media is completely dependent on DX9. Latest mozilla-central tree contains a new implementation in DX11, however it's not VS2013 compatible, apart from the fact that it requires some new files in other dirs as well.
Also, there's a new dir structure in dom/media, where wmf, android and other backeds are moved in a new dir platforms in order for the decoders to share common code. For example, in our current code, fmp4 contains its own WMF and android implemention, while it could just use the common code from platforms dir. So I guess the current code base should be changed to that direction.

I'll keep trying to make a buildable version, though I'm not so sure I'll manage.

wolfbeast:
I didn't expect the WMF back-end to rely on DX9 -- that doesn't make much sense since the WMF side is NT6+ only anyway... And as far as I understood you can build without the DX SDK installed and the only thing it affects is whether WebGL is available or not.... Unless Mozilla has changed the way that works?

I don't think rewriting the media back-end is something we should consider viable, if that is what is required here.

Gpower2:
Well, I finally managed to get a building version, without installing DirectX SDK.
However I just deleted the d3dcompiler_43.dll from all dependencies and replaced it with d3dcompiler_47.dll.
A lot of code depends on DX9 and perhaps it would be wise to port it to DX11 at some point.
[...]

PR #542, opened 2016-09-29
Remove d3dcompiler_43 dependency (Issue #529)

Gpower2:
Remove d3dcompiler_43 dependency


wolfbeast:
Almost there. It needs a bit more work in the build system etc.

* You should check for `MOZ_FOUND_BOTH_D3D_COMPILERS` and `MOZ_REQUIRE_ALL_D3DCS`, and simplify that (since we only use 1).
* You should also check and remove configure stuff for `--enable-require-all-d3dc-versions` since that's to force the presence of both XP and Vista+ ones.
* You should check for occurrences of `MOZ_D3DCOMPILER_XP_DLL` in configure/build system as well as in other source code to see where it's used and remove/rework accordingly.
* Potential other defines/conditionals are involved? Some research may be needed.

http://xref.palemoon.org/tycho-trunk/ is going to be useful for this kind of search/cross-ref work.

Gpower2:
Sorry for the delay, I believe I removed all references from all configurations. Tested both builds and it works perfectly. :)

Gpower2:
After reviewing the changes, I realized I could also remove MOZ_DIRECTX_SDK_PATH, hence the new commit...

wolfbeast:
Looks good! Thanks.
Merging this in and I'll fire off a try build.

PR #559, opened 2016-10-05
Abolish FIREFOX_VERSION

### LINKED ISSUE #30: Remove dead code for extensions/widgetutils and localize FIREFOX_VERSION to application specific configure
In the Mozilla source tree an extension used to be built called widgetutils which provided some manor of mouse related actions. This code however has not functioned since 2010 and has not been referenced anywhere for at least that long.

This will also end the only dependence on FIREFOX_VERSION variable for the build system in global configure. All other mozilla based applications do not require FIREFOX_VERSION and it is not needed outside of building the desktop browser.

I purpose removing extensions/widgetutils from the codebase and moving the configure code for setting FIREFOX_VERSION ala THUNDERBIRD_VERSION to the per-application configure file.

http://mxr.mozilla.org/mozilla-esr24/search?string=FIREFOX_VERSION
https://bugzilla.mozilla.org/show_bug.cgi?id=948251

mattatobin:
This abolishes FIREFOX_VERSION as the browser it's self is the only consumer in the codebase in Tycho.

As a result MOZ_APP_VERSION reads browser/config/version.txt directly. I have also updated the default skin install.rdf to use MOZ_APP_VERSION and MOZ_APP_ID instead of FIREFOX_VERSION and hardcoding the GUID.

This effectively Fixes #30


wolfbeast:
LGTM!

PR #569, opened 2016-10-07
Exterminate WebAppRT

mattatobin:
Just what it says on the Dalek casing...EX-TER-MIN-ATE


mattatobin:
DO NOT MERGE YET.. There are still some dead code (but not breaking) code left over in DOM and TOOLKIT that should be gotten with this..

mattatobin:
Everything seems green..

PR #575, opened 2016-10-08
Minor fix - browser - the context menu

janekptacijarabaci:
This error happens when you do select "Save Link As..." from context menu.

Pale Moon throws an errors in the Error Console:
```
Error: TypeError: doc is undefined
Source File: chrome://browser/content/nsContextMenu.js
Line: 1108
```
See also:
https://forum.palemoon.org/viewtopic.php?f=56&t=12581#p89650
__________


wolfbeast:
That was a dumb mistake - thanks for spotting!

Issue #495, opened 2016-09-01
Places "tags" and other categories -> uncaught exception


wolfbeast:
Created as a clone of MoonchildProductions/Tycho#76

We need to hunt down where this exception is thrown and fix it.

wolfbeast:
Verified fixed.

PR #577, opened 2016-10-09
Fix: Places "tags" and other categories -> uncaught exception (and "Most visited" works)

janekptacijarabaci:
Ad #495 
Selecting "tags" in the library window throws an uncaught exception.
```
Error: uncaught exception: 2147942487
```

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=1262639
__________


wolfbeast:
Thanks for finding the error point -- it had been eluding me for some time because of the nondescript nature of it (and the BMO bug doesn't actually have a description that came up in search results when trying to find relevant data...)!

PR #578, opened 2016-10-09
Minor fix - about:cache - throws an error

janekptacijarabaci:
Minor fix - about:cache - throws an error in Error Console:
```
Error: TypeError: context is null
Source File: chrome://global/content/aboutCache.js
Line: 13
```

See also:
https://forum.palemoon.org/viewtopic.php?f=56&t=12683#p89640
__________


wolfbeast:
Hah, thanks, although I was hoping the one in the thread there would actually be nudged to supply this simple patch instead of just sitting back and complaining ;)

Issue #579, opened 2016-10-09
Forget About This Site command isn't working properly


xbeyond1:
I noticed that the Forget About This Site command found in the History menu has a few issues. 

When I want to perform that command on a site that has a large number of entries (as shown on the screenshot; I chose CNN.com for this example) in the History menu, two things happen:

1.) It only deletes one or a couple of entries instead of all of them.
2.) The Forget About This Site and Delete This Page commands become unresponsive. When I try to apply them to any site in the History menu, they simply won't work. 

![1](https://cloud.githubusercontent.com/assets/22633092/19219448/2a46fc3c-8e1c-11e6-9d97-6e43c2d730f1.png)

Also, when I execute the Forget About This Site command on a site that stores my login credentials, it sometimes clears both my username and password, and other times only my password and keeps the username. 

Pale Moon 27b1 x86

wolfbeast:
Seems to be working as expected now. Closing.

xbeyond1:
I don't think it works as it should with 27b3. I just tried it on my desktop and on a portable x86 on my laptop and the issue is still there.

PR #590, opened 2016-10-12
Forget About This Site command isn't working properly

janekptacijarabaci:
Ad #579 
\+ style clean up

This update should solve those difficulties... AFAIK.
__________
The suggestion (for example).


wolfbeast:
Oh yuck, more promises.
Thanks for the fix, though!

Issue #585, opened 2016-10-11
Add an indicator for FF/PM extensions


wolfbeast:
We should add an indicator for installed extensions in the add-on manager to indicate if extensions are Firefox extensions or if they are targeting Pale Moon natively. A simple "Firefox Orange" glyph or "Pale Moon Blue" glyph (dot) would suffice -- also to prevent unnecessary bitmaps and potential ™ issues for use of the Firefox logo.

wsdfhjxc:
Would something like this be acceptable?

![test](https://cloud.githubusercontent.com/assets/20323160/19324074/0b255618-90c0-11e6-870b-c99c349476ac.png)

wolfbeast:
@wsdfhjxc Absolutely! Although the indicators should be a tad larger.

wsdfhjxc:
Should this cover only extensions or these indicators can be also shown for themes?

wolfbeast:
It's only needed for extensions. Firefox themes are not going to be compatible, anyway.

wolfbeast:
Verified.

PR #597, opened 2016-10-13
Add indicators for (non-)native extensions in Add-ons Manager

wsdfhjxc:
This is a proposition for #585.

Blue dot means a native extension directly targeting Pale Moon.
Orange dot means a non-native extension targeting Firefox, but still allowed to run.

preview

Initially this works only with default theme because some CSS rules are needed.


wolfbeast:
Looks good! thanks for tackling this!
@Lootyhoof -- maybe this can be made theme-independent, as opposed to having to edit every complete theme?

Lootyhoof:
This may be better placed in /toolkit/mozapps/extensions/content/extensions.css - this would satisfy this theme independence.

PR #598, opened 2016-10-14
Minor fix: Places - aOldNode is undefined

janekptacijarabaci:
Steps to reproduce:
- Deleting a bookmark (= aOldNode)
- Deleting another bookmark (aOldNode does not exist)

If that node was not found, Pale Moon throws an error in Browser Console:
```
TypeError: aOldNode is undefined treeView.js:402:8
```
__________


wolfbeast:
Better to use `==` instead of `===`.

janekptacijarabaci:
OK, it's the best solution in the world :-)

wolfbeast:
:+1:

PR #600, opened 2016-10-15
Minor fix: Async (Object.create vs. __proto__)

janekptacijarabaci:
Pale Moon throws a warning in Browser Console after run:
```
mutating the [[Prototype]] of an object will cause your code to run very slowly;
instead create the object with the correct initial [[Prototype]] value using Object.create
async.js:199:4
```
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1166406
__________


wolfbeast:
Yeah, I saw that one - thanks for the quick port!

PR #601, opened 2016-10-15
Places - "gEditItemOverlay is null" when using the keyboard to cancel "Bookmark all tabs" dialog

janekptacijarabaci:
Steps to reproduce:
1. Open 2 or more tabs
2. Right click one
3. Click bookmark all tabs
4. hit [esc] on the keyboard to cancel the dialog

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1033463
__________


wolfbeast:
Yeah, unfortunately we'll have a few things to still "catch up on" that have been done in Firefox's FE. I expect more of these to pop up now and then, especially with less-common operation of the browser.

PR #602, opened 2016-10-15
Places: Menu "Views" - "Sort" - throws an errors

janekptacijarabaci:
__Places: Menu "Views" - "Sort" - throws an errors in Browser Console__

```
NS_ERROR_FAILURE: Component returned failure code: 0x80004005 (NS_ERROR_FAILURE)
[nsIStringBundle.GetStringFromName]
PlacesUIUtils.jsm:116:0
```

__My suggestion:__
"Deleting" (now they are just invisible) the following items (Menu "Views" - "Sort"):
`placesContentParentFolder` and `placesContentParentFolderPath`

\+ better information on error

__The same problem is also in Pale Moon stable (including the latest version 26).__

See also #211
__________
The suggestion (for example).

### DIFF
--- a/browser/components/places/content/places.js
+++ b/browser/components/places/content/places.js
@@ -95,7 +95,7 @@ var PlacesOrganizer = {
         aIID.equals(Components.interfaces.nsISupports))
       return this;
 
-    throw Components.results.NS_NOINTERFACE;
+    throw new Components.Exception("", Components.results.NS_NOINTERFACE);
   },
 
   handleEvent: function PO_handleEvent(aEvent) {
@@ -880,7 +880,8 @@ var PlacesSearchBox = {
         }
         break;
       default:
-        throw "Invalid filterCollection on search";
+        throw new Components.Exception("Invalid filterCollection on search",
+                                       Components.results.NS_ERROR_INVALID_ARG);
     }
 
     PlacesSearchBox.showSearchUI();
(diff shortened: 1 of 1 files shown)


wolfbeast:
Instead of skipping these items, we should simply pull the correct locale strings from the bundle for both terms. We do want to be able to sort on these columns from the menu.

wolfbeast:
A quick look shows that the locale strings are simply missing for this menu from #211 
places.properties needs something like:
````
view.sortBy.parentFolder.label=Sort by Containing Folder
view.sortBy.parentFolder.accesskey=o
view.sortBy.parentFolderPath.label=Sort by Containing Folder Path
view.sortBy.parentFolderPath.accesskey=P
````

wolfbeast:
I'm merging this in as a workaround for now, we can see about enhancing this later on because it needs some more work to actually being able to sort on those columns.

janekptacijarabaci:
Yes, only this one doesn't work properly.

In this case, I leave it to you :-)

See also:
https://github.com/MoonchildProductions/Pale-Moon/blob/master/browser/components/places/content/treeView.js
https://github.com/MoonchildProductions/Pale-Moon/blob/master/toolkit/components/places/nsINavHistoryService.idl
https://github.com/MoonchildProductions/Pale-Moon/blob/master/toolkit/components/places/nsNavHistory.cpp
https://github.com/MoonchildProductions/Pale-Moon/blob/master/toolkit/components/places/nsNavHistoryQuery.cpp
https://github.com/MoonchildProductions/Pale-Moon/blob/master/toolkit/components/places/nsNavHistoryResult.cpp

PR #603, opened 2016-10-15
Restore OTS_API visibility

trav90:
This was accidentally removed when OTS was updated to 5.0.1.


wolfbeast:
Sorry if I wasn't clear -- I'm working on updating OTS to 5.1.0 and will patch it up in that commit :)

trav90:
Ah, ok. I blame lack of coffee on my part! :)

Thanks for pointing it out to me. I'll keep it in mind for the future.

PR #609, opened 2016-10-17
Clean up redundant sidebar component

mattatobin:
By order of the Department of Redundancy Department this redundant component has been deemed redundant.


wolfbeast:
What was this previously used for and why has it become redundant?

mattatobin:
browser/components/sidebar is redundant since we use the version in the platform. This is what setback the port a week looking for the the reason xpcom fell over.

In the past every application had to provide their own copy but they were all the same so it was moved to platform.

Until it was confirmed to be fine I simply made the DIRS entry contingent to a non-existent confvar and neglected to clean it up afterword until now.

PR #611, opened 2016-10-17
Move FE DevTools to toolkit

mattatobin:
This pull request moves everything from
/browser/devtools
/browser/locales/en-US/chrome/browser/devtools
/browser/themes/shared/devtools
/browser/themes/[os]/devtools

to their toolkit equivalent directories. The configure argument still applies but was moved to topsrcdir configure.

The hope is that, without compromising browser's ability to build with them, that other applications and consumers of the codebase (if there ever will be any) can pull in Mozilla DevTools perhaps to create a XUL/Web Developer IDE.

Since Tycho will not be consuming them by default there isn't any additional impact to extensions (I haven't found any but Firebug but that is a non-issue) since those extensions would be impacted without FE components anyway and most likely those targeting our front end would expect the older tools anyway.. 

However, there is an impact to themes ( @Lootyhoof forgive me) which **_I_** will resolve over time (since they are not being built anyway by default it isn't an emergency).

I have build tested on Windows with no ill effects and I don't expect any on Mac either. Linux, however, is a little funny with toolkit themes as it does jar.mn subset file replace after everything from Windows is copied to FINAL_TARGET. So that will need to be eval'd by my self or @trav90. Though, I don't anticipate any issues but won't know until a Linux build is performed with --enable-devtools.

In any event, I believe this is the best place for the code to live going forward.


wolfbeast:
You'll have to excuse me for not merging this in until I get at least a green on Linux builds.

trav90:
All appears to be green here.

mattatobin:
I can confirm!

Issue #522, opened 2016-09-14
Provide hints for parallelization


wolfbeast:
Although the auto-parallelizer in MSVC will analyze and decide automatically, it's very conservative in where it applies multi-threading for multiple cores.

Providing hints with `#pragma loop(hint_parallel(n))` and where needed re-ordering surrounding code or making it more parallel-friendly (e.g. duplicating scalar variables to exclusives) will help the compiler to make the right decision and parallelize where possible.

wolfbeast:
Vectorization is already used and in many cases this is mutually exclusive. Parallel code often means SIMD instructions can't be used; this is an important performance balance to keep in mind.

It may be a good idea to run with the vectorization reporter enabled to see what loops are already vectorized by the compiler and target loops for parallelization that aren't, first.

wolfbeast:
Apparently, Qpar `#pragmas` require the inclusion of `vcomp{nnn}.dll`, likely for the CPU core detection routine and dispatcher code. this needs to be included in the build system (manually added for now before packaging)

mattatobin:
Tycho's build system was updated to handle this in 993a7bc93c1855b1d9dda3768cc32cf91a4f15ec

Branch v26_Dev is WONTFIX unless I get bored
Branch v25_Dev is N/A

wolfbeast:
> Branch v26_Dev is WONTFIX unless I get bored

Branch v26_Dev is a WONTFIX. I doubt I'll build much more on that and even if so, a manual copy or not using `-Qpar` is easy enough.

wolfbeast:
Considering #915 we should remove the MSVC pragmas and replace them with standardized OpenMP pragmas.

PR #614, opened 2016-10-17
Fix vcomp110.dll packaging

rhinoduck:
The packaging step cannot find vcomp110.dll, this pull request fixes that. See the commit message for more information.

### DIFF
--- a/browser/installer/package-manifest.in
+++ b/browser/installer/package-manifest.in
@@ -99,7 +99,7 @@
 #elif MOZ_MSVC_REDIST == 1700
 @BINPATH@/msvcp110.dll
 @BINPATH@/msvcr110.dll
-; @BINPATH@/vcomp110.dll
+@BINPATH@/vcomp110.dll
 #endif
 #endif
 #endif
--- a/browser/installer/removed-files.in
+++ b/browser/installer/removed-files.in
@@ -699,6 +699,7 @@ xpicleanup@BIN_SUFFIX@
   #if MOZ_MSVC_REDIST != 1700
     msvcp110.dll
     msvcr110.dll
+    vcomp110.dll
   #endif
   plugins/npnul32.dll
 #endif
--- a/build/win32/Makefile.in
+++ b/build/win32/Makefile.in
@@ -54,6 +54,9 @@ REDIST_FILES = \
   msvcp110.dll \
   msvcr110.dll \
   $(NULL)
+REDIST_OPENMP_FILES = \
+  vcomp110.dll \
+  $(NULL)
 endif
 
 ifdef REDIST_FILES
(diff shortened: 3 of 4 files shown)


mattatobin:
See: https://github.com/MoonchildProductions/Pale-Moon/issues/522#issuecomment-252867943

wolfbeast:
If committing to v26, please make a PR against v26_Dev, not the RelBranches.

rhinoduck:
First, let me apologise for the short description, I was in a hurry to leave. I guess I should have delayed making the pull request.

To clarify, I've used Pale Moon for some time now and I wanted to look at the innards out of curiosity. I'm not familiar with the code base nor with the Mozilla build system. I ran into the issue, and as I seemed to have been able to fix it for myself, I thought I'd share.

I didn't know about the changes in Tycho that you linked, thus my changes look different. I now see there is a MSVC_OPENMP_DLL define in there for 27, but in 26, I did not immediately see a way or a reason to make the OpenMP dll/parallelisation optional. My starting point were the recommended settings to put into mozconfig which can be found on the wiki/in the forums post. Being new to all the options available, I may have made a false assumption that vcomp110.dll should be included in all builds where the CRT libs are included.

That brings me to "how it is used in mozconfig". It is based on the assumption mentioned above; all you need to do, is define WIN32_REDIST_DIR exactly as you did until now (which really should be called WIN32_REDIST_CRT_DIR, but I didn't want to make more changes than necessary), and the following line in configure.in will then take care of defining WIN32_REDIST_OPENMP_DIR for you.

```WIN32_REDIST_OPENMP_DIR=`dirname "$WIN32_REDIST_DIR"`/Microsoft.VC110.OPENMP```

If you need to get vcomp110.dll from a non-standard location, you can still explicitly define WIN32_REDIST_OPENMP_DIR in mozconfig yourself and that value will then take precedence.

I ran all of configure, build, package, and installer mach commands on the clobbered tree successfully, and checked that vcomp110.dll was present in the dist zip archive. I limited myself to VS2012 as this seems to be the preferred toolchain to build this version of PM on Windows. I'll give it another go tonight to confirm.

If you have the time to spare, I'll be happy to learn about what I overlooked. If not, feel free to close the request. Though EOL is nearing for v26, I guess, I just thought it would be nice for what is currently still recommended as the entry point to PM development to work out of the box completely when following official instructions as, apart from this little issue, things went really smoothly.

I changed the base branch, though note that the dev branch is currently missing 06c96a2893ba107cd4e5e62f37da6b9aab7a9dda which the further fix depends on... ah, I see you probably already added that to the PR. Again, not a very familiar territory for me yet.

wolfbeast:
Thanks for explaining and sharing your fix.
@mattatobin the fix does look complete, although the `dirname` magic to get the parent level is a little fragile.

There just wasn't going to be much pressing need for this because of the nearing EoL for v26, as you said, and a manual copy of the DLL would certainly be simple enough.

PR #619, opened 2016-10-21
Abstract MP4 code away from libstagefright

trav90:
This PR abstracts the majority of our MP4 code away from libstagefright (except for metadata), and replaces it with our own code (with more to come soon, including the glue and a demuxer for FFmpeg to work). Also wrapped up in this PR is some code cleanup of now unused stagefright code, several crash and regression fixes, and a few code styling fixes.

I've tested playing MP4, OGG, and mediasource files, as well as YouTube (both MP4 & WebM). All appears to be working as intended.

PR #622, opened 2016-10-22
Add Interval and IntervalSet objects

trav90:
This adds a couple objects that are used by the MediaDataDemuxer added in PR #619 (and will be used by additional code in the future).

Tested on Linux and everything appears to work as intended.


wolfbeast:
Wow, this is some of the cleanest code I've seen come out of Mozilla Labs for a long time.

PR #624, opened 2016-10-22
OS X theme improvements

Lootyhoof:
This improves the OS X default theme in the following ways:

1. Uses the spacing and overall styling of the bookmarks toolbar from the Firefox 24 OS X default theme to keep the spacing between icons consistent
2. Replaces the majority of icons in places with their OS X counterparts (including adding HiDPI versions), making them feel more OS-native
3. Uses the tab styling as-is from the Windows theme (it was doing this already, but using our PM25-era styling)
4. Fixes the icon mapping of the download status icon

Some before and afters:

Before After


wolfbeast:
Looks like a good improvement.

PR #650, opened 2016-11-02
Add variadic NS_NewRunnableMethodWithArgs

trav90:
My next media-related PR will be dependent on the patches here. I've built successfully, and both in-tree tests that are modified in these patches pass OK. Also been using the browser for a while, and nothing appears broken.

For reference:
[1131445](https://bugzilla.mozilla.org/show_bug.cgi?id=1131445)
[1137583](https://bugzilla.mozilla.org/show_bug.cgi?id=1137583)
[1153295](https://bugzilla.mozilla.org/show_bug.cgi?id=1153295)

### DIFF
--- a/mfbt/TypeTraits.h
+++ b/mfbt/TypeTraits.h
@@ -162,21 +162,34 @@ template<typename T>
 struct IsArray : detail::IsArrayHelper<typename RemoveCV<T>::Type>
 {};
 
+namespace detail {
+
+template<typename T>
+struct IsPointerHelper : FalseType {};
+
+template<typename T>
+struct IsPointerHelper<T*> : TrueType {};
+
+} // namespace detail
+
 /**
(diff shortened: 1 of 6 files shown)


wolfbeast:
Dangerous waters to be in. I'll make a Windows build and give it a thorough thrashing.

wolfbeast:
Looks all good with a Win x86 build running through some intensive benchmarks and general heavy websites. :+1:

trav90:
Thanks! Sorry about being a bit vague in the description, but to be honest I didn't fully understand what all this code was doing, so had to defer to BMO on this one.

wolfbeast:
I don't think anyone fully understands all aspects of how XPCOM works ;)

This is dangerous stuff because it changes the way refcounted pointers work (used everywhere in the code base), so it's a bit like touching every line of code.

trav90:
Ah ok. I see. Thanks!

PR #654, opened 2016-11-04
Allow readers to use asyncronous ReadMetadata

trav90:
Our soon-to-be new MediaFormatReader (which I'm working on and should be the final component needed before we can start using ffmpeg) will be a mostly asynchronous component, and is dependent on being able to read metadata asynchronously. This PR makes that happen.

Tested and appears to be working as intended.

### DIFF
--- a/dom/media/MediaDecoderReader.cpp
+++ b/dom/media/MediaDecoderReader.cpp
@@ -190,13 +190,13 @@ MediaDecoderReader::ComputeStartTime(const VideoData* aVideo, const AudioData* a
 }
 
 nsRefPtr<MediaDecoderReader::MetadataPromise>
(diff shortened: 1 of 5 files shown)


wolfbeast:
I guess we're forced to use all this async nonsense, eh? Thankfully for media handling it kind of makes sense to do so.

wolfbeast:
Tested, can't find anything amiss with MP4 parsing on several video sites.

trav90:
Yeah, pretty much. Mozilla was big on making much of the media backend async and "promise-y" right after ESR38. And so obviously everything we want to port is all tied into those changes.

trav90:
Great!

PR #657, opened 2016-11-05
Minor fix: Options - Applications - throws a warning

janekptacijarabaci:
Pale Moon throws a warning in the Browser Console:
```
JavaScript 1.7's let blocks are deprecated applications.js:1771:70
```

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1108311

---


wolfbeast:
Thanks -- there will probably be more deprecated let() blocks scattered around the code; No need to find the related BMO bug for those, the conversion to ES5+ let syntax is straightforward enough.

janekptacijarabaci:
> No need to find the related BMO bug for those, the conversion to ES5+ let syntax is straightforward enough.

OK, thank you.

PR #659, opened 2016-11-07
Media decoder fixes

trav90:
This PR improves our MSE support by preventing MSE from notifying that a resource is ready even if it is not. This PR lets the decoders manage their own readiness. These fixes will also prevent double-work later and make it easier for me to get the MediaFormatReader ported over. Other changes:

Cleans up some code.
Enables media.decoder.heuristic.dormant.enabled across all platforms
Added some more async code to OMX MediaCodecProxy
Various crash fixes in OMX and MP4Reader

Tested Youtube and a couple other test sites and all appears to be working as intended.

### DIFF
--- a/dom/media/fmp4/MP4Reader.cpp
+++ b/dom/media/fmp4/MP4Reader.cpp
@@ -155,7 +155,6 @@ MP4Reader::MP4Reader(AbstractMediaDecoder* aDecoder)
   , mDemuxerInitialized(false)
   , mFoundSPSForTelemetry(false)
   , mIsEncrypted(false)
-  , mAreDecodersSetup(false)
   , mIndexReady(false)
   , mLastSeenEnd(-1)
   , mDemuxerMonitor("MP4 Demuxer")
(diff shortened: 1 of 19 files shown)


mattatobin:
Does this help the MSE/VP9 seek and replay issues?

wolfbeast:
@mattatobin probably not. Can't hurt to try; you never know, considering YouTube.

wolfbeast:
Doesn't fix the looping -> no sound issue at least.

wolfbeast:
I've just created an intended release build and it seems something in this PR prevents the browser from properly shutting down after visiting YouTube. I've also had a hard hang with it on YouTube when switching resolutions.
Any idea what might be the problem? Perhaps b3e02f4 ?

wolfbeast:
I backed out b3e02f4 and it seems to have fixed these issues...

trav90:
I'll look into it.

trav90:
That patch was created at Moz to fix a crash caused by the first commit in the PR. Obviously, a hard hang on Windows is not the intended result. Looking back over the patch and BMO bug, I think it should be safe to back out that commit.

trav90:
This should also be fixed in the new MediaFormatReader (PR #660), as the sections of code affected by b3e02f4 have been rewritten.

trav90:
Background is that at this point Moz was making preparations to move from using the MP4Reader to MediaFormatReader (as we are), so that's what the end goals were. This particular issue was to improve YouTube/MSE/dormant compatibility, and enable mDormantEnabled across all platforms. While they were at it, they decided they needed to remove a bunch of other code (of course), which resulted in a regression for them that was fixed in the problem commit.

Since I did not remove/make every single change Moz has (especially in WIndows-land), I think the additional check that this commit made doesn't play nice with some of our existing code.

PR #661, opened 2016-11-10
Revert "Ensure decoders are properly shutdown"

trav90:
Should fix Windows hangs caused by PR #659.

Currently building on Linux to make sure no issues are seen (although I don't foresee any).

### DIFF
--- a/dom/media/fmp4/MP4Reader.cpp
+++ b/dom/media/fmp4/MP4Reader.cpp
@@ -342,8 +342,6 @@ MP4Reader::ReadMetadata(MediaInfo* aInfo,
   } else if (mPlatform && !IsWaitingMediaResources()) {
     *aInfo = mInfo;
     *aTags = nullptr;
-    NS_ENSURE_TRUE(EnsureDecodersSetup(), NS_ERROR_FAILURE);
-    return NS_OK;
   }
 
   if (HasAudio()) {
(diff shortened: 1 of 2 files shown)


trav90:
Hmm, seems to cause problems with seeking and switching resolutions on YT on Linux. Investigating.

trav90:
Sigh, looks like GStreamer versions above 1.8 break YouTube. I downgraded my distro back to gst1.8 and everything looks good with the reverted out.

wolfbeast:
Was there any more information about why these changes were made in bugzilla? any specifics there are about why this was changed to begin with?

trav90:
Background is that at this point Moz was making preparations to move from using the MP4Reader to MediaFormatReader (as we are), so that's what the end goals were. This particular issue was to improve YouTube/MSE/dormant compatibility, and enable mDormantEnabled across all platforms. While they were at it, they decided they needed to remove a bunch of other code (of course), which resulted in a regression for them that was fixed in the problem commit.

Since I did not remove/make every single change Moz has (especially in Windows-land), I think the additional check that this commit made doesn't play nice with some of our existing code.

wolfbeast:
Thanks for digging into it a bit deeper! Makes perfect sense.

trav90:
No problem!

PR #664, opened 2016-11-11
fix missed deniedPortAccess.longDesc

JustOff:
restored from FF49 en


wolfbeast:
Note that you were using mozilla.org as an example there by verbatim copy from Firefox. I've updated that before merge, as well as using a more apt port number for this issue.

JustOff:
ok, now I got it

PR #670, opened 2016-11-13
About Dialog - missing translated links

janekptacijarabaci:
If it is not intentionally...

@JustOff 
Also FYI

---


wolfbeast:
It is intentional.

Issue #642, opened 2016-11-01
Add desktop notifications permissions to Permissions Manager


wolfbeast:
Displaying of desktop notifications by the browser should be a per-domain entry in about:permissions

Blocks #638

Issue #641, opened 2016-11-01
Add installing of add-ons permissions to Permissions Manager


wolfbeast:
Installing of add-ons should be a per-domain entry in about:permissions

Blocks #638

Issue #640, opened 2016-11-01
Add hiding of mouse pointer permissions to Permissions Manager


wolfbeast:
Hiding of the mouse pointer should be a per-domain entry in about:permissions

Blocks #638

Issue #639, opened 2016-11-01
Add loading of images permissions to Permissions Manager


wolfbeast:
Loading of images should be a per-domain entry in about:permissions

Blocks #638

PR #675, opened 2016-11-14
Permissions - more values: Load Images, Install Addons, Hide the Mouse Pointer

janekptacijarabaci:
## Permissions Manager

Ad:
#639 (Add loading of images permissions)
#641 (Add installing of add-ons permissions)
#640 (Add hiding of mouse pointer permissions)

\+ style clean up - sort items (`Permissions Manager` vs. `Page Info`)

Ad (not created):
#642 (Add desktop notifications permissions)
I don't know which key you want to use... `dom.push.enabled`?
In addition, it is necessary to change more...

---

The suggestion (for example).

__It is necessary to test all options!__

---

### DIFF
--- a/browser/components/preferences/aboutPermissions.js
+++ b/browser/components/preferences/aboutPermissions.js
@@ -178,6 +178,28 @@ Site.prototype = {
   },
 
   /**
+   * Gets logins stored for the site.
+   *
+   * @return An array of the logins stored for the site.
+   */
+  get logins() {
+    let httpLogins = Services.logins.findLogins({}, this.httpURI.prePath, "", "");
+    let httpsLogins = Services.logins.findLogins({}, this.httpsURI.prePath, "", "");
+    return httpLogins.concat(httpsLogins);
+  },
+
+  get loginSavingEnabled() {
+    // Only say that login saving is blocked if it is blocked for both http and https.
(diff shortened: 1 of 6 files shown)


wolfbeast:
I think you misunderstand what this is supposed to do.
You are dealing with site-specific preferences here, so you don't want to read or set the global preferences in the browser, but rather deal with the site-specific preferences database.

janekptacijarabaci:
> You are dealing with site-specific preferences here, so you don't want to read or set the global preferences in the browser, but rather deal with the site-specific preferences database.

I think I'm doing... But if I have misunderstood something - you take it as an unfinished example :-)

wolfbeast:
I'll have to check in detail when my head is on right. sorry if it'll take a bit; I'm not familiar with this specific code yet.

wolfbeast:
After checking the overarching logic involved, this looks good. I'll merge it in and give it a test.

janekptacijarabaci:
@JustOff 
FYI, new strings to translate (potentially):
https://github.com/MoonchildProductions/Pale-Moon/commit/629aadc1a01471d9668f926486c60159a9137f5e#diff-58ebd4ce0478b88cc49e6e016fa17b8d

PR #676, opened 2016-11-14
Permissions vs. Page Info - cookies

janekptacijarabaci:
__Steps to reproduce (e.g.):__

1) Go to http://www.palemoon.org/
2) Main menu: `Tools - Page Info - Permissions` - uncheck `Set Cookies` - `Use Default`
3) Go to `about:permissions`
4) Select the appropriate domain
5) Select `Set Cookies` - `Allow First Party Only`
6) Go back to `Tools - Page Info - Permissions`

Throws an error in Error Console:
```
Error: TypeError: radio is null
Source File: chrome://browser/content/pageinfo/permissions.js
Line: 205
```

But my patch is just a suggestion...

---


janekptacijarabaci:
@JustOff 
FYI, a new string for translation: `permAllowFirstPartyOnly`

JustOff:
We can just copy from [this one](https://github.com/MoonchildProductions/Pale-Moon/blob/master/browser/locales/en-US/chrome/browser/preferences/aboutPermissions.dtd#L23), right?

PR #677, opened 2016-11-14
OS X: Use static PNGs for checkmarks

Lootyhoof:
This re-introduces static images for checkmark items on context menus on OS X. As of Tycho these were using `-moz-appearance: checkmenuitem`, though oddly no checkmarks are rendered at all with this method. While a stopgap (ideally the `-moz-appearance` value should be fixed) this fix will at least correctly render, using the same method as was used in pre-Australis Firefox (and Pale Moon <=26).


JustOff:
@Lootyhoof, and what about radio checkmarks?
They have exactly the same problem.

PR #678, opened 2016-11-14
MediaFormatReader fixes

trav90:
This PR is a "roll-up" patch containing many bug/regression fixes, performance improvements, and potential crash point fixes for our MediaFormatReader. This PR should make MediaFormatReader's code quality at least on par with our current MP4Reader, if not above it. Also makes some improvements to our MSE support. Highlights:

- Add a MediaDecoderReader::NotifyDataRemoved() method -- This allows for the MSE TrackBuffer to notify the sub reader that it's just finished reading the data. It wasn't required with MP4Reader because calling MP4Reader::GetBuffered() always forced to recalculate the index.

- We now *never* do blocking reads with MSE. Occasionally there was a rounding issue (usually by one microsecond) between the time of the last sample and the moof's end time. This caused the MediaSourceReader attempt to decode one more sample that we don't have, entering a blocking read. We ensure that the sample is within the resource's content, or report EOS otherwise.

- A few fixes to our MP4Demuxer.

- Fix some crashes and regressions when reading MP4 metadata.

- Cache MediaFormatReader's buffered time ranges, and improve performance of GetBuffered() -- (lowers the CPU usage when playing videos).

- Do not allocate memory spanning across MP4 ftyp and moov atom -- currently, we typically allocate an amount of memory equivalent to the size of the MP4 file. Obviously for very large files, this causes problems. with this patch, we now reconstruct the metadata to only have the ftyp and moov atoms contiguously.

- Crash fixes in MP4TrackDemuxer.

- Improve buffering time ranges with MediaFormatReader.

- Make web audio read metadata async, and hook it up to MFR.

- Fix a few crash/hang points in MFR.

Been testing on YouTube (and elsewhere) for the past few hours, and everything is working as intended here on Linux.


trav90:
If everything in the PR works good on Windows, I believe we can enable MediaFormatReader by default (I'll submit a PR later).

wolfbeast:
I'll make another trunk build, flip the prefs, and have a check.

PR #690, opened 2016-11-18
Add bits-per-second interface to MediaRecorder

trav90:
Listed on BMO as a blocker (of a blocker) for enabling FFmpeg. I can confirm that it compiles with no problems, but unfortunately I do not have a mic or camera to be able to test. Sorry :(

For reference:
BMO: [1161276](https://bugzilla.mozilla.org/show_bug.cgi?id=1161276)
MDN: [MediaRecorder Web API](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder)
MDN: [Using the MediaStream Recording API](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream_Recording_API/Using_the_MediaStream_Recording_API)

### DIFF
--- a/dom/media/MediaRecorder.cpp
+++ b/dom/media/MediaRecorder.cpp
@@ -581,9 +581,17 @@ private:
 
     // Make sure the application has permission to assign AUDIO_3GPP
     if (mRecorder->mMimeType.EqualsLiteral(AUDIO_3GPP) && Check3gppPermission()) {
-      mEncoder = MediaEncoder::CreateEncoder(NS_LITERAL_STRING(AUDIO_3GPP), aTrackTypes);
+      mEncoder = MediaEncoder::CreateEncoder(NS_LITERAL_STRING(AUDIO_3GPP),
+                                             mRecorder->GetAudioBitrate(),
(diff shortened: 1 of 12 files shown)


wolfbeast:
I have to wonder why recording/encoding stuff is a blocker?...  Why do we need a recorder **web api** anyway, and what does it have to do with ffmpeg used as a playback platform?
Do we even want this web api? Are there any security considerations to take into account for camera/microphone use? can this api be disabled easily?

trav90:
I didn't add the API. It has been present in Mozilla tree since I think the FF25 era. This change only allows for adjusting bitrate values when capturing HD video (which ffmpeg will be decoding).

It doesn't look too invasive if we want to rip it out at some point. I do not know if it can be easily disabled.

wolfbeast:
Well if the API itself isn't there, then it's less of an issue, of course. Just some concerns :)

trav90:
I didn't add the API, but it has been present in Moz tree since post esr-24 (and as such is in Tycho).

wolfbeast:
Right, right, sorry. Not paying attention.

PR #692, opened 2016-11-19
Update jemalloc to 4.3.1

Gpower2:
Update jemalloc to 4.3.1
Built and checked extensively.


wolfbeast:
How did you build/test this? if you just built with "enable-jemalloc" then it won't actually use this lib.

Gpower2:
Hmm, I thought that jemalloc is the default memory lib used by PaleMoon, so I just built it with the usual parameters and as for testing, I opened a lot of memory hungry tabs and used it extensively for more than an hour monitoring its performance.

The options I used for build are:

ac_add_options --enable-application=browser
ac_add_options --enable-optimize="-O2"
ac_add_options --enable-jemalloc
ac_add_options --disable-shared-js
ac_add_options --disable-sandbox
ac_add_options --disable-eme
ac_add_options --disable-tests

wolfbeast:
If you use `--enable-jemalloc` then it will use our in-tree (non-lib, mozilla-derivative) implementation.
to use the full lib you have to also specify `--enable-jemalloc-lib` in addition to `--enable-jemalloc` which will set the appropriate build flag to switch to the full lib. Otherwise you're not testing anything you changed ;)

Gpower2:
Thanks! Building as we speak...
However I have to ask, do Palemoon official releases use jemalloc-lib or not?

wolfbeast:
Official releases don't use the lib. It causes some issues (we've tried using it for a few versions but had to return to the custom code because it would cause memory inflation or similar problems). If the lib update improves things though, then we can have another look.

wolfbeast:
Did you get around to giving this a good tryout?

Gpower2:
Unfortunately I never managed to get a working build, and I haven't found the time to check it further.
I will look into it later this weekend.

wolfbeast:
All right! I'll keep this PR open as-is then

Gpower2:
Unfortunately I never managed to build PaleMoon with jemalloc 4.3.1
Upstream Mozilla code has changed the build system, and I didn't manage to port the necessary changes to PaleMoon's current build system to run autoconf for jemalloc...
So I will revert all commits and revoke this PR.

PR #693, opened 2016-11-20
Minor fix: DevTools - Network

janekptacijarabaci:
__Steps to reproduce (e.g.):__

1) Open `DevTools` - `Network`
2) Go to https://www.google.com/maps/@37.3565066,-122.0322134,13z

Throws an errors in Browser Console:
```
TypeError: Cannot convert string to ByteString because
the character at index 0 has value 63377 which is greater than 255.
network-monitor.js:347:22
```

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1240959

---


wolfbeast:
Thanks!

PR #703, opened 2016-11-22
Update mozconfig per Tobin, switch to release branding for 27.0.0 release!

stevenschlansker:


build/macosx/mozconfig.tycho
```diff
@@ -14,18 +15,10 @@ ac_add_options --enable-application=browser
...
+#ac_add_options --enable-debug
 #
-ac_add_options --disable-crashreporter
-ac_add_options --disable-accessibility
-ac_add_options --disable-parental-controls
-ac_add_options --disable-webrtc
-ac_add_options --disable-safe-browsing
-ac_add_options --disable-url-classifier
-ac_add_options --disable-sandbox
+ac_add_options --disable-url-classifier # XXX: what does this do, if anything?
```
wolfbeast: This shouldn't do anything anymore, I think. Pretty sure this was pulled out.
URL-classifier is a database of known-bad sites, part of Google safebrowsing.

build/macosx/mozconfig.tycho
```diff
@@ -14,18 +15,10 @@ ac_add_options --enable-application=browser
...
 #
-#ac_add_options --enable-strip
+# ac_add_options --enable-strip
```
wolfbeast: You'll probably want `--enable-strip` for release builds to strip superfluous symbols.

stevenschlansker: I left it in for now so that people can report crashes with stack traces.  I'll do a second run and see how much the executable shrinks, see if the savings justify the loss of debug info.

stevenschlansker:
So, the built sizes with and without `--enable-strip` are nearly identical.  Maybe a 2MB savings or so.

Do you think that reduction is worth the loss of debugging?

stevenschlansker:
Hm, after a closer inspection of the build log, it seems that the `--enable-strip` did not actually cause any invocation of the `strip` utility.  Or if it did, the Makefile did not emit a message for it.

trav90:
I know that `strip` works here on Linux - otherwise you end up with an absolutely HUGE tarball.

Issue #700, opened 2016-11-22
'Download status button action' selections doesn't work (PM27)


THEtomaso:
The 'Download status button action' selections doesn't do anyting (Tools > Status bar options > Download).
Clicking the button will show the new default download window (which I don't like at all) no matter what!

(Pale Moon v27.0.0,x64)

janekptacijarabaci:
There are two different "downloads buttons" - do you use the "right" button?

But I have another problem:
![1](https://cloud.githubusercontent.com/assets/2373486/20560904/f4130d8a-b17b-11e6-8819-821e7bf0a542.png)

1. Status text is initially `null`.
2. After clicking to "Downloads" button (the original Downloads button) - the popup window is elsewhere.

wolfbeast:
I tested this -- using the "download status" button, which is the status bar downloads indicator equivalent of the toolbar download button, making the selections in the status bar options window work as they should. I tested library and tab functions and they open the library and downloads tab, respectively.

If you have both indicators in your UI, the status bar one will be the priority anchor (by design); you should remove the one you don't want to use to have it be consistent.

In the customize window, the two entries are:
- `Downloads` - the default (panel) indicator as-used in Firefox.
- `Download status` - the status bar indicator which responds to the selections made in its options.

wolfbeast:
Unless there's something specific that doesn't work as I said, this issue should be closed.

THEtomaso:
Yeah, I figured this out a while ago, but forgot to post it here.
The settings are a bit confusing, that's all.
I posted some info here, in case anyone else wants it to work like I do:
https://forum.palemoon.org/viewtopic.php?f=3&t=14560/

PR #709, opened 2016-11-24
[Suggestion] Minor fix: Status Bar - Status Text - null

janekptacijarabaci:
Ad https://github.com/MoonchildProductions/Pale-Moon/issues/700#issuecomment-262497785:

Status text is initially `null`.
___

But I'm not sure if you want it that way...

---

### DIFF
--- a/browser/components/statusbar/content/urlbarBindings.xml
+++ b/browser/components/statusbar/content/urlbarBindings.xml
@@ -178,8 +178,8 @@
 					this._status = aURL;
 
 					// Hide the over-link immediately if necessary.
-					if((!aURL && (XULBrowserWindow.hideOverLinkImmediately
-					   || this._hideOverLinkImmediately)) || this.focused)
+					if(!aURL && (XULBrowserWindow.hideOverLinkImmediately
+					   || this._hideOverLinkImmediately))
 					{
 						this._setOverLinkState(null);
 						return;


wolfbeast:
The status should be null when focusing the address bar; i don't think this change does what it is supposed to achieve there.

Ultimately, I want to get rid of this whole status-in-the-address-bar nonsense that is only useful if you don't have an actual status bar.

janekptacijarabaci:
> The status should be null when focusing the address bar

IMHO: After this change - the status should be `nothing` (the text is not visible) when focusing the address bar. It still holds true.

But...

> Ultimately, I want to get rid of this whole status-in-the-address-bar nonsense that is only useful if you don't have an actual status bar.

If you want to close it, feel free to do it :-)

wolfbeast:
No idea when anyone gets around to removing this, so in the meanwhile taking this fix ;)

janekptacijarabaci:
After you try it yourself (maybe I missed something) :-)

wsdfhjxc:
> Ultimately, I want to get rid of this whole status-in-the-address-bar nonsense that is only useful if you don't have an actual status bar.

Does it mean ditching all address bar related features from S4E, including progress meter?

wolfbeast:
> Does it mean ditching all address bar related features from S4E, including progress meter?

Yes, because it all uses a binding to the address bar which is the cause of these kinds of issues.

Issue #708, opened 2016-11-23
about-dialogue: incorrect architecture info


MingcongBai:
When running Pale Moon 27.0.0 on PowerPC64, the about dialogue says that the browser was built for x64, which does not make sense as x64 really refers to x86_64 or x86 with 64-bit extension.

![snapshot7](https://cloud.githubusercontent.com/assets/5006263/20573245/8b4b1e5c-b174-11e6-8268-5dbce85d85f1.png)

wolfbeast:
It will require a simple patch to adjust the aboutdialog XUL based on built architecture to correct this.
Should be self-evident :) patches welcome!

MingcongBai:
https://github.com/MoonchildProductions/Pale-Moon/blob/5214789d035f4e4c08899a6e281116520f605ec6/browser/base/content/aboutDialog.xul#L45

But it is a simple 64-or-32 condition branch here... How could this possibly be simple?

wolfbeast:
You can add more conditions to it! if x86/x64 only applies to certain architectures, then add a condition to use those terms for the applicable architectures, and use a "generic" term otherwise (32-bit/64-bit etc.)
The only "not simple" part of this is taking the time to research what conditions you need.

MingcongBai:
Experimental PR opened here, https://github.com/MoonchildProductions/Pale-Moon/pull/710.

PR #710, opened 2016-11-24
aboutDialog: more architectural detection

MingcongBai:


wolfbeast:
I've thought about this some more and I don't think it'd be a good place to do this kind of geek-level detection.
It's probably just better to generalize it to architecture-independent bit-ness statements instead of trying to bake in all theoretically possible architectures for our browser about box.
This kind of statement is probably better placed in about:support or about:buildconfig if we really want it.

MingcongBai:
> It's probably just better to generalize it to architecture-independent bit-ness statements instead of trying to bake in all theoretically possible architectures for our browser about box.

I do agree. I will change the PR now.

MingcongBai:
Done.

wolfbeast:
Nit: "bit" shouldn't be capitalized.

MingcongBai:
Good?

wolfbeast:
Yes! Thanks :)

wolfbeast:
This resolves #708.

PR #720, opened 2016-11-27
Don't use a FallibleTArray as backend for MediaRawData

trav90:
This PR heads off some potential memory inflation in MediaRawData. Also fixes build bustage for some of my local ffmpeg patches that I will be testing and submitting soon.

These patches tested and appear to work as intended on Linux.


wolfbeast:
I've never been happy about Mozilla's use of fallible arrays. Happy to see more of that gone.

PR #743, opened 2016-12-04
Revamp extension target indicators for tooltip support (alternative)

JustOff:
Alternative to #741:

- Different tooltips for native and compatible addons
- Indicator itself isn't supposed to be locale dependent
- Added transparency to indicator for disabled addons
~~- Size of indicator reduced from 150% to 120%~~

### DIFF
--- a/toolkit/locales/en-US/chrome/mozapps/extensions/extensions.dtd
+++ b/toolkit/locales/en-US/chrome/mozapps/extensions/extensions.dtd
@@ -210,6 +210,9 @@
 <!ENTITY addon.loadingReleaseNotes.label      "Loading…">
 <!ENTITY addon.errorLoadingReleaseNotes.label "Sorry, but there was an error loading the release notes.">
 
+<!ENTITY addon.nativeAddon                    "This add-on directly targeting Pale Moon">
+<!ENTITY addon.compatAddon                    "This add-on targeting Firefox, but compatible with Pale Moon">
+
 <!ENTITY addon.createdBy.label                "By ">
 
 <!ENTITY eula.title                           "End-User License Agreement">
--- a/toolkit/locales/en-US/chrome/mozapps/extensions/extensions.dtd
+++ b/toolkit/locales/en-US/chrome/mozapps/extensions/extensions.dtd
@@ -210,8 +210,8 @@
 <!ENTITY addon.loadingReleaseNotes.label      "Loading…">
 <!ENTITY addon.errorLoadingReleaseNotes.label "Sorry, but there was an error loading the release notes.">
 
-<!ENTITY addon.nativeAddon                    "This add-on directly targeting Pale Moon">
-<!ENTITY addon.compatAddon                    "This add-on targeting Firefox, but compatible with Pale Moon">
+<!ENTITY addon.nativeAddon                    "This add-on directly targets Pale Moon">
+<!ENTITY addon.compatAddon                    "This add-on targets Mozilla Firefox and runs in compatibility mode">
 
 <!ENTITY addon.createdBy.label                "By ">
 
--- a/toolkit/mozapps/extensions/content/extensions.css
+++ b/toolkit/mozapps/extensions/content/extensions.css
@@ -264,9 +264,9 @@ richlistitem:not([selected]) * {
 
 /* Indicator style for extension target application */
 .addon[native] .nativeIndicator {
-  font-size: 120%;
+  font-size: 150%;
   margin-top: -5pt;
-  margin-bottom: 1pt;
+  margin-bottom: -1pt;
 }
 .addon[native][active="false"] .nativeIndicator {
   opacity: 0.4;
--- a/toolkit/mozapps/extensions/content/extensions.css
+++ b/toolkit/mozapps/extensions/content/extensions.css
@@ -148,6 +148,8 @@ setting[type="menulist"] {
 .addon:not([notification="info"]) .info,
 .addon:not([pending]) .pending,
 .addon:not([upgrade="true"]) .update-postfix,
+.addon:not([native="true"]) .nativeAddon,
+.addon:not([native="false"]) .compatAddon,
 .addon[active="true"] .disabled-postfix,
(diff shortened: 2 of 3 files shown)


wolfbeast:
Why did you reduce the size of the indicator? I do not agree with that.

toolkit/locales/en-US/chrome/mozapps/extensions/extensions.dtd
```diff
@@ -210,6 +210,9 @@
 <!ENTITY addon.loadingReleaseNotes.label      "Loading…">
 <!ENTITY addon.errorLoadingReleaseNotes.label "Sorry, but there was an error loading the release notes.">
 
+<!ENTITY addon.nativeAddon                    "This add-on directly targeting Pale Moon">
```
wolfbeast: Broken English.

toolkit/locales/en-US/chrome/mozapps/extensions/extensions.dtd
```diff
@@ -210,6 +210,9 @@
...
+<!ENTITY addon.compatAddon                    "This add-on targeting Firefox, but compatible with Pale Moon">
```
wolfbeast: Broken English, and you shouldn't claim it is compatible - there's no guarantee!

toolkit/mozapps/extensions/content/extensions.xml
```diff
@@ -797,6 +797,8 @@
                 <xul:label anonid="name" class="name" crop="end" flex="1"
                            xbl:inherits="value=name,tooltiptext=name"/>
                 <xul:label anonid="version" class="version"/>
+				<xul:label class="nativeIndicator nativeAddon" value="•" tooltiptext="&addon.nativeAddon;"/>
```
wolfbeast: Whitespace!

toolkit/mozapps/extensions/content/extensions.css
```diff
@@ -261,14 +263,17 @@ richlistitem:not([selected]) * {
 }
 
 /* Indicator style for extension target application */
-.addon[native] .name-container::after {
-  content: " •";
-  font-size: 150%;
-  line-height: 75%;
+.addon[native] .nativeIndicator {
+  font-size: 120%;
```
wolfbeast: Why did you reduce the size?

wsdfhjxc:
Maybe something like this...

`Blue dot means that an extension natively supports Pale Moon.`

`Orange dot means that an extension supports Firefox, but could still be installed.`

toolkit/locales/en-US/chrome/mozapps/extensions/extensions.dtd
JustOff: "This add-on is natively supports Pale Moon" is ok?

toolkit/locales/en-US/chrome/mozapps/extensions/extensions.dtd
JustOff: "This add-on is made for Firefox, but can be run on Pale Moon" is ok?

toolkit/mozapps/extensions/content/extensions.xml
JustOff: Whitespace isn't required when dot is defined via xml, see [screenshot](https://vgy.me/vmaUCJ.png)

toolkit/mozapps/extensions/content/extensions.css
JustOff: I think the small dots look more neatly: [current](https://vgy.me/RLMU3k.png) vs [proposal](https://vgy.me/vmaUCJ.png). Am I wrong?

wsdfhjxc: I can't remember these dots being so big on Windows/Linux with several themes. Anyway, font-size of 120% seems to fit better regardless of applied theme.

toolkit/mozapps/extensions/content/extensions.xml
JustOff: Oops... s/tab/spase/g, got it, thanks @wsdfhjxc

toolkit/mozapps/extensions/content/extensions.css
wolfbeast: The dots aren't as big in the current release version of Pale Moon (screenshot on Win7), reducing the size further would make it too small.
dots1

toolkit/locales/en-US/chrome/mozapps/extensions/extensions.dtd
wolfbeast: "This add-on is directly targeting Pale Moon" or
"This add-on directly targets Pale Moon"

toolkit/locales/en-US/chrome/mozapps/extensions/extensions.dtd
wolfbeast: I'd say "This add-on targets Mozilla Firefox and runs in compatibility mode"

toolkit/mozapps/extensions/content/extensions.css
JustOff: It's not a problem to revert back to 150%, the only thing I can't understand why on @wolfbeast's screenshot the dots are twice smaller than I have (Win7, default theme too, screenshots are above) ...

JustOff: @Lootyhoof could you please say something about size of dots? Why they look so different within the same default theme on Win7?

Lootyhoof: @JustOff Likely due to font size of the `richlistitem` it's housed within. The glyph is using `150%` rather than a `pt` size, so is relative to the content it's within. If you change that then the glyph will change size also.

JustOff: Looks like something strange going on specifically on my PC, I can't still understand what is it and why. Ok, let's use size and margins from #741, but I can't say exactly what it will be, 'cause I have it look differently)

PR #750, opened 2016-12-06
Various FFmpeg fixes

trav90:
This PR fixes a few more problems with our FFmpeg support, most notably we will now throw an error instead of crashing if we try to use a codec that FFmpeg was not compiled with (i.e. with disable-nonfree).

Tested on Youtube, Twitch, and others and appears to work as intended.


wolfbeast:
Crashing when a codec wasn't found? Really? Amateurs.

Issue #34, opened 2015-03-19
Remove telemetry


wolfbeast:
Telemetry function calls have infected the entire source tree. Lots of dead weight that should NOT be there.

wolfbeast:
Majority of the leftover telemetry has been removed in commit 020b6ea6c588cfea20b3c9522ef4c25c8112a6bd

Thanks ketmar :)

wolfbeast:
Seems something broke:

Timestamp: 02/09/2015 16:23:08
Error: TypeError: BrowserSearch.recordSearchInHealthReport is not a function
Source File: chrome://browser/content/search/search.xml
Line: 478

PR #754, opened 2016-12-06
[suggestion]: Remove healthreport, datareport

janekptacijarabaci:
Only for the desktop client.

__Telemetry is more challenging. I could do at least part of this. If you don't plan it (soon).__

See also #34

---

__But please check it in detail.__


wolfbeast:
Telemetry is extremely invasive and will need to be handled extremely carefully. It can cause very hard-to-track bugs in the browser. Be aware that that is a very large task, and if you plan to undertake it, let me know; I'll make a telemetry-work branch in that case that you can work in instead of pulling against master, to keep risks to a minimum.

janekptacijarabaci:
> Telemetry is extremely invasive and will need to be handled extremely carefully. It can cause very hard-to-track bugs in the browser. Be aware that that is a very large task, and if you plan to undertake it, let me know; I'll make a telemetry-work branch in that case that you can work in instead of pulling against master, to keep risks to a minimum.

I would make at least some of the telemetrics. I don't know if it's enough to create a new branch...
I don't even know how much I make.

wolfbeast:
We've axed telemetry before (unfortunately it'll have to be done again on the platform side with the re-base) and maybe you can look through commit history to see what all was done so you have an idea if you can take #34 on or not.

janekptacijarabaci:
> if you can take #34

IMHO/AFAIK: An individual part can be used, e.g.:
TelemetryAccumulate
TelemetryStopwatch
RecordShutdown...
TelemetryTimestamps
TelemetryPing
ProcessedStack.h (maybe)

It may prove to be wrong, of course.

janekptacijarabaci:
...also `Crash Reporter` isn't completely cleared, see also:
https://github.com/MoonchildProductions/Pale-Moon/commit/b22832b4a82f22f9f9c3440e5ad5e87266f77660

I mentioned it only :-)

wolfbeast:
I'm going to close this PR for now but do keep it for reference. It's got a bunch of different things mixed in (crash reporter doesn't mean telemetry/health report). It should be done one step/component at a time when possible.

Sorry for not accepting it as-is; I appreciate the work, but I want to be super careful here -- it's broken the browser before when not doing it just perfect and slowly.

janekptacijarabaci:
> Sorry for not accepting it as-is; I appreciate the work, but I want to be super careful here -- it's broken the browser before when not doing it just perfect and slowly.

Of course, it's okay :-)

PR #782, opened 2016-12-19
Revive Browser Debugger

JustOff:
It actually works, but was lost in space)


wolfbeast:
Stupid unnecessary renaming.
And in what way is "toolbox" a better name than "debugger" for it being a debugger?... o.O

PR #783, opened 2016-12-22
Avoid TypeError in 'sdk/context-menu' caused by bad e10s code

JustOff:
Context menu support in e10s via https://bugzilla.mozilla.org/show_bug.cgi?id=1060138 led to the TypeError in FF38 - https://bugzilla.mozilla.org/show_bug.cgi?id=1130830

In FF39+ context-menu.js was rewritten from scratch with processes, but it's not our way.

Let's do extra check it for now, though to rollback #1060138 would be better.

### DIFF
--- a/toolkit/jetpack/sdk/context-menu.js
+++ b/toolkit/jetpack/sdk/context-menu.js
@@ -686,21 +686,22 @@ function remoteItemRequest({ target: { messageManager } }) {
   if (items.length == 0)
     return;
 
-  messageManager.sendAsyncMessage("sdk/contextmenu/createitems", {
-    items,
-    addon: ADDON,
-  });
-}
-
 // Context menu support in e10s via https://bugzilla.mozilla.org/show_bug.cgi?id=1060138
 // led to the TypeError in FF38 - https://bugzilla.mozilla.org/show_bug.cgi?id=1130830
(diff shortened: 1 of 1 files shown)


wolfbeast:
Rolling back 1060138 would be the way to go, here. There's no reason for the context menu to be async (and causing an added delay for the user) just because it's needed for e10s that we won't use anyway.

I'll merge this in as a quick fix, but can you please open a new issue for rolling back the bug and doing it properly?

JustOff:
We should probably first discuss in general what we are going to do with SDK. Are we going to provide only the compatibility for current Firefox addons (as I try to do now) or API for making the new ones. Despite the fact that I personally hate SDK-way of writing extensions the answer isn't so obvious even for me. And in both cases, I'm not sure that we can safely remove all content/frame related stuff from SDK without breaking interaction with the current browser core, this requires a deeper investigation. But even if we can, then I think that it would be better open corresponding global issue, and 'sdk/context-menu' is only a special case and likely not most important.

wolfbeast:
From an SDK point of view, rolling back this change won't make a difference.

JustOff:
Hmm ... in that case rolling back 1060138 looks easy enough, moreover as far as I see it will allow safely rollback #778 and #780.

PR #784, opened 2016-12-23
Remove bad e10s code from 'sdk/context-menu'

JustOff:
Roll back https://bugzilla.mozilla.org/show_bug.cgi?id=1060138
Roll back #778 and #780 (no longer required), also supersedes #783


JustOff:
We should probably also rollback https://bugzilla.mozilla.org/show_bug.cgi?id=1023326, thus we could completely get rid of e10s frame scripts in SDK.

wolfbeast:
Bug # 1070952 was backed out (it's an open bug)

JustOff:
Unfortunately 1070952 was not backed out actually and it's present in our codebase, so we should deal with it by yourself.

wolfbeast:
I see. I guess it was in the Mozilla tree for a while then before they backed it out.

PR #785, opened 2016-12-23
Remove e10s code from 'sdk/tabs'

JustOff:
Roll back https://bugzilla.mozilla.org/show_bug.cgi?id=1015621 (tab-firefox.js related part)
Roll back https://bugzilla.mozilla.org/show_bug.cgi?id=1073978
Roll back https://bugzilla.mozilla.org/show_bug.cgi?id=1023326


wolfbeast:
LGTM!

JustOff:
Looks good, but don't work properly... I need to do some more tests, but probably I will ask for the revert. It seems that from a certain time SDK emits events from documents only to frame scripts and there are too much dependencies that prevents the safe rebase.

wolfbeast:
All right. I'll back this out now then. See if it can be done another way and if not we'll just have to live with framscripts.

JustOff:
I'll try it once again after some time, but I need to take a pause to take a fresh look.

Just for the reference, here are the other e10s related updates that we might also want to rollback:
1) https://bugzilla.mozilla.org/show_bug.cgi?id=1116004
2) https://bugzilla.mozilla.org/show_bug.cgi?id=1058698

PR #789, opened 2016-12-26
Sync toolkit/loader with FF50

JustOff:
The initial goal was to backport [#1309350](https://bugzilla.mozilla.org/show_bug.cgi?id=1309350) and [#1309351](https://bugzilla.mozilla.org/show_bug.cgi?id=1309351), but a lot of useful updates were found, so let's do the maximum possible sync.

1) Import from Firefox trunk as of 160608 (after [#1170258](https://bugzilla.mozilla.org/show_bug.cgi?id=1170258))
2) Roll back [#1196975](https://bugzilla.mozilla.org/show_bug.cgi?id=1196975)
3) Fix path to Console.jsm
4) Apply [#1309350](https://bugzilla.mozilla.org/show_bug.cgi?id=1309350)
5) Rewrite DefaultMap in ES5 (without 'class')
6) Use forEach instead of for-of (separate lexical scopes for for-of iterations require FF51 core)
7) Apply [#1309351](https://bugzilla.mozilla.org/show_bug.cgi?id=1309351)

The result was tested on [forum](https://forum.palemoon.org/viewtopic.php?p=102075#p102075), SDK-based addons are loading in about five times faster.

### DIFF
--- a/toolkit/jetpack/bootstrap.js
+++ b/toolkit/jetpack/bootstrap.js
@@ -299,8 +299,11 @@ function loadSandbox(uri) {
 }
 
 function unloadSandbox(sandbox) {
-  if ("nukeSandbox" in Cu)
-    Cu.nukeSandbox(sandbox);
+  if ("nukeSandbox" in Cu) {
+    try {
+      Cu.nukeSandbox(sandbox);
+    } catch (e) {}
+  }
 }
 
 function setTimeout(callback, delay) {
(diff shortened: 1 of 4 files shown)


wolfbeast:
These modules are used by devtools -- did you make sure to also test that devtools don't break?

JustOff:
Yes, devtools works properly with this loader.

janekptacijarabaci:
> Yes, devtools works properly with this loader.

It seems that yes.

@wolfbeast
But... You add the label `Devtools`, please.

PR #795, opened 2016-12-30
Allow PMkit add-ons created with JPM

JustOff:
Allow install and run add-ons that contain package.json and directly target to Pale Moon.

### DIFF
--- a/toolkit/mozapps/extensions/internal/XPIProvider.jsm
+++ b/toolkit/mozapps/extensions/internal/XPIProvider.jsm
@@ -1102,6 +1102,8 @@ function loadManifestFromZipReader(aZipReader) {
       } else {
         addon.wrongSDK = true;
       }
+    } else {
+      addon.wrongSDK = false;
     }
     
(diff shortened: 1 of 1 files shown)


wsdfhjxc:
@JustOff, there are still [warning message](https://github.com/MoonchildProductions/Pale-Moon/blob/8dac196100b7266d07bc0cac8e9c973e8a6a2a8b/toolkit/mozapps/extensions/content/extensions.xml#L1300) and [installation error](https://github.com/MoonchildProductions/Pale-Moon/blob/5214789d035f4e4c08899a6e281116520f605ec6/browser/base/content/browser-addons.js#L168) leftovers referring to the old property.

wolfbeast:
Backing this out. renaming the property wasn't a good idea.

JustOff:
The bad idea was to post PR between :christmas_tree: parties :wine_glass: :stuck_out_tongue_winking_eye:
I'm :pray: about that!

I'm taking a break until :santa: will not :rocket:

Happy New Year! :bell:

:rooster:

PR #805, opened 2017-01-02
Adjust some media prefs

trav90:
Some cleanup now that we've got FFmpeg working and enabled. Changes:

1) Remove "fragmented-mp4" from pref names -- Not necessary as we can play both fragmented and regular MP4's. Also, it shortens and simplifies some of the pref names to boot.

2) Put platform-specific media pref caches behind appropriate #ifdef's.

3) Make the "media.ffmpeg.enabled" pref actually work when set to "false" (in which case GStreamer will be used again :nauseated_face: )

4) Add a test and fatal error to configure if trying to build FFmpeg support without FMP4 support.

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -288,7 +288,7 @@ pref("media.decoder.heuristic.dormant.timeout", 60000);
 pref("media.directshow.enabled", true);
 #endif
 #ifdef MOZ_FMP4
-pref("media.mp4.enabled", true);
+pref("media.fragmented-mp4.enabled", true);
 #endif
 // Specifies whether the PDM can create a test decoder that
 // just outputs blank frames/audio instead of actually decoding. The blank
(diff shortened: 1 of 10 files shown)


modules/libpref/init/all.js
```diff
@@ -284,25 +284,25 @@ pref("media.play-stand-alone", true);
...
-#endif
 #ifdef MOZ_DIRECTSHOW
 pref("media.directshow.enabled", true);
 #endif
 #ifdef MOZ_FMP4
-pref("media.fragmented-mp4.enabled", true);
-pref("media.fragmented-mp4.ffmpeg.enabled", true);
-pref("media.fragmented-mp4.gmp.enabled", false);
-// Specifies whether the fragmented MP4 parser uses a test decoder that
+pref("media.mp4.enabled", true);
```
wolfbeast: Isn't this too ambiguous? Or does it actually completely disable any and all MP4 decoding when set to false?

trav90: Yes, it disables the MP4Decoder completely (including with FFmpeg, GStreamer, etc).

trav90:
I gave this some more thought and do think it'd be better to leave the media.fragmented-mp4.enabled pref name as-is. While disabling this pref does break MP4 decoding, it specifically only disables the FMP4 decoder, not MP4 decoding as a whole.

trav90:
New commits have been added.

PR #806, opened 2017-01-03
Allow PMkit add-ons created with JPM

JustOff:
Allow install and run add-ons that contain package.json and directly target to Pale Moon.


JustOff:
Supersedes #795

PR #811, opened 2017-01-07
Allow access to 'sdk/ui/buttons' via require('sdk/ui')

JustOff:
Some add-ons use require('sdk/ui') to access 'sdk/ui/buttons/*', we should support it.

### DIFF
--- a/toolkit/jetpack/sdk/ui.js
+++ b/toolkit/jetpack/sdk/ui.js
@@ -6,12 +6,12 @@
 module.metadata = {
   'stability': 'experimental',
   'engines': {
-    'Firefox': '> 28'
+    'Firefox': '> 27'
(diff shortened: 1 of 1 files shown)


wolfbeast:
Won't this break other add-ons that want to access ui.Sidebar, ui.Frame and ui.Toolbar (since you're commenting those out)?

JustOff:
As current state of PMkit the attempt to require 'ui/sidebar', 'ui/toolbar' or 'ui/frame' leads to runtime error - they are simple inconsistent with Pale Moon. I could write the shims for them too, but can't find a single extension that actually uses them. So I think it's safe to accept this PR as is.

wolfbeast:
Works for me, then!

PR #819, opened 2017-01-09
Set release notes URL in About dialog based on channel

JustOff:
Original code is from [about.xhtml](http://xref.palemoon.org/palemoon-trunk/source/toolkit/content/about.xhtml) / [about.js](http://xref.palemoon.org/palemoon-trunk/source/toolkit/content/about.js)

PS: This dialog should also be properly localized, I'll prepare a separate PR.

### DIFF
--- a/browser/base/content/aboutDialog.js
+++ b/browser/base/content/aboutDialog.js
@@ -56,6 +56,15 @@ function init(aEvent)
   window.sizeToContent();
   window.moveTo((screen.availWidth / 2) - (window.outerWidth / 2), screen.availHeight / 5);
 #endif
+
+// get release notes URL from prefs
+  var formatter = Components.classes["@mozilla.org/toolkit/URLFormatterService;1"]
+                            .getService(Components.interfaces.nsIURLFormatter);
+  var releaseNotesURL = formatter.formatURLPref("app.releaseNotesURL");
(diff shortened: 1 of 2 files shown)


wolfbeast:
> PS: This dialog should also be properly localized, I'll prepare a separate PR.

Sorry, but it should not. I've mentioned this before. The about box should not be localized since Pale Moon is an English-language product. What people do to change their UI language doesn't change the fact that "about this software" should be in its original form at all times. The parts that matter for functionality (for updates) are already localized.

JustOff:
Even [these lines](https://github.com/JustOff/Pale-Moon/blob/fca3f1c776c28a81c94c126a3c8b600fbffa7aeb/browser/base/content/aboutDialog.xul#L107)?
```
<label class="text-link bottom-link" href="about:license">Licensing information</label>
<label class="text-link bottom-link" href="about:rights">End-user rights</label>
<label class="text-link bottom-link" id="releaseNotesURL">Release notes</label>
```
They are localized in all other places and we already have the first two in chrome://browser/locale/aboutDialog.dtd, only the last one is missed.

wolfbeast:
Yes, especially those lines. What do they link to? English pages. I'm NOT going to localize the website as well. This was localized at Mozilla because they have everything localized.

JustOff:
Thanks, got it.

PR #787, opened 2016-12-25
Remove 'sdk/context-menu@2'

JustOff:
'sdk/context-menu@2' was introduced by https://bugzilla.mozilla.org/show_bug.cgi?id=1070952 for e10s, never used, undocumented, no dependencies with the rest of code.

PR #820, opened 2017-01-10
Align PMkit with commonly used path to SDK

JustOff:
Move jetpack tree to commonjs path and update all corresponding URI.

This reverts https://github.com/MoonchildProductions/Tycho/commit/be7b03857d0fb53497d7832daae5deebacec33a8 (tests were removed earlier).

The reasons are explained in [this post](https://forum.palemoon.org/viewtopic.php?p=103190#p103190) and around.


wolfbeast:
Why did you physically move the files? the whole point would be that that would not be necessary with an abstract path mapping.

JustOff:
Sorry, maybe I missed something, let's look together once again:

1) _jpm_ by default builds _bootstrap.js_ that is looking for the loader in _resource://gre/modules/commonjs/_
2) The loader uses relative path from itself to load all following PMkit modules
3) By moving PMkit physically we satisfy all this requirements
4) We can safely align all our code to the new location of PMkit

Do you mean that mapping the path via _chrome.manifest_ in the root of _omni.jar_ is more preferred than moving PMkit physically? Currently this manifest has no single mapping and I thought that keeping old physical path and adding the extra rules could only lead to the confusion in the future. Am I wrong?

wolfbeast:
I guess it doesn't matter. merging.

JustOff:
Could you please also add PMkit label to #776, #777, #778, #781, #783, #784, #785, #787, #788, #789, #791 and #818?

wolfbeast:
Why is this so important to you? I don't see a point in back-labeling everything, especially things that are already closed or merged. I have better things to do with my time.

JustOff:
It's not important for me at all. I just saw that you have added PMkit tag, but the assignment is incomplete. You are free to ignore the additions, if it's not important to you.

janekptacijarabaci:
The same goes for label `Devtools` (etc.). It was "a plea", for clarity (I'll leave it in PR), but it's up to you.

PR #824, opened 2017-01-12
Update Soundcloud SSUAO

JustOff:
https://forum.palemoon.org/viewtopic.php?p=103414#p103414

I had the same problem in Firefox mode + MSE enabed.

Works for me either in Native or Gecko mode (regardless of MSE), set to Native.


JustOff:
https://github.com/MoonchildProductions/Pale-Moon/commit/414543ddc9c987de039805df8422d00d31b4fb3f

Oops, looks like we did it in parallel ;-)

PR #833, opened 2017-01-14
Add (and use) new MP3Demuxer/Decoder

trav90:
This PR adds a new MP3 demuxer and decoder that is based on/tied in with the MediaFormatReader/MediaDataDemuxer interfaces that were introduced some time ago. This drops our dependence on external readers for playing MP3, and on Linux will allow us to do away with GStreamer completely.

Verified that it builds and plays MP3 files in various places around the net. Also tested using this test page (among others): http://hpr.dogphilosophy.net/test/. Also tested using the internal Mozilla gtests, where is also passed.


trav90:
While Linux will see the most benefit from this (being able to remove GStreamer support), this code will also be used on Windows as well.

Removing the Linux label as it is cross-platform/

trav90:
Added new commit that helps prevent random and intermittent issues when seeking.

In my testing on Linux, there is no noticeable change in behavior when playing MP3 files in 27.0.3 using GStreamer or in current trunk + this PR & Gstreamer disabled.

wolfbeast:
In my testing on Windows, apart from the issue found in #841, there is no regression noticeable; playing and seeking in both CBR and VBR MP3 files seems to work as-intended.

Issue #831, opened 2017-01-13
27.1.0b2 Windows: Dailymail video playback issues because of media.fragmented-mp4.enabled


SJ-Diamond:
http://www.dailymail.co.uk/video/index.html?ito=embedded

Videos there fail under Windows if the about:config value for "media.fragmented-mp4.enabled" is true - Setting it to false is no option because it makes the memory usage of the browser sky-rocking.

There seems to be some large issues with that feature/function which makes Mp4 playback either a hit or a miss.

With that one disabled other problematic Mp4 videos are also running... at least until memory runs out ;)

Seems Mozilla has done something after ESR 38 was done in that department which made Mp4 playback more problem free but which is still missing or not working as it should in your implementation?

System is 8.1 with 64 Bit

trav90:
While I do have issues playing videos on 27.0.3, They play perfectly for me on current 27.1.0b2 builds.

JustOff:
Just for the record, I can't play any videos from the link above (27.0.1b2, win7).

SJ-Diamond:
@JustOff you will see it will work if you set "media.fragmented-mp4.enabled" to false in about:config - but then Pale Moon consumes all memory available until Windows complains nothing is left anymore.

So there is indeed something either missing or not complete or bugged with that special value and it's surrounding. Because it is clearly visible that MP4 videos - even problematic are still playing... only not with that pref on.

wolfbeast:
I can confirm that the latest trunk does not play the dailymotion videos with the fragmented mp4 reader enabled. I do however not see any memory inflation with it disabled.

I can see if I can find out where this regressed, @trav90

wolfbeast:
Looks like this broke with PR #707

SJ-Diamond:
@wolfbeast Well, here under Windows 8.1 64 Bit every second where i play MP4 videos without media.fragmented-mp4.enabled set to true, it consumes more and more memory until Windows has no free memory anymore ;)

wolfbeast:
I don't think the memory inflation issue is directly related.

wolfbeast:
I've narrowed it down to commit 7f2dd37de3120bdf4347c9a5c5adbb477518eb85 -- after that commit, the types of MP4 found on dailymail (I've saved one locally) stop playing.

wolfbeast:
Considering this breaks fragmented MP4 playback of a good percentage of videos on the web, this is considered a blocker. A solution must be found before the next release version for this.

[...]

PR #838, opened 2017-01-15
Fix problems with MP4 playback

trav90:
Fixes issue #831.


JustOff:
Thanks, with this update Dailymail videos play fine for me.

wolfbeast:
Today's unstable build includes this fix.

SJ-Diamond:
Thanks, works great again. It really looks like MP4 Windows video playback in general has improved, also problematic videos are now running and seeking seems not so slow anymore.

PR #845, opened 2017-01-18
Remove redundant downloads indicator styling

### LINKED ISSUE #695: Downloads button: Text colour issue (Tycho)
The text for the time remaining counter, on the Downloads button in the Navigation Toolbar, doesn't detect theme colors.
With my current settings, the text is almost impossible to make out!

(Pale Moon v27.0.0b3,x64 / Windows 8.1 Pro / ColorizationColorBalance = 0)

Lootyhoof:
As part of #437 I added `[brighttext]` attributes to the toolbars. As it turns out, I forgot to remove the legacy download indicator styling which was in use before this (whoops!). This removes that, which in turn also improves the appearance of the indicator on other Windows platforms, as this legacy styling was designed primarily for Windows 7 Aero Glass.

Fixes #695.

### DIFF
--- a/browser/themes/windows/downloads/downloads.css
+++ b/browser/themes/windows/downloads/downloads.css
@@ -447,40 +447,6 @@ toolbar[mode="full"] > #downloads-indicator > .toolbarbutton-text {
   text-align: center;
 }
 
-@media (-moz-windows-compositor) {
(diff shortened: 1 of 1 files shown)


wolfbeast:
Good catch!

PR #847, opened 2017-01-20
Add a type-erased callable wrapper (mozilla::Function) to MFBT.

trav90:
This PR adds a new function that a media-related patch depends on. Mozilla explains this better than I can, so deferring to BMO on this:

[Bug 1198451](https://bugzilla.mozilla.org/show_bug.cgi?id=1198451)
[Bug 1212745](https://bugzilla.mozilla.org/show_bug.cgi?id=1212745)

Verified builds and appears to work.


wolfbeast:
This is out of my area of expertise. I've never fully understood c++ template overloading.
Since the patches use specifics for MSVC2013 it should be safe to apply on our tree, but I'll have to have a look how this pans out with a build.

wolfbeast:
Uplifted to 27.1_RelBranch

PR #848, opened 2017-01-20
Piggyback Quicktime support onto MP4Decoder

trav90:
On Linux (where the Quicktime plugin is not available), .mov files are played using GStreamer. Since mov files are in most cases essentially an MP4, we can use our MP4Decoder to play them. This PR will allow Linux users to play mov files in the browser without needing GStreamer.

The PR is dependent on the new mozilla::Function in PR #847 and will not compile without it.

Built and tested together with #847 and appears to work as intended.

### DIFF
--- a/dom/media/DecoderTraits.cpp
+++ b/dom/media/DecoderTraits.cpp
@@ -360,8 +360,7 @@ IsMP4SupportedType(const nsACString& aType,
 #ifdef MOZ_OMX_DECODER
   return false;
 #else
-  bool haveAAC, haveMP3, haveH264;
-  return MP4Decoder::CanHandleMediaType(aType, aCodecs, haveAAC, haveH264, haveMP3);
+  return MP4Decoder::CanHandleMediaType(aType, aCodecs);
 #endif
 }
 #endif
--- a/dom/media/VideoUtils.cpp
+++ b/dom/media/VideoUtils.cpp
@@ -13,8 +13,11 @@
 #include "mozilla/Preferences.h"
 #include "mozilla/Base64.h"
 #include "mozilla/Telemetry.h"
+#include "mozilla/Function.h"
 #include "nsIRandomGenerator.h"
 #include "nsIServiceManager.h"
+#include "nsCharSeparatedTokenizer.h"
+#include "nsContentTypeParser.h"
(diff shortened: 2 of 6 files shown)


wolfbeast:
People still use Quicktime? XD

JustOff:
> People still use Quicktime? XD

A lot of old cameras still produce mov files, but yes, they are actually mp4.

trav90:
> People still use Quicktime? XD

lol I know right? XD

trav90:
Forgot to mention, I'd like to see this land in 27.1 if possible. With this patch, GStreamer will be no longer be used for anything.

wolfbeast:
If OK I'll uplift these to the relbranch.

wolfbeast:
I can't even find a single mov file on the net to test this right now...

wolfbeast:
Oh, and apparently on Windows it'll default to the plugin anyway even if we could theoretically play it?

wolfbeast:
Anyway, no build issues here. Can't play mov files to check so... *shrug*

wolfbeast:
Uplifted to 27.1_RelBranch

trav90:
> I can't even find a single mov file on the net to test this right now...

Less than a year ago Mozilla received so many complaints/bug reports that they considered this a blocker for one of ESR45's point releases (hence the reason I even looked into this to begin with). Didn't take long for .mov to pretty much disappear.
Here's a couple links to basic mov files (one video & one audio only):
http://dl1.streamzilla.jet-stream.nl/demo/quicktime.mov
http://dl1.streamzilla.jet-stream.nl/preludecd/okt_15_bach_kunst_der_fuge_contrapunctus_I.mov

> Oh, and apparently on Windows it'll default to the plugin anyway even if we could theoretically play it?

Exactly. There should be no change in behavior on Windows and Mac where the Quicktime plugin is available and is still required. These changes simply allow our MP4Decoder to play the mov files it supports when on Linux.

The primary thing that needs testing/verification are the changes in parsing logic for MP4.

wolfbeast:
The parsing logic looks good. I removed the conditional for linux only for testing, and it pops up the internal player (won't play of course, since it's not hooked up on Windows) so the MP4 meta parsing logic seems to be OK.

PR #850, opened 2017-01-21
FFmpeg cash fix: Initialize PDM before attempting to use it

trav90:
This will prevent a crash should FFmpeg try to use the PDM before it is initialized.


wolfbeast:
Uplifted in 8074574

trav90:
Thanks!

Issue #849, opened 2017-01-20
Improve toolbar button visuals for brighttext situations


wolfbeast:
As noted in #695, the gradient on toolbar buttons in the UI is less than optimal in `[brighttext]` situations (the gradient lightens button faces too much on dark colors to make light text stand out).

This could be addressed by making the gradient on toolbar buttons darker to blend in better with dark accent colors and improving contrast with white text as a result.

Lootyhoof:
This is something I've been looking into lately. How does this look?

![Brighttext buttons](http://i.imgur.com/8IVg6U2.png)

I've yet to test it on Win7 et al still, though.

wolfbeast:
I think it looks fine - it's a special-case button anyway if it works like an indicator, so special-case styling when it's active in that role would be fine.

wolfbeast:
Considering this resolved.

PR #862, opened 2017-01-28
Improve downloads indicator in brighttext situations

Lootyhoof:
Ref #849 

This doesn't currently have a hover or open state. I'm not entirely sure why it won't accept them, to be honest (as I've tried them to no avail).

Additionally, this background styling only takes effect on `#nav-bar`, as this is the only place that buttons have backgrounds to them.


wolfbeast:
It's fine with its current limitations - it'll just be a quirk of the default theme.
Thanks for your time and effort!

PR #869, opened 2017-01-30
CORS: A "data:" URL = the same-origin

janekptacijarabaci:
__Steps to reproduce:__
Go to: `http://codepen.io/simonsarris/pen/raxXZa`

Actual results:
Throws an error in Browser Console:
`NS_ERROR_NOT_AVAILABLE: pen.js:22:0`

Expected results:
PM allows the drawing of the canvas.

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1109693
https://bugzilla.mozilla.org/show_bug.cgi?id=1157451

See also:
https://forum.palemoon.org/viewtopic.php?f=29&t=13005&p=104943

---

### DIFF
--- a/dom/base/EventSource.cpp
+++ b/dom/base/EventSource.cpp
@@ -796,7 +796,7 @@ EventSource::InitChannelAndRequestEventSource()
 
   nsRefPtr<nsCORSListenerProxy> listener =
     new nsCORSListenerProxy(this, mPrincipal, mWithCredentials);
-  rv = listener->Init(mHttpChannel);
+  rv = listener->Init(mHttpChannel, DataURIHandling::Allow);
   NS_ENSURE_SUCCESS(rv, rv);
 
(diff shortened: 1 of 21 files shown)


JustOff:
Probably you should consider to uplift this to 27.1, because [it seems to solves the problem with reloading on Facebook](https://forum.palemoon.org/viewtopic.php?p=104943#p104943).

wolfbeast:
See the forum thread.
The funny thing is: I went over the HTML spec and CORS fetch spec, also because there was a reference to it in one of the BMO bugs, but whatever was previously in there that was referred to, is not in the spec!

So this means treating data: as same-origin is not defined and our behavior was not wrong. Since it's undefined, I'm fine with treating data: as trusted, though, since it makes sense (for a change).

wolfbeast:
Uplifted in [d755156]

PR #873, opened 2017-02-02
Fix SSUAO & Navigator::GetUserAgent() regression

JustOff:
If general.useragent.override is defined, then Navigator::GetUserAgent() ignores SSUAO and always uses value from general.useragent.override. 

This problem did not exist in the Pale Moon 26.x. The regression probably caused by that Mozilla no longer considers SSUAO.

### DIFF
--- a/dom/base/Navigator.cpp
+++ b/dom/base/Navigator.cpp
@@ -2576,17 +2576,7 @@ Navigator::GetUserAgent(nsPIDOMWindow* aWindow, nsIURI* aURI,
                         nsAString& aUserAgent)
 {
   MOZ_ASSERT(NS_IsMainThread());
-/*
-  if (!aIsCallerChrome) {
-    const nsAdoptingString& override =
-      mozilla::Preferences::GetString("general.useragent.override");
 
-    if (override) {
-      aUserAgent = override;
-      return NS_OK;
-    }
-  }
-*/
   nsresult rv;
   nsCOMPtr<nsIHttpProtocolHandler>
     service(do_GetService(NS_NETWORK_PROTOCOL_CONTRACTID_PREFIX "http", &rv));
--- a/dom/base/Navigator.cpp
+++ b/dom/base/Navigator.cpp
@@ -2576,7 +2576,7 @@ Navigator::GetUserAgent(nsPIDOMWindow* aWindow, nsIURI* aURI,
                         nsAString& aUserAgent)
 {
   MOZ_ASSERT(NS_IsMainThread());
-
+/*
   if (!aIsCallerChrome) {
     const nsAdoptingString& override =
       mozilla::Preferences::GetString("general.useragent.override");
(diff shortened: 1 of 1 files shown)


mattatobin:
Can you better explain your patch cause it looks to just comment out a huge chunk of code.. Will the global UA override still work in instances where ssuao aren't applicable?

JustOff:
> Will the global UA override still work in instances where ssuao aren't applicable?

Yes, I have tested all possible variants.

wolfbeast:
If you want to propose a change, please don't comment out code unless it is going to be a temporary band-aid that will need to be reverted later. Actually removing the lines will make the diff a lot more readable in terms of what is changed, and won't leave dead code behind ;)

JustOff:
Ok, I'll post proper update in a few minutes.

wolfbeast:
I had a quick look, and it does seem that the general (global) override would no longer work with this since it'll fall back to the default UA as determined by the application, not what people have written in the pref (general.useragent.override).

If I follow this correctly, it would now call to `SiteSpecificUserAgent.js` in /dom/base, which pulls the default UA from `HTTP_PROTO_HANDLER.userAgent` where HTTP_PROTO_HANDLER is the `nsIHttpProtocolHandler` service/interface, which will construct the default UA. So the global override would no longer work if you remove these lines.

wolfbeast:
Actually... it may work. seems what is removed in this PR is just a short-circuit. nsHttpHandler::UserAgent() still honors `mUserAgentOverride` which is kept updated with a pref observer.

http://xref.palemoon.org/palemoon-trunk/source/netwerk/protocol/http/nsHttpHandler.cpp#599

JustOff:
As far as I understand [siteSpecificUA->GetUserAgentForURIAndWindow(aURI, aWindow, aUserAgent);](https://github.com/JustOff/Pale-Moon/blob/4871d51a79de5b63c0d451c0dcf13b738c14fe09/dom/base/Navigator.cpp#L2607) does all the job the same way as [in PM26.x](http://xref.palemoon.org/palemoon-rel26/source/dom/base/Navigator.cpp#333).  I have built a test with this PR and all works as expected.

wolfbeast:
The question isn't whether the site-specific ones work or not. The question is whether `general.useragent.override` (without a site) is still honored. if there is no sitespecific one, it falls back to using the nsHTTPHandler to get the useragent:

````
38   getUserAgentForURIAndWindow: function ssua_getUserAgentForURIAndWindow(aURI, aWindow) {
39     if (this.inParent) {
40       return UserAgentOverrides.getOverrideForURI(aURI) || HTTP_PROTO_HANDLER.userAgent;
41     }
````
No ss => use `HTTP_PROTO_HANDLER.userAgent`
That calls userAgent(), and that in turn calls buildUserAgent() -- but thankfully Mozilla didn't do their job properly and the general override there is still in place (basically they had it twice, and what you remove here is the shortcut at the higher level)

JustOff:
I can't say that I followed the whole chain, but I already see where my assumption was wrong. Thanks for resolving this regression.

JustOff:
Could you please consider to uplift this?

wolfbeast:
I'll uplift once verified everything works as-intended.
Last minute uplift; already in the RelEng stage.

wolfbeast:
Verified working correctly with general-only, general+SSUAO and SSUAO-only.

wolfbeast:
Uplifted in [48b41af]

JustOff:
Thank you!

Issue #871, opened 2017-02-01
Add support for Element.insertAdjacentText() and Element.insertAdjacentElement()


wolfbeast:
Since this originally-IE-only sugar for Element.insertAdjacentHTML() has (unfortunately!) made it into the HTML LS in 2016, we should add this for Web Compat and Standards Compliance reasons, even if practical usage is very low.

BMO ref: https://bugzilla.mozilla.org/show_bug.cgi?id=811259

wolfbeast:
Resolved by #879.

PR #879, opened 2017-02-07
Add support for Element.insertAdjacentText() and Element.insertAdjacentElement()

janekptacijarabaci:
Ad #871 (https://forum.palemoon.org/viewtopic.php?f=29&t=14663)

__Steps to reproduce:__
Go to: https://www.dotnetperls.com/

Throws an error in Browser Console:
```
TypeError: b.insertAdjacentText is not a function
i.js:1:331
```
---

See also:
https://w3c-test.org/dom/nodes/insert-adjacent.html
http://tests.caniuse.com/?feat=insert-adjacent

__But... Please you check tests (mainly the file: `interfaces.html.ini`).__

### DIFF
--- a/dom/base/Element.cpp
+++ b/dom/base/Element.cpp
@@ -3378,6 +3378,53 @@ Element::InsertAdjacentHTML(const nsAString& aPosition, const nsAString& aText,
   }
 }
 
+nsINode*
+Element::InsertAdjacent(const nsAString& aWhere,
+                        nsINode* aNode,
+                        ErrorResult& aError)
+{
+  if (aWhere.LowerCaseEqualsLiteral("beforebegin")) {
+    nsCOMPtr<nsINode> parent = GetParentNode();
+    if (!parent) {
+      return nullptr;
+    }
+    parent->InsertBefore(*aNode, this, aError);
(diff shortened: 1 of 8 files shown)


wolfbeast:
LGTM, the interfaces html has 2 tests added for the two new functions, and the numbers were increased by 2, so that's correct AFAIU.

wolfbeast:
> I've created the new build (x32, Windows) and tested.

And yet that wouldn't have succeeded with the RefPtr/nsRefPtr mixup in your patch (see #880 and #881)... Are you sure your test builds are done correctly?

trav90:
> And yet that wouldn't have succeeded with the RefPtr/nsRefPtr mixup in your patch

The last couple times the RefPtr/nsRefPtr mixup has happened, it didn't affect Windows. I don't know why MSVC doesn't choke on it like GCC does.

janekptacijarabaci:
@wolfbeast 
> Are you sure your test builds are done correctly?

I am sure.

> => The last couple times the RefPtr/nsRefPtr mixup has happened, it didn't affect Windows. I don't know why MSVC doesn't choke on it like GCC does.

@trav90 
But thank you for your quick patch...

Issue #883, opened 2017-02-08
Undeclared identifiers used in an assertion in js/src/gc/Marking.cpp break debug build


rhinoduck:
Introduced in: https://github.com/MoonchildProductions/Pale-Moon/commit/daeb68a71fdd13308865e86a2269a892a89b12fb

Undelared identifiers: `ThingIsPermanentAtomOrWellKnownSymbol`, and potentially `js::Zone::isSelfHostingZone` (at least that was what the second error the compiler produced contained)

Offending lines highlighted: https://github.com/MoonchildProductions/Pale-Moon/blob/daeb68a71fdd13308865e86a2269a892a89b12fb/js/src/gc/Marking.cpp#L158-L160

rhinoduck:
You might want to merge or at least look at https://github.com/MoonchildProductions/Pale-Moon/pull/884 before doing something about this.

wolfbeast:
I just copied the assertion, not checking for stupid renames.

https://bugzilla.mozilla.org/show_bug.cgi?id=1147670#c6
`ThingIsPermanentAtom` -> `ThingIsPermanentAtomOrWellKnownSymbol`

I need to see the compiler output to know what exactly your second error would be.

wolfbeast:
Never mind, looks like another wrapper function that got added, and we don't have that in js::zone -- it just needs to call the runtime directly.

PR #884, opened 2017-02-08
Fix line endings in js/src/gc/Marking.cpp

rhinoduck:
Noticed Windows line endings in the diff of https://github.com/MoonchildProductions/Pale-Moon/commit/daeb68a71fdd13308865e86a2269a892a89b12fb while investigating https://github.com/MoonchildProductions/Pale-Moon/issues/883


wolfbeast:
Oh bah. Thanks for catching this!

Issue #105, opened 2015-06-17
Add JPEG-XR support


wolfbeast:
As an alternative and standardized format for images, JPEG-XR (extended range jpeg, especially suitable for HD photography, wide-gamut images, and a superior compression format compared to standard JPEG) would be good to have. This format was rejected by Mozilla based on mostly political reasoning. Adding this alongside WebP should provide maximum choice for Pale Moon users.
A basic implementation with decode-after-load should be straightforward.

On hold for the moment as I communicate with the original coder and/or MS OpenTech to get a suitable implementation for our code base.

wolfbeast:
Clearing the milestone for now since after initial communication, both the dev and MSOT have become completely unresponsive. :-1: 
Do they want their format to be adopted, or not...?

wolfbeast:
I got a reply from MSOT, and a repo was committed for the decoder as-written for them, and released to the public that way, building on jxrlib and published under the Community Promise.
This means we can go forward. I'm keeping myself assigned to this since adding decoders is still fresh in memory.

wolfbeast:
Well, it looks like the code dump supplied by MSOT is mostly useless.
The build system files are incomplete, e.g. there are NO symbol&header exports listed, and trying to add them manually, both by adopting the library properly in /media/libjxr and as-supplied in /image fails with unresolved externals. I've wasted 2 days on this and can't get it to work, so maybe someone with the appropriate knowledge can look at this.

Repo of jxr FF code:
https://github.com/wolfbeast/ff-jpegxr
Integration commit (separate branch): Commit: 307b09bcba21f42fdd4cfac196182db9efe855ec

wolfbeast:
Clearing milestone - doesn't look like the existing files are of any help (including the VSO file dump I received end of Aug). Looks like it will not make it into 26.0

wolfbeast:
if we want this, it will have to be a completely new implementation.

wolfbeast:
Bounty-candidate.

rhinoduck:
I couldn't sleep last night so I took a look at this. Putting the code together was messy business, but once I actually got it to compile, I was met with a pleasant surprise of an image loading fine on the first run.

[...]

PR #895, opened 2017-02-12
Silence compiler warnings in JXR decoder

rhinoduck:
As discussed in #105.

I didn't compile with GCC so maybe @trav90 will want to confirm they are all gone first (or maybe it just isn't worth a separate compile).

### DIFF
--- a/image/decoders/nsJXRDecoder.cpp
+++ b/image/decoders/nsJXRDecoder.cpp
@@ -147,6 +147,7 @@ nsJXRDecoder::nsJXRDecoder(RasterImage* aImage, bool hasBeenDecoded) : Decoder(a
     m_numAvailableBands(0),
     m_currentTileRow(0),
     m_currentSubBand(0),
+    m_startedDecodingSubband(false),
     m_skippedTileRows(false),
     m_alphaBitDepth(0),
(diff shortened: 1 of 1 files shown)


trav90:
I'll kick off a build now.

trav90:
Looks good here! Only 2 warnings left:

https://gist.github.com/trav90/93c35e099b4921ee8628bf8724604d86

Thanks!

rhinoduck:
Agh, silly me. Copied the wrong names. Pushed a fixup commit, that should do it.

rhinoduck:
Thanks for the feedback, I'll be going to bed now. I don't know what GitHub allows while rebasing in pull requests, maybe you can squash the fixup commit there. If not, but it allows you to force-push into this branch, you can do the squashing locally as `git rebase -i --autosquash HEAD~3` (an editor will open; just save and close) and then force-pushing it. If neither is possible, I'll do it when I wake up; if everything is alright.

trav90:
Perfect! It now compiles without any warnings at all. Thanks!

rhinoduck:
Squashed the fixup commit, and I think we're ready here. @wolfbeast

wolfbeast:
Thanks for chasing these down :)

PR #899, opened 2017-02-16
Resolve confusion between Flash and VLC plugins

JustOff:
Plugin permissions are confused between Flash and the VLC plugins. Fix this by checking that the plugin name for the special Flash plugin is "Shockwave Flash", which is true of Adobe Flash but not replacement plugins such as VLC or even gnash. This causes other plugins to have their own pref tag so that users can enable/disable them separately.

https://bugzilla.mozilla.org/show_bug.cgi?id=1293062

### DIFF
--- a/dom/plugins/base/nsPluginTags.cpp
+++ b/dom/plugins/base/nsPluginTags.cpp
@@ -201,7 +201,11 @@ void nsPluginTag::InitMime(const char* const* aMimeTypes,
         mIsJavaPlugin = true;
         break;
       case nsPluginHost::eSpecialType_Flash:
-        mIsFlashPlugin = true;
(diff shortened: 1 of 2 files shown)


JustOff:
Could you please consider to uplift this? 
It's so uncomfortable not to have the ability to disable VLC plugin, but leave Flash enabled.

wolfbeast:
How critical is this? Because I don't really see this meet the criteria for a point release uplift...

JustOff:
It's fairly important to me not to have the extra unused plugins enabled, but I can well live with the custom build, so you decide.

wolfbeast:
Very low risk patch, so might as well roll it in.

wolfbeast:
Uplifted in [a359367]

JustOff:
Thank you!

PR #901, opened 2017-02-16
Restyle extension target indicators

wsdfhjxc:
The ASCII bullet symbol has been replaced with a Unicode circle character. It's larger than the bullet by default, which makes it more appropriate as a glyph without need for resizing anything. This fixes ugly alignment on Linux and should make it consistent regardless of applied theme.

preview

### DIFF
--- a/toolkit/mozapps/extensions/content/extensions.css
+++ b/toolkit/mozapps/extensions/content/extensions.css
@@ -268,9 +268,8 @@ richlistitem:not([selected]) * {
 
 /* Indicator style for extension target application */
 .addon[native] .nativeIndicator {
-  font-size: 150%;
-  margin-top: -5pt;
-  margin-bottom: -1pt;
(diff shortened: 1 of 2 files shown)


JustOff:
I like the idea, but aren't they too small? In my browser this looks as:

glyph

wsdfhjxc:
Initially the dots supposed to be not too big and not too small for the default theme. I'm not sure what theme do you use, but they seem way oversized in your screenshot. With these changes the dots should follow font size used for extension title. See [here](https://github.com/MoonchildProductions/Pale-Moon/pull/743/files/93df0003fb937dcf43e9a520a89985c0b1e72205#r90946840). If you have another idea how to fix ugly alignment and properly deal with glyph size, please share it.

wolfbeast:
Simple and efficient. thanks.

Issue #760, opened 2016-12-08
Clarify UI for off-line data storage (and enable by default)


wolfbeast:
The current check-box UI in Advanced -> Network for asking the user to allow off-line storage is in conflict with the "always allow" preference that bypasses this selection (see #739).

What we need to do here is make this from a check-box into a 3-state choice:
1. *Allow by default* (similar to default 27.0.0 release, never prompts except for over-quota)
2. *Ask me every time* (similar to default 26.* and prior)
3. *Don't allow by default* (similar to default 26.* and prior with the prompt disabled)

The new **default** should be to **always allow** (because it's the sane choice for 2016/2017 and Web 2.0/3.0).

For the different states:
1. `offline-apps.allow_by_default` = `true` and `browser.offline-apps.notify` = `true`
1. `offline-apps.allow_by_default` = `false` and `browser.offline-apps.notify` = `true`
1. `offline-apps.allow_by_default` = `false` and `browser.offline-apps.notify` = `false`

These preferences should both be synced (only notify is synced currently).

THEtomaso:
More than those other options, I'd much rather like to see an option to always DENY offline data, without asking!
This option was missing in v26.* and prior as well.

wolfbeast:
No, it wasn't missing, just not explicitly stated. If you disabled the "ask" option (checkbox) then it would by default deny. The v26 option is "prompt to allow" or "deny", with a list of sites where you've chosen "never allow" and they would silently block.

Denying off-line data by default **will** break web applications. just be aware of that.

wolfbeast:
@JustOff Please note the string changes for 27.2 (not 27.1)

JustOff:
Yes, thanks. And it's good that you add new names for new strings so we can keep old ones for the backward compatibility.

PR #903, opened 2017-02-18
[suggestion] Customizable toolbars - persist the attribute "collapsed"

janekptacijarabaci:
Ad https://forum.palemoon.org/viewtopic.php?f=3&p=106977

There are more options, but this is maybe the easiest...

See also https://hg.mozilla.org/comm-central/rev/d4ff0342ca4e#l10.120

---

### DIFF
--- a/browser/base/content/browser.js
+++ b/browser/base/content/browser.js
@@ -4359,6 +4359,37 @@ function setToolbarVisibility(toolbar, isVisible) {
   toolbar.setAttribute(hidingAttribute, !isVisible);
   document.persist(toolbar.id, hidingAttribute);
 
+  // Customizable toolbars - persist the hiding attribute.
+  if (toolbar.hasAttribute("customindex")) {
+    var toolbox = toolbar.parentNode;
+    var name = toolbar.getAttribute("toolbarname").replace(" ", "_");
+    var toolbarInfoSep1 = ":";
+    var toolbarInfoSep2 = "-";
+    if (toolbox.toolbarset) {
+      try {
+        // Checking all attributes starting with "toolbar".
+        Array.prototype.slice.call(toolbox.toolbarset.attributes, 0)
+            .find(x => {
+              if (x.name.startsWith("toolbar")) {
+                var toolbarInfo = x.value;
+                var infoSplit = toolbarInfo.split(toolbarInfoSep1);
+                if (infoSplit[0] == name) {
+                  infoSplit[1] = [
+                    infoSplit[1].split(toolbarInfoSep2, 1), !isVisible
+                  ].join(toolbarInfoSep2);
+                  toolbox.toolbarset.setAttribute(
+                      x.name, infoSplit.join(toolbarInfoSep1));
+                  document.persist(toolbox.toolbarset.id, x.name);
+                }
+              }
+            });
+      } catch (e) {
+        Components.utils.reportError(
+            "Customizable toolbars - persist the hiding attribute: " + e);
+      }
+    }
+  }
+
   PlacesToolbarHelper.init();
   BookmarkingUI.onToolbarVisibilityChange();
   gBrowser.updateWindowResizers();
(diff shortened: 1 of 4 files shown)


wolfbeast:
Can you adds a few code comments to explain what you are doing, for future ease of maintenance? You're obviously purposefully making this extensible to other attributes by using an attribute array, so that should be at least somewhat documented for doing this for other attributes as well in the future.

janekptacijarabaci:
Ad illegal characters...

I suggest you add it to:
https://github.com/MoonchildProductions/Pale-Moon/issues/760#issuecomment-274151995

toolkit/locales/en-US/chrome/global/customizeToolbar.properties
```diff
@@ -3,8 +3,9 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 enterToolbarTitle=New Toolbar
-enterToolbarName=Enter a name for this toolbar:
+enterToolbarName=Enter a name for this toolbar (must not contain the ":" character):
```
wolfbeast: This won't be needed. People will be informed of their error when they make it; I'd rather keep prompts short.

browser/base/content/browser.js
```diff
@@ -4363,6 +4363,7 @@ function setToolbarVisibility(toolbar, isVisible) {
   if (toolbar.hasAttribute("customindex")) {
     var toolbox = toolbar.parentNode;
     var name = toolbar.getAttribute("toolbarname").replace(" ", "_");
+    // "toolbarInfoSepX" - it is used in multiple places of a code.
```
wolfbeast: What exactly do you mean with this? It's totally unclear what you're trying to say with this comment.

toolkit/content/customizeToolbar.js
```diff
@@ -178,6 +178,7 @@ function persistCurrentSets()
         // Attributes:
         // Names: "toolbarX" (X - the number of the toolbar)
         // Values: "Name:HidingAttributeName-HidingAttributeValue:CurrentSet"
+        // "toolbarInfoSepX" - it is used in multiple places of a code.
```
wolfbeast: ditto

toolkit/content/customizeToolbar.js
```diff
@@ -497,7 +498,9 @@ function addNewToolbar()
       continue;
     }
 
-    if (name.value.includes(":")) {
+    // "toolbarInfoSepX" - it is used in multiple places of a code.
```
wolfbeast: ditto

toolkit/content/widgets/toolbar.xml
```diff
@@ -57,6 +57,7 @@
           if (this.toolbarset) {
             // Create each toolbar described by the toolbarset.
             var index = 0;
+            // "toolbarInfoSepX" - it is used in multiple places of a code.
```
wolfbeast: ditto

janekptacijarabaci: If this character was changed. I can give it away.

wolfbeast: I still don't understand.

janekptacijarabaci: The separator character can be changed. If anyone would like to do it => it has to change everywhere, otherwise it will not work.

wolfbeast: OK... two things then.
1. You may want to use a less commonly used character instead of `:`, like the pipe `|` character.
2. If this is going to be configurable or needs to likely be changed in the future, then you should probably use a `const` value to define it once, and then use it throughout the files that way, or use a `define` and leverage the preprocessor.

wolfbeast:
Great! Thank you :)

janekptacijarabaci:
@JustOff 
FYI:
https://github.com/MoonchildProductions/Pale-Moon/pull/903/files#diff-97230e99d5d211b65ed801692da1e314R8
I suggest you add it to:
https://github.com/MoonchildProductions/Pale-Moon/issues/760#issuecomment-274151995
(add it all at once)

Thank you.

PR #904, opened 2017-02-18
Add an option to highlight all in findbar enabled by default

### LINKED ISSUE #364: Add an option to have "highlight all" for Find enabled by default.
Ref: https://forum.palemoon.org/viewtopic.php?f=13&t=11035

This should be relatively simple to do: when opening the find bar, check a pref and check "find all" based on it.

wsdfhjxc:
This is an attempt to resolve #364. Seems to be working fine.

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -753,6 +753,7 @@ pref("accessibility.typeaheadfind.autostart", true);
 //     1 - "always" (case-sensitive)
 // other - "auto"   (case-sensitive for mixed-case input, insensitive otherwise)
 pref("accessibility.typeaheadfind.casesensitive", 0);
+pref("accessibility.typeaheadfind.highlightallbydefault", true);
 pref("accessibility.typeaheadfind.linksonly", true);
 pref("accessibility.typeaheadfind.startlinksonly", false);
 pref("accessibility.typeaheadfind.timeout", 4000);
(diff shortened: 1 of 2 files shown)


wolfbeast:
LGTM! Thanks.

JustOff:
Such behaviour of highlight option looks very strange for me because it does not keeps state neither across restarts, nor even on close/open of the search bar. Besides, if you switch between tabs visual state of 'highlight all' button looses the correspondence with the option. It seems the old bug, but it became more noticeable when 'highlight all' is enabled by default. And last but not the least, I'm sure that accessibility.typeaheadfind.highlightallbydefault should be disabled by default to keep existing experience of browser usage.

wolfbeast:
>  I'm sure that accessibility.typeaheadfind.highlightallbydefault should be disabled by default to keep existing experience of browser usage.

Ah yes I overlooked that. It has to be false by default.

Also, if the option in the bar doesn't properly reflect the state of the feature, then that is introducing a bug.
I'll back it out from master; looks like it'll need some more work.

wolfbeast:
Backed out: fa21ff04b060efebbb0c4b8c21355cff80def55e & c15446e07a89fd19fdf0585c78a507f7a4e89f3b

wsdfhjxc:
Apparently I misunderstood the initial request. How should this work? The state of highlight all option in findbar should be based on the pref only for the very first search and then should be remembered per browser window accordingly to user's selection? Originally it's not remembered at all when reopening the findbar and with this PR it's always set based on the pref value. Remembering it accordingly to user's selection is easy to do, but I can't figure out what is causing this button to be unchecked after switching to another tab and that is ruining the whole purpose. Maybe someone smarter will take a look into this sometime.

wsdfhjxc:
Ok, I think I've a better solution which should cover most use cases for this kind of enhancement. I'll do some more testing and come back if it will be really working as intended.

PR #909, opened 2017-02-22
Fix potential crash & video stall in MFR

### LINKED ISSUE #908: Issue with media.mediasource.format-reader.mp4 enabled on Vimeo
People have reported on https://forum.palemoon.org/viewtopic.php?f=29&t=14887 that Vimeo videos fail to play when MSE is activated. Also MSE-MP4.

For example that video here: https://vimeo.com/203451175?ref=em-share

Tested a bit with the about:config entries responsible for Mediasource. If the media.mediasource.format-reader.mp4 entry is activated it makes Vimeo videos fail, but deactivated - while not totally shutting down MSE the videos are playing.

So something seems to be not working as expected in combination with media.mediasource.format-reader.mp4

trav90:
This PR prevents MSE streams from stalling (i.e. not playing, buffering forever, or never loading at all), and also fixes a potential crash point in MediaFormatReader.

Works as intended, and appears to fix #908 during my testing (will push out a new unstable build once this is merged for the people who reported the problem to confirm).

### DIFF
--- a/dom/media/MediaFormatReader.cpp
+++ b/dom/media/MediaFormatReader.cpp
@@ -924,6 +924,10 @@ MediaFormatReader::WaitForData(MediaData::Type aType)
   TrackType trackType = aType == MediaData::VIDEO_DATA ?
     TrackType::kVideoTrack : TrackType::kAudioTrack;
   auto& decoder = GetDecoderData(trackType);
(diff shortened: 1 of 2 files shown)


wolfbeast:
I'll kick off an unstable build and push it before I call it a night.

PR #918, opened 2017-02-24
Strict icon size for PMkit 'ui/button'

JustOff:
Based on [Bug #882910](https://bugzilla.mozilla.org/show_bug.cgi?id=882910) & [Bug #967115](https://bugzilla.mozilla.org/show_bug.cgi?id=967115).

### DIFF
--- a/browser/base/content/browser.css
+++ b/browser/base/content/browser.css
@@ -679,3 +679,8 @@ html|*#gcli-output-frame,
 toolbarbutton[type="badged"] {
   -moz-binding: url("chrome://browser/content/urlbarBindings.xml#toolbarbutton-badged");
 }
+
+/* Strict icon size for PMkit 'ui/button' */
+toolbarbutton[pmkit-button="true"] > .toolbarbutton-badge-container > .toolbarbutton-icon {
+  width: 16px;
+}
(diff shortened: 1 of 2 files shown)


browser/base/content/browser.css
```diff
@@ -679,3 +679,8 @@ html|*#gcli-output-frame,
 toolbarbutton[type="badged"] {
   -moz-binding: url("chrome://browser/content/urlbarBindings.xml#toolbarbutton-badged");
 }
+
+/* Strict icon size for PMkit 'ui/button' */
+toolbarbutton[pmkit-button="true"] > .toolbarbutton-badge-container > .toolbarbutton-icon {
+  width: 16px;
```
wolfbeast: Wouldn't you also want to restrict the height to 16px?

JustOff: Actually these icons are always square, so setting one of dimensions works fine. I'm not sure about performance, do you think it's better to specify the both explicitly?

wolfbeast: They are always -supposed to be- square but you never know with devs :p 
There won't be a performance impact, so I do think it's better to explicitly state width and height, just in case there's some odd-ball extension that uses asymmetrical icons.

JustOff: Ok!

PR #922, opened 2017-02-26
[suggestion] Page Info: The cipher name -> The cipher suite

janekptacijarabaci:
I propose to add this information...

Before:
_before

After:
_after

See:
https://github.com/MoonchildProductions/Tycho/commit/b47b094d9811e49b23d4bd7b0a48f266d511f925

Ugly... The mainly, important information :-)

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=886752

\+ style clean up

---

### DIFF
--- a/browser/base/content/pageinfo/security.js
+++ b/browser/base/content/pageinfo/security.js
@@ -55,7 +55,6 @@ var security = {
         cAName : issuerName,
         encryptionAlgorithm : undefined,
         encryptionStrength : undefined,
-        encryptionSuite : undefined,
         version: undefined,
         isBroken : isBroken,
         isMixed : isMixed,
@@ -66,9 +65,8 @@ var security = {
 
       var version;
       try {
-        retval.encryptionAlgorithm = status.cipherName; //FIXME: contains suite.
+        retval.encryptionAlgorithm = status.cipherSuite;
         retval.encryptionStrength = status.secretKeyLength;
-        retval.encryptionSuite = status.cipherName; //status.cipherSuite
         version = status.protocolVersion;
       }
       catch (e) {
@@ -96,7 +94,6 @@ var security = {
         cAName : "",
         encryptionAlgorithm : "",
         encryptionStrength : 0,
-        encryptionSuite : "",
         version: "",
         isBroken : isBroken,
         isMixed : isMixed,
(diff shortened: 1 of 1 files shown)


wolfbeast:
Absolutely not.
Dumping the raw cipher suite there is a bug. The suite name is available (added specifically) in our API for extensions to use if they want to analyze/display it, but it does not belong in the page info window. What belongs there is the name, and the name should be the cipher name.

This was a mistake made by a Firefox dev, but crypto freaks insisted it was kept. It's ugly as all hell and I won't have it. I extended the API for a reason.

janekptacijarabaci:
> It's ugly as all hell

OK. But everyone has a different opinion what is ugly :-)

wolfbeast:
Oh, come on. It's an internal identifier string - putting that in a common language dialog box is ugly. That's not just a subjective opinion either; it's (objectively) really bad style. :-)

Now, with the fix-up in place and the suite available to scripting, it is of course a possibility to parse that string and lift out more generally-known details to put into the page info dialog, like whether it has forward secrecy or not, but it has to be in a format that doesn't clash with the language already there.

janekptacijarabaci:
Is it still ugly (my latest suggestion - meant in good - `tooltiptext`)?


I will not parse it.

janekptacijarabaci:
@wolfbeast 
Can you please answer me to the question I posted 2 month ago?

wolfbeast:
What? Um.. you deleted your work branch and stating "I will not parse it"... I stopped paying attention to this after that.

Yes it's still ugly. It doesn't matter how you present the raw string.

janekptacijarabaci:
OK. No information is better then ugly information (which is not visible).

wolfbeast:
I don't know what you are trying to say, sorry. You ask if it's still ugly, and yes it is.
If the suite details are going to be presented, then they have to be in a readable/parsed format. the raw string is "cool for nerds" but won't help anyone else. If you want to improve this, then write a simple parser that pulls the most pertinent information out and displays it in plain language.

janekptacijarabaci:
> a simple parser

I'd think it wouldn't be forward compatible like that - see also (e.g.):
https://github.com/timw/groktls/blob/master/src/main/java/org/archie/groktls/impl/cipher/CipherSuiteParserImpl.java
That's the main reason.

> cool for nerds

This is how NSS presents the "cipher suite". I don't know why this information doesn't have a better format.

mattatobin:
I question showing the cypher string at all.. No end user cares what it is except maybe if it is Ok and what bits. Only crypto-freaks care if it is AES or Cameillia or what that even means.

Crypto-freaks can simply create and submit or use an extension to expand on that.. They would likely want it in the toolbar or statusbar anyway..

wolfbeast:
Already exists: cipherfox is Pale Moon compatible and can show the suite.

PR #923, opened 2017-02-27
Don't build NSS using compiler warnings as errors by default

trav90:
The recent NSS update caused build bustage on Linux by forcing the compiler to treat errors as warnings (which it does not do on the rest of the tree unless explicitly defined). This PR fixes that.


trav90:
A build with all recent NSS/NSPR related commits backed out compiles and packages successfully.

wolfbeast:
What was the build bustage with warnings as errors enabled?

trav90:
I don't have the exact output (I'm not home with my build machine at the moment) but it was a -Wmaybe-unitialized warning here (specifically it pointed to the asterik): https://github.com/MoonchildProductions/Pale-Moon/blob/master/security/nss/lib/libpkix/pkix_pl_nss/module/pkix_pl_ldapdefaultclient.c#L371

(only one warning was thrown, as the compiler stopped at that point).

PR #929, opened 2017-03-01
Remove use of decoder's monitor in MFR::ResetDecode()

### LINKED ISSUE #920: Hangs on Facebook (and possibly others)
From what I can gather from the hang reports submitted by different people, we're still seeing hangs in the mediadecoder, specifically `MediaDecoder::NotifyOwnerActivityChanged()`, where it will hang on a re-entrant guard `ReentrantMonitorAutoEnter mon(GetReentrantMonitor());`

The fact that this guard calls into nss threw me off but I've traced back the first usable xul.dll address to this particular function.

Perhaps this is running on a different thread than the main thread, which would likely assert in a debug build, but in a release build just hangs.
Because it is a hang, the call stack doesn't help much but I'll see if I can dig a bit deeper.

trav90:
This prevents a potential deadlock and appears to improve the hangs reported in #920.

Has been lightly tested by Tobin on both Linux and Windows, but probably should have some more testing done by those who have experienced the hang before pushing out a point release.

### DIFF
--- a/dom/media/MediaFormatReader.cpp
+++ b/dom/media/MediaFormatReader.cpp
@@ -938,8 +938,6 @@ MediaFormatReader::ResetDecode()
 {
   MOZ_ASSERT(OnTaskQueue());
 
-  ReentrantMonitorAutoEnter mon(mDecoder->GetReentrantMonitor());
-
   mAudioSeekRequest.DisconnectIfExists();
   mVideoSeekRequest.DisconnectIfExists();
   mSeekPromise.RejectIfExists(NS_OK, __func__);
--- a/dom/media/MediaFormatReader.cpp
+++ b/dom/media/MediaFormatReader.cpp
@@ -1240,6 +1240,15 @@ MediaFormatReader::GetBuffered()
     }
     intervals = mCachedTimeRanges;
   } else {
+    if (OnTaskQueue()) {
+      // Ensure we have up to date buffered time range.
+      if (HasVideo()) {
+        UpdateReceivedNewData(TrackType::kVideoTrack);
+      }
+      if (HasAudio()) {
+        UpdateReceivedNewData(TrackType::kAudioTrack);
+      }
+    }
     if (HasVideo()) {
       MonitorAutoLock lock(mVideo.mMonitor);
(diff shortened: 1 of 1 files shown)


trav90:
Added a new commit. This should be ready to ship to unstable. Hopefully the people who have reported hangs will test and report back.

wolfbeast:
I've just pushed Tobin's binaries with this PR out to publication.
Hopefully others will be able to verify this fixes the deadlocks. Possible that 1 or 2 still run into something as is usually the case with intermittent things that may look caused by the same thing but actually aren't, but fingers crossed for this workaround to be effective.

It's been bad design to pull all this off of the main thread, I think. But I know why Mozilla wanted to: e10s, and minimizing the actual tasks performed by the master process (because it'll be swamped with IPC message processing).
Ultimately, I'd prefer to put all this decoder allocation/destruction back on the main thread so you don't have to deal with possible re-entrancy, but that's going to be a massive task.

trav90:
> Ultimately, I'd prefer to put all this decoder allocation/destruction back on the main thread so you don't have to deal with possible re-entrancy, but that's going to be a massive task.

Agreed. Mozilla did get this all media thread mess sorted out and removed the troublesome ReentrantMonitorAutoEnter by ESR45, but it has numerous dependencies and as you say will require a lot of work. In the meantime, hopefully this will at least resolve the hanging issue until we can fix the root problem.

> with this PR out to publication

I'll push Linux builds out shortly.

wolfbeast:
Uplifted in 25f4b4989f56c0a37830aabaf320fde8f18608a7

wolfbeast:
@trav90 av I uplifted this as requested but it causes build bustage due to the OnTaskQueue() identifier.

I'm assuming that commit is non-critical for the point release hang fix?

trav90:
Hmm, odd that MSVC trips up on that but GCC doesn't. But regardless, that commit is non-critical for the point release. b52c5ab is the critical commit here.

trav90:
> odd that MSVC trips up on that but GCC doesn't

Doh, the Relbranch doesn't include PR #887. My bad!

Issue #932, opened 2017-03-03
Large pictures are not loaded and displayed


janekptacijarabaci:
Large pictures are not loaded and displayed (blank page or page with error text).

__Steps to reproduce:__

_Go to:_

http://imgsrc.hubblesite.org/hu/db/2003/28/images/a/formats/full_jpg.jpg
\- OK
(Pale Moon 27.1.1 - x32, x64)

http://imgsrc.hubblesite.org/hu/db/2006/10/images/a/formats/full_jpg.jpg
\- OK
(Pale Moon 27.1.1 - x64)
\- The image [...] cannot be displayed, because it contains errors.
(Pale Moon 27.1.1 - x32)

http://imgsrc.hubblesite.org/hu/db/2006/01/images/a/formats/full_jpg.jpg
http://www.vidiani.com/maps/maps_of_the_world/large_scale_detailed_satellite_map_of_the_World.jpg
\- The image [...] cannot be displayed, because it contains errors.
(Pale Moon 27.1.1, Firefox Nightly 54.0a1 (2017-03-02) (64-bit))

There are more images to test at http://hubblesite.org/ and http://hiroc.lpl.arizona.edu/images/PSP/

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=370763

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=1283961

See also (downscale all important image formats during decode):
https://bugzilla.mozilla.org/show_bug.cgi?id=1045926

mattatobin:
Stunning images but god even chrome is having some difficulty coping.. Obviously coping better than we are but jesus.

Atavic:
Both K-Meleon and Pale Moon load the stunning galaxy image here.

wolfbeast:
We are a web browser, not an image viewer - such resolutions are not really within scope and if too large (based on available resources or settings), Pale Moon should refuse to display them instead of gobbling up RAM (and crashing if OOM).

mattatobin:
I had a thought that was the case.

wolfbeast:
I don't particularly see a reason to have the browser display images > 16k pixels wide/high (which is about our current limit), so it'll be a wontfix.

Atavic:
Such images should be available from ftp, to be d/l and opened locally.

THEtomaso:
> http://imgsrc.hubblesite.org/hu/db/2006/10/images/a/formats/full_jpg.jpg

[...]

PR #933, opened 2017-03-03
Using "View Image" and "Save Image As" on a large canvas doesn't work

janekptacijarabaci:
Using "View Image" and "Save Image As" on a large canvas doesn't work

__Steps to reproduce:__

Go to (e.g. `canvas.html`):
```
<canvas width="11472" height="6429"></canvas>
<img src="http://imgsrc.hubblesite.org/hu/db/2003/28/images/a/formats/full_jpg.jpg" style="display:none"/>
<script>
var i = document.querySelector("img"),
    c = document.querySelector("canvas"),
    ctx = c.getContext("2d");
i.onload = function () {
  ctx.drawImage(i, 0,0, c.width, c.height);
}
</script>

```

Pale Moon 27.1.1 (27.2.0a1):

__I warn you:__
In the pop-up menu, select "View Image"
- It takes a long time, consume large amounts of memory, and __the browser crashes__
(dataURL vs. blobURL)

In the pop-up menu, select "Save Image As":
- It takes a long time, consume large amounts of memory

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1200524

Other large images see also:
#932 

---

I've created the new build (x32/x64, Windows) and tested.

### DIFF
--- a/browser/base/content/nsContextMenu.js
+++ b/browser/base/content/nsContextMenu.js
@@ -4,6 +4,8 @@
 
 Components.utils.import("resource://gre/modules/PrivateBrowsingUtils.jsm");
 
+Components.utils.importGlobalProperties(["URL"]);
+
 function nsContextMenu(aXulMenu, aIsShift) {
   this.shouldDisplay = true;
   this.initMenu(aXulMenu, aIsShift);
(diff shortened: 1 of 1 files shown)


wolfbeast:
Such large canvases are not very typical ;) -- but using blobs instead of base64 DataURLs is certainly more sane of an approach to pass the data around.

PR #930, opened 2017-03-02
Implement AbstractThread

trav90:
AbstractThread is a simplification of our existing MediaPromise code that (unlike MediaPromises) guarantees that two events will not run at the same time - either because it's a single thread or because it's a task queue.

Built and lightly tested on several YouTube vids and Twitch streams and appears to be working as intended.

PR #935, opened 2017-03-04
Improve ergonomics and efficiency of AbstractThread

trav90:
Followup to #930.

This PR removes wapper-style AbstractThreads and allows callers to pass an AbstractThread directly to promise APIs.

Been streaming Twitch and YouTube for the past several hours and appears to be working as intended.

### DIFF
--- a/dom/media/AbstractThread.cpp
+++ b/dom/media/AbstractThread.cpp
@@ -6,7 +6,6 @@
 
 #include "AbstractThread.h"
 
-#include "MediaTaskQueue.h"
 #include "nsThreadUtils.h"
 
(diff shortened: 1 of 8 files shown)


mattatobin:
Travis, issues seem to be able to happen most when things sit on an inactive tab either in a not has been played state or in a state where paused. Be sure to test this as well in addition to active playing/streaming.

trav90:
Yes I did. I switched between YT & Twitch after various amounts of time, leaving one in a background tab while the other played, then going vice versa, and leaving both in background tabs for extended periods while doing other things, etc.

> but indications of the possibility is concerning.

Agreed, and I am looking into this area as well.

PR #938, opened 2017-03-05
Fix deadlocks in SharedDecoderManager

trav90:
This PR should hopefully fix the hang in #920 for good (one of the commits here landed at Mozilla to fix a hang with the same stack trace). The remaining commits here are followups to that, ensuring that no other deadlocks happen in the surrounding code.

According to Mozilla, this hang was triggered only by some video formats and also varies based on the decoder being used (whether it be WMF, certain versions of FFmpeg, etc.). This would explain why the hang was difficult (or impossible) to reproduce on some systems.

Been testing on Twitch and YouTube the past couple hours and appears to work as intended.

### DIFF
--- a/dom/html/HTMLMediaElement.cpp
+++ b/dom/html/HTMLMediaElement.cpp
@@ -3857,7 +3857,7 @@ void HTMLMediaElement::UpdateMediaSize(const nsIntSize& aSize)
 
 void HTMLMediaElement::UpdateInitialMediaSize(const nsIntSize& aSize)
 {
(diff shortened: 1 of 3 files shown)


mattatobin:
I am building this for Windows. Will let you know.

wolfbeast:
Waiting for preliminary feedback from @mattatobin before merging.

mattatobin:
I like it.

wolfbeast:
Merging, then!

PR #942, opened 2017-03-08
Rework stream clock calculation in MediaDecoderStateMachine

trav90:
This PR simplifies and reworks the stream clock calculation in the MDSM, while also removing now unused code. This is a small step towards eliminating the troublesome reentrant MediaDecoder monitor.

Been testing on YT & Twitch with no issues.


wolfbeast:
Definitely looks like it simplifies code flow there.

PR #945, opened 2017-03-08
Properly handle add-on name in case it localized

JustOff:
This resolves following exceptions when addons.palemoon.org provides no data:

1) If install.rdf contains '<em:localized>' blocks without '<em:name>' (it's rather common situation), then corresponding add-on displays 'null' instead of real extension name.
2) If install.rdf contains '<em:localized>' block for en-US without '<em:name>' (this is uncommon, but allowed), then corresponding add-on can't be installed at all.

See also https://github.com/Pale-Moon-Addons-Team/phoebus/issues/42.


mattatobin:
See: https://github.com/Pale-Moon-Addons-Team/phoebus/issues/17

PR #948, opened 2017-03-09
Using "View Image" and "Save Image As" on a large canvas doesn't work (follow up)

janekptacijarabaci:
Follow up to #933 

We cannot use the chrome window. Every blob allocated memory which has not been released after closing the window/tab. And the blob still exist within its chrome window.

---

### DIFF
--- a/browser/base/content/nsContextMenu.js
+++ b/browser/base/content/nsContextMenu.js
@@ -1210,7 +1210,7 @@ nsContextMenu.prototype = {
       var win = doc.defaultView;
       if (!win) {
         Components.utils.reportError(
-            "View Image (on the <canvas> element):\n" +
+            "Save Image As (on the <canvas> element):\n" +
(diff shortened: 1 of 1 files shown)


browser/base/content/nsContextMenu.js
```diff
@@ -1196,18 +1203,26 @@ nsContextMenu.prototype = {
...
     if (this.onCanvas) {
       // Bypass cache, since it's a blob: URL.
       var target = this.target;
-      new Promise.resolve({then: function (resolve) {
-        target.toBlob((blob) => {
-          resolve(URL.createObjectURL(blob));
+      var win = doc.defaultView;
+      if (!win) {
+        Components.utils.reportError(
+            "View Image (on the <canvas> element):\n" +
```
wolfbeast: You probably mean "Save Image" here.

PR #950, opened 2017-03-10
Limit number of history entries saved by SessionStore

JustOff:
Limit number of back/forward button history entries saved by SessionStore for each tab.

Based on https://bugzilla.mozilla.org/show_bug.cgi?id=943339.


JustOff:
After additional testing I found that this patch breaks the saving of form data and scrolling position. I have almost finished with the correct solution, but to avoid patching the patch and keep changes clear I'd like to ask you to revert this one completely. I'm sorry about it.

wolfbeast:
No problem, just try to get it right in the next PR :)

PR #953, opened 2017-03-13
Implement a new media TailDispatcher

trav90:
From the TaskDispatcher header file:

> TaskDispatcher is a general abstract class that accepts tasks and dispatches them at some later point. These groups of tasks are per-target-thread, and contain separate queues for two kinds of tasks - "state change tasks" (which run first, and are intended to be used to update the value held by mirrors), and regular tasks, which are other arbitrary operations that the are gated to run after all the state changes have completed.

This will work together with our new media AbstractThread (#930) and will help to improve the raciness in some of our current media code. It is also another step towards eliminating the decoder monitor that's caused some issues.

In addition to implementing the TaskDispatcher, this PR also makes some improvements to AbstractThread, improves atomicy (will be needed for future patches), and adds a few more assertions to the MDSM.

There will be future improvements made to and heavier use of the TaskDispatcher, but this basic implementation should be good enough for now to merge in based on my testing yesterday and today (although I am adding the "Verification Needed" label because it is still a fairly substantive change).

### DIFF
--- a/dom/media/AbstractThread.cpp
+++ b/dom/media/AbstractThread.cpp
@@ -6,6 +6,7 @@
 
 #include "AbstractThread.h"
 
+#include "MediaTaskQueue.h"
 #include "nsThreadUtils.h"
 
 #include "mozilla/ClearOnShutdown.h"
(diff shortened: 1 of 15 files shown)


wolfbeast:
Verified on Twitch and YouTube HD on Windows, seems to be OK. Vimeo bucks but that seems to be JS issues rather than media. Some play fine, others throw in the console, but rather about JS than about the media.
image1
image3

PR #954, opened 2017-03-15
Clean up and improve AbstractThread dispatch semantics

trav90:
The main accomplishments of this PR is that the assertion of dispatch success is now automated and the API entry points on MediaTaskQueue have been streamlined. Additionally, some unused code has been removed and a few shutdown checks have been added to help prevent crashes.

Been testing on YouTube HD, Twitch, and a couple Vimeo vids and appears to be working as intended.

Issue #927, opened 2017-02-28
27.1.0 build broken on Mac build


stevenschlansker:
The 27.1.0 Mac build is currently failing:

```
27:45.05 In file included from tycho/default/ws/obj-x86_64-apple-darwin12.5.0/dom/media/platforms/Unified_cpp_dom_media_platforms0.cpp:29:

27:45.06 tycho/default/ws/dom/media/platforms/apple/AppleATDecoder.cpp:496:21:
error: no matching function for call to 'AudioFileStreamParseBytes'

27:45.07   OSStatus status = AudioFileStreamParseBytes(mStream,
```

Hopefully not a terribly hard issue to track down; I'll work on it over the next few days.  (can't assign it to myself)

wolfbeast:
Assigned you

trav90:
@stevenschlansker try building with this extra commit: ab938f4

stevenschlansker:
Thanks @trav90 that seems to help.  There's more build errors down the line though, so I will track those down and make sure the build succeeds before closing this one for sure.

trav90:
This should be fixed by #955

PR #955, opened 2017-03-15
fmp4 decoder: add missing #include for mac

stevenschlansker:
You might want to consider this a WIP -- while it helps the build get further
along, I still don't have a successful build.

manual cherry-pick of ab938f4e956e4f53c9cec092892e9f308fcafaf9 by @trav90

### DIFF
--- a/dom/media/fmp4/MP4Decoder.cpp
+++ b/dom/media/fmp4/MP4Decoder.cpp
@@ -23,7 +23,9 @@
 #include "nsIGfxInfo.h"
 #include "AndroidBridge.h"
 #endif
-
+#ifdef MOZ_APPLEMEDIA
+#include "AppleDecoderModule.h"
+#endif
 #ifdef MOZ_FFMPEG
 #include "FFmpegRuntimeLinker.h"
 #endif


wolfbeast:
It may be better to simply focus on the current intended release (27.2) that I've just made a branch for. Skip 27.1 and get all the needed updates in one go ;)

trav90:
As mentioned above, this patch doesn't fully fix building on Mac, but it does resolve the bustage reported in #927. I suggest going ahead and merging this in (with uplift to the 27.2 RelBranch would probably be good also).

This way, one of the (hopefully only) two issues with building for Mac will be resolved.

wolfbeast:
Is an uplift OK here? Where did this bust? I ask because the latest media additions from #954 were NOT merged into `27.2_RelBranch`.

trav90:
It should be. This was actually busted back in commit 60b158f, which didn't land until 27.1.0.

trav90:
See also commit b5d96ee, which fixes the exact same issue, but on Windows.

wolfbeast:
Right! uplifting.

wolfbeast:
Uplifted in d768214

stevenschlansker:
Thanks!

PR #958, opened 2017-03-16
Prettify rendering of JXR images with transparency

rhinoduck:
Prevent fully transparent areas from being black (or whatever the color
values there are) when rendering JXR images with planar alpha before all
data is available.


wolfbeast:
Nice enhancement there. Thanks!

PR #960, opened 2017-03-16
Fix Windows build bustage

trav90:
Should hopefully fix the bustage cause by #954. Unfortunately I do not have a way to verify this myself, so if someone (@JustOff?) could verify this works that'd be great.


JustOff:
Thanks, now all builds fine.

trav90:
Thanks for confirming!

intika:
i confirm too it works

PR #968, opened 2017-03-19
XMLHttpRequest: Expose mozAnon constructor js binding to system scopes

janekptacijarabaci:
## 1) __Depends on: 2)__

XMLHttpRequest.mozAnon is a boolean. If true, the request will be sent without cookie and authentication headers.

An example:
```
Components.utils.importGlobalProperties(["XMLHttpRequest"]);

var x = new XMLHttpRequest({mozAnon: true});
```
The problem is that this object property (x.mozAnon) has still the same value = false.

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1156101
https://bugzilla.mozilla.org/show_bug.cgi?id=1163898

## 2)

__!! This example crashes browser (Pale Moon 27.2.0) !!:__

__Steps to reproduce__

1) Open Scratchpad

3) Insert the code:
```
Components.utils.importGlobalProperties(["XMLHttpRequest"]);

var x = new XMLHttpRequest({mozAnon: true});
```
4) Run
5) Result: ...crashes browser.

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1137591
https://bugzilla.mozilla.org/show_bug.cgi?id=1141510

---

Regression (`Components.utils.importGlobalProperties(["URL"])`):
https://github.com/MoonchildProductions/Pale-Moon/pull/933/files#diff-78edc04f5fd7f14f6c33615a003f60f8R7
(but it is no longer needed - see (follow up): https://github.com/MoonchildProductions/Pale-Moon/pull/948)

See also (e.g.):
https://github.com/greasemonkey/greasemonkey/commit/535d996c3266b99158fa0cd2729027551d8b4ebf

---

__Please would you all try.__

---

You add the label `Extensions`, please.

---

### DIFF
--- a/browser/base/content/nsContextMenu.js
+++ b/browser/base/content/nsContextMenu.js
@@ -4,8 +4,6 @@
 
 Components.utils.import("resource://gre/modules/PrivateBrowsingUtils.jsm");
 
-Components.utils.importGlobalProperties(["URL"]);
-
 function nsContextMenu(aXulMenu, aIsShift) {
   this.shouldDisplay = true;
   this.initMenu(aXulMenu, aIsShift);
--- a/dom/base/nsGlobalWindow.cpp
+++ b/dom/base/nsGlobalWindow.cpp
@@ -1148,7 +1148,6 @@ nsGlobalWindow::nsGlobalWindow(nsGlobalWindow *aOuterWindow)
 
     mObserver = new nsGlobalWindowObserver(this);
     if (mObserver) {
-      NS_ADDREF(mObserver);
       nsCOMPtr<nsIObserverService> os = mozilla::services::GetObserverService();
       if (os) {
         // Watch for online/offline status changes so we can fire events. Use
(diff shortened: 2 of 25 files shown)


wolfbeast:
No crash experienced...

wolfbeast:
If you just ported those 4 bugs, then that should be fine; merging.

JustOff:
It crashes for me in browser & content environment (27.2.0 & 27.3.0a1).

wolfbeast:
Tried both. In content it throws (as it should because it won't have access to `Components.utils.importGlobalProperties` and in browser it just doesn't do anything visibly. No crashes either way.

JustOff:
As far as I understand, `Components.utils.importGlobalProperties` is also available in content mode when active tab contains a chrome page (for example about:addons), that was my case.

With this PR - no more crashes, but: `Exception: [Exception... "Component returned failure code: 0x80040111 (NS_ERROR_NOT_AVAILABLE) [nsIXPCComponents_Utils.importGlobalProperties]"  nsresult: "0x80040111 (NS_ERROR_NOT_AVAILABLE)"  location: "JS frame :: Scratchpad/1 :: <TOP_LEVEL> :: line 10"  data: no]`- so it should be?

janekptacijarabaci:
> so it should be?

Yes.

PR #970, opened 2017-03-20
Fixes to nsWEBPDecoder

rhinoduck:
Since I had spent some time learning the workings of ImageLib while adding JXR support, I was able to spot a few issues in the WebP decoder while experimenting with downscaling during decode. A cleanup commit, which you will hopefully consider an improvement, has been packaged with this also.

Everything should be pretty clear, the only doubts I have are about the "Simplify libwebp usage in nsWEBPDecoder" commit where I am not sure whether what I removed wasn't there as a test for ABI compatibility for cases when system/custom libwebp is loaded on unix-like systems. If that was so, let me know, and I will try to find a clearer way to do this, or I will at least add a comment to clarify this.

### DIFF
--- a/image/decoders/nsWEBPDecoder.cpp
+++ b/image/decoders/nsWEBPDecoder.cpp
@@ -36,24 +36,19 @@ nsWEBPDecoder::~nsWEBPDecoder()
          ("nsWEBPDecoder::~nsWEBPDecoder: Destroying WEBP decoder %p",
           this));
 
-  // Flush the Decoder and let it free the output image buffer.
+  // It is safe to pass nullptr to WebPIDelete().
   WebPIDelete(mDecoder);
-  WebPFreeDecBuffer(&mDecBuf);
 }
 
 
 void
 nsWEBPDecoder::InitInternal()
 {
-  if (!WebPInitDecBuffer(&mDecBuf)) {
-    PostDecoderError(NS_ERROR_FAILURE);
-    return;
-  }
+  mDecoder = WebPINewRGB(MODE_rgbA, nullptr, 0, 0);
 
-  mDecBuf.colorspace = MODE_rgbA;
-  mDecoder = WebPINewDecoder(&mDecBuf);
   if (!mDecoder) {
     PostDecoderError(NS_ERROR_FAILURE);
+    return;
   }
 }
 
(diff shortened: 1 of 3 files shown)


wolfbeast:
As far as your concern is involved for system-libwebp, if it turns out to be a requirement, I'm sure the *nix folks will let us know. As a critical note there: I'm no proponent of system libs on *nix for things we have in our tree. While I understand the principal thought behind using shared libs, and will happily use them on Windows systems, on *nix there is the problem that there are literally thousands of possible combinations of system/tree libs. We can only reasonably test against known code and linking against system libs that may have unknown configurations isn't something that can guarantee stability or proper operation. It's a grey area -- it may work well for simple tools that only use 1 or 2 libs, but not something as complex as this browser.

rhinoduck:
Sure, it is a compilation option for those who want it. If you are going to compile the browser with custom settings, you probably already know what you are doing. While distributing binaries, it is preferable for them to work out of the box, I won't dispute that :)

Anyway, my only concern in this case was what the intention of those particular pieces of code was; if it wasn't this (and at the same time it wasn't something else which possibly failed to cross my mind), then it should be OK to leave the commit as it is = those pieces of code will be removed because they are not needed.

wolfbeast:
> If you are going to compile the browser with custom settings, you probably already know what you are doing.

Unfortunately it's not that simple :) Plenty Linux versions rely on source building on end user's machines using package/build instructions supplied by someone else. So the end user running into problems won't be the one who "knows what he's doing" in that case.

As for the actual PR, I looked through it and seems good otherwise. I'm bit fuzzy on class inheritance and virtual vs. override, but LGTM.

rhinoduck:
This is one of the times when cpprefence.com is actually quite readable: [override specifier](http://en.cppreference.com/w/cpp/language/override)

> In a member function declaration or definition, override ensures that the function is virtual and is overriding a virtual function from the base class. The program is ill-formed (a compile-time error is generated) if this is not true. 

Basically, since C++11, you can use `virtual` the first time and then just use `override` every time you, well, override, and that way get additional compiler checks that you are actually overriding a virtual method which exists in the base class (direct or somewhere down the inheritance tree); and also easily see this when reading code.

---

Ugh, I called it the `override keyword` in the commit message which just got merged and embedded in history forever... now I'll never be able to become a language lawyer :(

wolfbeast:
Just performed a quick build and using the Google WebP gallery to test functionality; all good!

rhinoduck:
Glad to hear that! I try to test everything before I "push", but I am just one man with his one specific approach. The more eyes the thing gets seen by and the more directions it gets poked from, the better.

PR #972, opened 2017-03-21
Allow anchorObject.hash getters do unescape only if dom.url.getters_decode_hash is true

### LINKED ISSUE #971: URL encoding error (%2B / plus character)
There seems to be an issue with URL encoding.
The following URL (Compiler Explorer) link demonstrates the issue:
https://godbolt.org/g/dW4xbr
When opening with the current version of Pale Moon (27.2.0), the `+` is missing from `return x + y` as well as from `-std=c++1z`. The link works correctly with other browsers (tested with regular Firefox and Vivaldi).

It seems that Pale Moon does one extra step of encoding compared to the above. A minimal example to reproduce the problem: 
- Save the following as a local HTML document, say, `filename.html`.
- Open `filename.html#test%2B%2B%2Btest`.
- Should display: `#test%2B%2B%2Btest`.
- Displays instead: `#test+++test`.

```
<!DOCTYPE HTML>
<html>
<head>
<script>
alert(document.location.hash);
</script>
</head>
<body></body>
</html>
```

Thanks to @mattgodbolt for the test case!

JustOff:
Based on https://bugzilla.mozilla.org/show_bug.cgi?id=1093611.

Changes default behavior to not unescape anchorObject.hash.

Resolves #971.

### DIFF
--- a/dom/base/Link.cpp
+++ b/dom/base/Link.cpp
@@ -442,7 +442,7 @@ Link::GetHash(nsAString &_hash, ErrorResult& aError)
   nsresult rv = uri->GetRef(ref);
   if (NS_SUCCEEDED(rv) && !ref.IsEmpty()) {
     _hash.Assign(char16_t('#'));
-    if (nsContentUtils::EncodeDecodeURLHash()) {
+    if (nsContentUtils::GettersDecodeURLHash()) {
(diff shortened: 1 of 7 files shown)


wolfbeast:
Ah that was a lot simpler than I thought, considering we already had most of it in place. Thanks for going through the steps!

PR #973, opened 2017-03-22
Apply prefs-related prefs before applying other synced prefs

JustOff:
Based on https://bugzilla.mozilla.org/show_bug.cgi?id=753289.

This fixes [the following wrong behavior](https://discourse.mozilla-community.org/t/how-to-sync-preferences-of-a-bootstrapped-extension-via-sync/3024/6): if user create new profile and enable sync in the new profile before adding all extensions then the preferences from extensions that are not installed yet will be remove from the sync server, thus will set to default in all other profiles (devices).

Reported at https://forum.palemoon.org/viewtopic.php?t=15213&p=109958.

Need verification.

### DIFF
--- a/services/sync/modules/engines/prefs.js
+++ b/services/sync/modules/engines/prefs.js
@@ -8,7 +8,7 @@ const Cc = Components.classes;
 const Ci = Components.interfaces;
 const Cu = Components.utils;
 
-const WEAVE_SYNC_PREFS = "services.sync.prefs.sync.";
+const PREF_SYNC_PREFS_PREFIX = "services.sync.prefs.sync.";
 
 Cu.import("resource://services-sync/engines.js");
 Cu.import("resource://services-sync/record.js");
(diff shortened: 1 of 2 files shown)


services/sync/modules/engines/prefs.js
```diff
@@ -8,7 +8,7 @@ const Cc = Components.classes;
 const Ci = Components.interfaces;
 const Cu = Components.utils;
 
-const WEAVE_SYNC_PREFS = "services.sync.prefs.sync.";
+const PREF_SYNC_PREFS_PREFIX = "services.sync.prefs.sync.";
```
wolfbeast: That is one ugly rename!
Please make this `SYNC_PREFS_PREFIX` -- that's clear enough without having "PREF" in there twice.
Apply this change to all other occurrences.

JustOff: Done.

PR #974, opened 2017-03-22
Add documents to embed element capabilities

JustOff:
Based on https://bugzilla.mozilla.org/show_bug.cgi?id=1237963.

Resolves problem with embedded YouTube videos, also [allows to play internally supported media types using embed tag without external plugins](https://forum.palemoon.org/viewtopic.php?t=15165&p=109995).


wolfbeast:
Bad web coding wins again because Chrome allows it. :sigh:

JustOff:
Yeah, but it's funny that this patch has also such a positive side effect (working embed without plugins).

wolfbeast:
No, it's bad. `<embed>` isn't supposed to be used this way -- and the spec clearly says so as well. What we do here is allow more sloppy coding of webmasters who don't know what they are doing and don't think beyond "*shrug* it works in chrome so I must be doing it right".

Yeah I know, "just make it work already" -- but let me complain about it a little ;-)

PR #975, opened 2017-03-22
Customizable toolbars - backward compatible (but unidirectional)

janekptacijarabaci:
Ad:
https://forum.palemoon.org/viewtopic.php?f=3&t=15212
("The 27.2.0 update deleted a custom toolbar")

It's my fault (#903), I'm sorry :unamused:

The structure is different:
`Name:CurrentSet` (27.1.2-)
vs.
`Name|HidingAttributeName-HidingAttributeValue|CurrentSet` (27.2.0+)

The structure should be (I don't know if forever - but maybe... it would bother later):
`Name:CurrentSet:HidingAttributeName-HidingAttributeValue` (e.g. - or JSON instead of strings?!)

AFAIK: But it's too late now... (IMHO: apparently three would be too much)

---

This (+ style clean up) is one way compatibility
(27.0/27.1 => 27.2.0+, but not vice versa - but in that case... use a script?!)

I haven't better idea...

---

### DIFF
--- a/toolkit/content/customizeToolbar.js
+++ b/toolkit/content/customizeToolbar.js
@@ -179,7 +179,7 @@ function persistCurrentSets()
         // Persist custom toolbar info on the <toolbarset/>
         // Attributes:
         // Names: "toolbarX" (X - the number of the toolbar)
-        // Values: "Name:HidingAttributeName-HidingAttributeValue:CurrentSet"
+        // Values: "Name|HidingAttributeName-HidingAttributeValue|CurrentSet"
         gToolbox.toolbarset.setAttribute("toolbar"+(++customCount),
                                          toolbar.toolbarName
                                          + gToolbarInfoSeparators[0]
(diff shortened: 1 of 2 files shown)


wolfbeast:
Should this be uplifted, or doesn't it matter at this point?

janekptacijarabaci:
> Should this be uplifted

IMHO yes. Because not everybody updated. If it will be fixed soon: 27.0/27.1 => (e.g.) 27.2.1, and will not have a problem there.

But... You add the label `Verification needed`, please.

wolfbeast:
I don't really see an easy way to verify this by others "in the wild".

wolfbeast:
One-way migration to a later version is fine - it happens all the time.

janekptacijarabaci:
> I don't really see an easy way to verify this by others "in the wild".

IMHO: Portable/*.zip version... :-) But it was just an idea.

Issue #984, opened 2017-03-31
Referer not sent when submitting forms to the same host and scheme


KrzysiekJ:
## Steps to reproduce

1. Go to http://jurewicz.org.pl/en/blog/2016/02/06/per-file-cryptocurrency-payment-examples/
2. Click on one of the file links.
3. Click “Reset payments”.

## What should happen

`Referer` header should be sent. Payments should be reset.

## What happens.

`Referer` header is not sent. CSRF validation fails because of that and payments are not reset.

wolfbeast:
> Referer header should be sent.

According to whom?
https://tools.ietf.org/html/rfc7231#section-5.5.2 states it's not required:

    Some servers use the Referer header field as a means of
    denying links from other sites (so-called "deep linking") or
    restricting cross-site request forgery (CSRF), but not all requests
    contain it.

We are conservative with sending referrers because:

    The Referer field has the potential to reveal information about the
    request context or browsing history of the user, which is a privacy
    concern if the referring resource's identifier reveals personal
    information (such as an account name) or a resource that is supposed
    to be confidential (such as behind a firewall or internal to a
    secured service).

In addition, for a request on the same origin, sending a referrer makes very little sense. Also, relying on referrers is potentially dangerous because they can easily be spoofed.

wolfbeast:
Ah so a referrer policy request was sent? Essential information there.
Silly typo on Mozilla's side, looks like.

wolfbeast:
Solved by #987 fixing the typo.

KrzysiekJ:
This site has not been delivering any referrer policy, so the above typo fix would not probably fix the issue in this particular case. 

… The `referrer-policy: same-origin` header has been just added, but still no `Referer` is sent. Since the typo fix is about another value, I guess it will not solve the problem.

The HTTP specification states that:

> Because the source of a link might be private information or might reveal an otherwise private information source,it is strongly recommended that the user be able to select whether or not the Referer field is sent. For example, a browser client could have a toggle switch for browsing openly/anonymously, which would respectively enable/disable the sending of Referer and From information.

Indeed, Pale Moon has [many `about:config` settings regarding `Referer`](https://github.com/MoonchildProductions/Pale-Moon/blob/0381c6af924b366ae24e9339d55e57e3d2a9ef81/modules/libpref/init/all.js#L1509-L1517), but they seem to be ignored.

[...]

PR #987, opened 2017-03-31
The referrer policy specification: "origin-when-crossorigin" should actually be "origin-when-cross-origin" (fix typo)

janekptacijarabaci:
Ad #984 

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1163743

---


wolfbeast:
Confirmed that this resolves #984 -- thanks for spotting the typo!

PR #988, opened 2017-03-31
Align the check of [[avilableLocales]] with ucol_getAvailable() from ICU55+

### LINKED ISSUE #976: toLocaleDateString() not working properly after ICU58 update
It looks like some API changes were made to ICU, and we'll have to check what needs changing in `/js/src/builtin/Intl.cpp` to match the new API in ICU58+

Potentially relevant BMO bugs: 1075758 1210165 1299615 1323254

re: https://forum.palemoon.org/viewtopic.php?f=3&t=15227

JustOff:
Consider all locales implied from the default locale and the last-ditch locale (currently en-GB) to be always present in [[availableLocale]].

Based on [1175347](https://bugzilla.mozilla.org/show_bug.cgi?id=1175347) and [1272523](https://bugzilla.mozilla.org/show_bug.cgi?id=1272523).

Resolves #976.

### DIFF
--- a/js/src/builtin/Intl.js
+++ b/js/src/builtin/Intl.js
@@ -437,15 +437,99 @@ function CanonicalizeLanguageTag(locale) {
 }
 
 
-// mappings from some commonly used old-style language tags to current flavors
-// with script codes
+function localeContainsNoUnicodeExtensions(locale) {
+    // No "-u-", no possible Unicode extension.
+    if (callFunction(std_String_indexOf, locale, "-u-") === -1)
+        return true;
+
+    // "-u-" within privateuse also isn't one.
+    if (callFunction(std_String_indexOf, locale, "-u-") > callFunction(std_String_indexOf, locale, "-x-"))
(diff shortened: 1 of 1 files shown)


wolfbeast:
How exactly would I verify this?

JustOff:
Before this PR `Intl.Collator().resolvedOptions().locale` returns `en-GB`, after - actual system locale.

wolfbeast:
That works too ;) I didn't have the ICU reference handy.

wolfbeast:
Verified in 20170331180426

Behavior is using the locale setting from the O.S. and using default formats for that locale.
Any changes to the formats in the O.S. specifically diverging from the default for the overall locale aren't picked up, though, but that's to be expected.

JustOff:
That's wonders me too, why not to use browser locale for `Intl.NumberFormat` and `Intl.DateTimeFormat` ? Even Chrome does so, but not Mozilla ;-)

PR #989, opened 2017-03-31
[Minor fix] Invalid TYPE attribute makes UL look like OL

janekptacijarabaci:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=241719

An example: https://bugzilla.mozilla.org/attachment.cgi?id=147039

---


wolfbeast:
Verified fixed in 20170331180426

PR #990, opened 2017-03-31
[Minor fix] The first letter - should behave according to specification

janekptacijarabaci:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1260366

An example: http://codepen.io/worg/pen/wGqBLd

---


wolfbeast:
Verified fixed in 20170331180426

PR #991, opened 2017-03-31
[Minor fix] XMLHttpRequest - GetAllResponseHeaders() should return an empty string (instead null)

janekptacijarabaci:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1286744

An example: http://w3c-test.org/XMLHttpRequest/xmlhttprequest-network-error.htm

---


wolfbeast:
Verified PASS result on linked test in 20170331180426

Issue #996, opened 2017-04-02
After importing from HTML, a bookmark loses it's description text if a separator is in place next to it


Lorienna:
Hello,

After importing bookmarks from HTML, a bookmark loses it's description text if a separator is in place next to it (due to lack of enclosing </DD> tags in descriptions' HTML code); this affects only the 'bookmarks' themselves, not the 'folders'.
Apparently, some elements in the code are lacking enclosing tags by design, and it's unclear why.

This affects all Mozilla-based desktop browsers since at least version 3.6 (maybe even earlier).

<br>
<b>Steps to reproduce:</b>

1. Create a bookmark or choose an already existing one;

2. Fill the description field of this bookmark with text;

3. Create a separator after this bookmark;

4. Export the bookmarks into HTML;

5. Import the HTML file created in the step 4.

<b>Actual results:</b>
The bookmark loses it's description text if the separator is in place next to it.

<b>Expected results:</b>
The bookmark's desctiption text stays intact even if the separator is in place next to it.

<br>
Also, there's a <b>lack of clear code structure in the exported HTML files</b> in Pale Moon 27 (likewise in the other recent Mozilla-based desktop browsers), which makes it very difficult to check or edit manually, so please fix the regression.

PR #997, opened 2017-04-02
[Minor fix] Importing from HTML - A bookmark loses it's description text if a separator is in place next to it

janekptacijarabaci:
Ad #996 

---

### DIFF
--- a/toolkit/components/places/BookmarkHTMLUtils.jsm
+++ b/toolkit/components/places/BookmarkHTMLUtils.jsm
@@ -1173,6 +1173,6 @@ BookmarkExporter.prototype = {
     let descriptionAnno = aItem.annos &&
                           aItem.annos.find(anno => anno.name == DESCRIPTION_ANNO);
     if (descriptionAnno)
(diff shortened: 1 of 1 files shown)


JustOff:
Are you sure this actually fixes the problem? As far as I see from `BookmarkHTMLUtils.jsm` the `DD` tag in Netscape Bookmark Format does not implied to be explicitly closed and much more crazy logic is used.

janekptacijarabaci:
> Are you sure this actually fixes the problem?

It works for me. I am not sure if it breaks something else - but I don't know what.

JustOff:
As far as I understand the problem is with parser, not with writer. I think we can't just change Netscape Bookmark Format, even if it's ugly.

janekptacijarabaci:
This idea of not closing (some) tags is very strange (there may be more problems in the future).

But... I was looking for an interim solution. The simplest solution. I was not looking at it precisely.

wolfbeast:
If there's an issue with this, it should be solved on the reading end, not the writing end, anyway. I'll take the proposed solution in #999 in favor of this, since it'll fix it for reading from old and new exports, IIUC.

PR #1006, opened 2017-04-06
WMF fixes

trav90:
This PR should fix a couple bugs we inherited with Tycho. Since these changes only affect Windows, I unfortunately have no way to build or test, so verification will definitely be needed.

- Aligns the behavior of the WMF decoder with the decoders used on other platforms by not skipping audio samples that have a negative presentation time
- Ensures that we reliably get expected behavior when making calls after Shutdown() (our current behavior could potentially be inconsistent if the decoder enters a dormant state).

### DIFF
--- a/dom/media/platforms/wmf/WMFAudioMFTManager.cpp
+++ b/dom/media/platforms/wmf/WMFAudioMFTManager.cpp
@@ -251,7 +251,6 @@ WMFAudioMFTManager::Output(int64_t aStreamOffset,
   // reset the frame counters, and capture the timestamp. Future timestamps
   // will be offset from this block's timestamp.
   UINT32 discontinuity = false;
-  int32_t numFramesToStrip = 0;
   sample->GetUINT32(MFSampleExtension_Discontinuity, &discontinuity);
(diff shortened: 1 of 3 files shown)


wolfbeast:
This sounds promising for people reporting audio desync errors on Twitch. Was that the intent to tackle?

trav90:
While I don't think that this alone will fix the problem on Twitch, it shouldn't hurt. I've got some pretty big fixes coming down the pipe (including more audio fixes, some MSE improvements, and a pretty big change that will affect the media subsystem as a whole). This PR lays some groundwork for those changes by making the decoder behavior on Windows more consistent with Linux & Mac to hopefully reduce the chances of platform-specific regressions that I would be unable to troubleshoot alone.

Stay tuned; big things are coming!

wolfbeast:
Built on Windows, used to watch twitch live streams at 1080p60 and some vimeo HD nature sequences, played with quality settings during playback, skipped back and forth, no issues seen (but then again I never saw the desync either ;P )

Consider this verified.

trav90:
Thanks!

PR #1010, opened 2017-04-07
Update format of readme

kirkins:
I was just going to add the space betwe '#Pale' on line 1 so it format correctly but I figured might as well make links in a list format.

Browser was installed by default on Manjaro i3-wm edition and I'm liking it so far. Hopefully I can find something more to contribute.


wolfbeast:
Thanks!

PR #1011, opened 2017-04-09
Use nsRefPtr in nsSMILCompositor.cpp

trav90:
Fixes Linux build bustage. Followup to commit 3d7fd80.


wolfbeast:
Damnit. sorry about that. I keep overlooking that since MSVC accepts RefPtr.

PR #1012, opened 2017-04-09
Improve seeking behavior when audio and video buffered ranges differ

trav90:
Currently, when seeking to video we may seek to a different time in the audio stream in order to get good A/V sync. But in some cases this causes our media code to get stuck in waiting mode resulting in the player never actually starting playback. Some examples of this:

https://www.youtube.com/watch?t=61&v=ktbGjjzSKao
https://www.youtube.com/watch?t=61&v=5m5harNeUd8

This PR adjusts seeking in MSE and lets the MDSM handle dropping video frames if needed so we can keep good A/V sync.
Additionally, a few calls to RemoveDecoder have been added for if TrackBuffer initialization is aborted (prevents hangs in some cases).

Tested and now plays the links above, as well as other YouTube, Vimeo, and Twitch videos. While this PR doesn't seem to fully fix the A/V desync issue on Twitch, it does appear to improve some live streams there.

### DIFF
--- a/dom/media/MediaFormatReader.cpp
+++ b/dom/media/MediaFormatReader.cpp
@@ -1156,7 +1156,6 @@ MediaFormatReader::OnVideoSeekCompleted(media::TimeUnit aTime)
 
   if (HasAudio()) {
     MOZ_ASSERT(mPendingSeekTime.isSome());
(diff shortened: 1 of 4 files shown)


wolfbeast:
confirmed proper init and play after the patch.

Issue #993, opened 2017-03-31
PageMenu.jsm - Ci.nsIXULContextMenuBuilder no longer exists (throws an error)


janekptacijarabaci:
__Steps to reproduce__

1) Go to: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu

2) Go to: `Examples` - `Context menu`

3) Click to button: `OPEN IN JSFIDDLE`

4) Open the context menu... - throws an error in Browser Console:

```
NS_ERROR_XPC_BAD_CONVERT_JS: Could not convert JavaScript argument arg 0
[nsISupports.QueryInterface] PageMenu.jsm:48:0
```

See also:
http://xref.palemoon.org/palemoon-rel25/search?string=nsIXULContextMenuBuilder&find=&findi=&filter=%5E%5B%5E%5C0%5D*%24&hitlimit=&tree=palemoon-rel25
http://xref.palemoon.org/palemoon-rel25/search?string=nsIXULContextMenuBuilder&find=&findi=&filter=%5E%5B%5E%5C0%5D*%24&hitlimit=&tree=palemoon-rel26
https://bugzilla.mozilla.org/show_bug.cgi?id=1066383

wolfbeast:
Resolved by #1013

PR #1013, opened 2017-04-09
[suggestion] PageMenu.jsm - Ci.nsIXULContextMenuBuilder no longer exists (throws an error), allow separator lines

janekptacijarabaci:
Ad:
https://github.com/MoonchildProductions/Pale-Moon/issues/993

Ad also:
https://github.com/MoonchildProductions/Tycho/issues/31

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=870388

---

You add the label `Verification needed`, please.

---

I've created the new build (x32, Windows) and tested
(only: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu).

### DIFF
--- a/browser/modules/PageMenu.jsm
+++ b/browser/modules/PageMenu.jsm
@@ -11,80 +11,175 @@ PageMenu.prototype = {
   PAGEMENU_ATTR: "pagemenu",
   GENERATEDITEMID_ATTR: "generateditemid",
 
-  popup: null,
-  builder: null,
+  _popup: null,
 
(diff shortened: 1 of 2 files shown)


wolfbeast:
I still think it's a nonsense tag, but hey, who am I.

wolfbeast:
Looks like there may be some fallout from this:

https://forum.palemoon.org/viewtopic.php?f=29&t=15639

janekptacijarabaci:
The admin account was not used for tests. More details are needed.

PR #1015, opened 2017-04-10
PageMenu.jsm (follow up) - removal of redundant code, style clean up

janekptacijarabaci:
Removal of redundant code (e10s).

Style clean up:
https://hg.mozilla.org/mozilla-central/diff/239dd56733c4/toolkit/modules/PageMenu.jsm

---

I've created the new build (x32, Windows) and tested
(only: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu).

### DIFF
--- a/browser/modules/PageMenu.jsm
+++ b/browser/modules/PageMenu.jsm
@@ -12,8 +12,6 @@ PageMenu.prototype = {
   GENERATEDITEMID_ATTR: "generateditemid",
 
   _popup: null,
-
-  // Only one of builder will end up getting set.
   _builder: null,
 
   // Given a target node, get the context menu for it or its ancestor.
--- a/browser/modules/PageMenu.jsm
+++ b/browser/modules/PageMenu.jsm
@@ -13,13 +13,11 @@ PageMenu.prototype = {
 
   _popup: null,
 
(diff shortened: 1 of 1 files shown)


browser/modules/PageMenu.jsm
```diff
@@ -13,13 +13,11 @@ PageMenu.prototype = {
...
   _popup: null,
 
-  // Only one of builder or browser will end up getting set.
+  // Only one of builder will end up getting set.
   _builder: null,
-  _browser: null,
 
   // Given a target node, get the context menu for it or its ancestor.
   getContextMenu: function(aTarget) {
-    let pageMenu = null;
```
wolfbeast: Why are you removing this?

janekptacijarabaci: It is not used in the function body. It is a local variable.

wolfbeast: You're wrong. It's used by the caller. if you remove this, then the previous context menu won't be cleared out.

wolfbeast: Oh right.. that's weird. I think that's a mistake and should not have the let keyword. pageMenu is used by the caller, I'm sure the intention is to clear that out.

Issue #841, opened 2017-01-16
MP3 seeking will intermittently make it jump to the end and stop playing.


wolfbeast:
What the title says.
As part of my testing for #833 I noticed some odd behavior when seeking in the MP3 file. There would sometimes be a noticeable delay before playing resumes and other times playing would completely stop, with the cursor jumping to the very end of the stream.

This is also the case in 27.0.3-release, so not related to anything added in that commit, hence a separate issue.

wolfbeast:
BMO references that may be relevant:
https://bugzilla.mozilla.org/show_bug.cgi?id=1089586
https://bugzilla.mozilla.org/show_bug.cgi?id=1208953
https://bugzilla.mozilla.org/show_bug.cgi?id=1159343
https://bugzilla.mozilla.org/show_bug.cgi?id=1163667

wolfbeast:
This is also present in v27 a1, prior to most media changes. Likely this is a media bug in the underlying Mozilla code in the v38 platform.

wolfbeast:
I can't seem to reproduce this with MP4 video, so assuming this is MP3 only.

wolfbeast:
OK, this is not present in our latest common code point with Firefox platform code, so it must have been introduced early in Tycho development. That makes this a regression in our court.

wolfbeast:
Made a mistake. a1 does not display this issue, public beta 1 does.

wolfbeast:
Last good commit: eaa21516c7c71e24314fced1e834f5c41f6880b4 
First bad commit: 5167abf3947837e01586ae8a70e05f897a6b51a8 

Culprit: Making MDSM async in PR #540 -- @trav90 this is your work; did you make a mistake in porting that code or does it rely on other code changes... and TBH, why do we even need this to be async?

trav90:
I think the culprit is commit a0125e4, which was landed by Mozilla (together with a few other fixes) to fix  regression caused by the MDSM changes in PR #540. However, it's apparently also talked about in a different BMO bug where it's mentioned as a "nice to have" instead of actually a fix to the problem. :roll_eyes: I think it should be safe to back out that commit. I'll run a build and confirm.

trav90:
Should be resolved by PR #843.

wolfbeast:
I built in parallel and can confirm.

PR #1017, opened 2017-04-11
Re-land "Dispatch AudioSink notifications asynchronously" + fixes

trav90:
I have been putting a lot of work recently into implementing some so-called "state machinery", which will be a major step towards stabilizing and reducing the raciness of our media code as a whole, as well as being a prerequisite to updating our MSE code. However, without AudioSink notifications being dispatched async, the state machinery does not work as-intended and instead segfaults reliably.

Dispatching AudioSink notifications async previously landed in commit a0125e4, but was backed out for causing a regression as noted in #841. This PR re-lands it, with these additional fixes:

- Make AudioSink:mWritten an Atomic (fixes a potential crash point).
- Suppress call to DispatchOnAudioSinkComplete (fixes an issue that could cause the decode to enter COMPLETED state before reaching the end of the stream).
- Add a cancelable runnable to fix a problem with slow (and in some cases incorrect) seeking.
- Rename OMX AudioSink to GonkAudioSink to prevent conflicts with the new, above mentioned runnable.

Built on Linux and tested playing/seeking the various audio formats here: http://hpr.dogphilosophy.net/test/. Appears to work as intended without introducing any regressions.

### DIFF
--- a/dom/media/AudioSink.cpp
+++ b/dom/media/AudioSink.cpp
@@ -257,7 +257,10 @@ AudioSink::Cleanup()
   AssertCurrentThreadInMonitor();
   nsRefPtr<AudioStream> audioStream;
   audioStream.swap(mAudioStream);
-  mStateMachine->DispatchOnAudioSinkComplete();
+  // Suppress the callback when the stop is requested by MediaDecoderStateMachine.
+  if (!mStopAudioThread) {
+    mStateMachine->DispatchOnAudioSinkComplete();
+  }
 
   ReentrantMonitorAutoExit exit(GetReentrantMonitor());
   audioStream->Shutdown();
--- a/dom/media/AudioSink.cpp
+++ b/dom/media/AudioSink.cpp
@@ -25,6 +25,38 @@ extern PRLogModuleInfo* gMediaDecoderLog;
 #define SINK_LOG_V(msg, ...)
 #endif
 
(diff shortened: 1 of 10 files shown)


JustOff:
Failed to build on win32:

```
 1:29.03 Unified_cpp_dom_media0.cpp
 1:29.05 e:\pale-moon-build\dist\include\mozilla/Atomics.h(901) : error C2338: mozilla/Atomics.h only supports 32-bit and pointer-sized types
 1:29.06         e:\pale-moon-build\dist\include\mozilla/Atomics.h(954) : see reference to class template instantiation 'mozilla::detail::AtomicBase<T,SequentiallyConsistent>' being compiled
 1:29.08         with
 1:29.09         [
 1:29.09             T=int64_t
 1:29.11         ]
 1:29.11         e:\pale-moon-build\dist\include\mozilla/Atomics.h(1010) : see reference to class template instantiation 'mozilla::detail::AtomicBaseIncDec<T,SequentiallyConsistent>' being compiled
 1:29.12         with
 1:29.13         [
 1:29.14             T=int64_t
 1:29.14         ]
 1:29.15         e:\pale-moon\dom\media\AudioSink.h(122) : see reference to class template instantiation 'mozilla::Atomic<int64_t,SequentiallyConsistent,void>' being compiled
```

PR #1025, opened 2017-04-15
Resolve confusions between Flash and VLC plugins on permissions pages

JustOff:
Follow-up to #899, proper display and control Flash and VLC plugins on permissions pages. 

Also changes display name of Shockwave Flash to Adobe Flash in PluginProvider.jsm to unify the UI.

Before:
palemoon_2017-04-15_22-21-27
palemoon_2017-04-15_22-21-56

After:
palemoon_2017-04-15_22-22-45
palemoon_2017-04-15_22-23-27


wolfbeast:
Thanks - not something I would ever have caught because I don't use VLC.

PR #1026, opened 2017-04-16
Re-land "Dispatch AudioSink notifications asynchronously" + (more) fixes

trav90:
Third time's a charm?

Same as #1017, but with the top bullet point reverted as it is not critical and can be revisited later. This should allow it to build on Windows.


JustOff:
> This should allow it to build on Windows.

I confirm.

Issue #823, opened 2017-01-12
Restyle the permissions manager and make it more prominent


wolfbeast:
Since we've been working on improving the permissions manager, we should also consider restyling it and hooking it up to the menu for easy discoverability.

- [x] Add title
- [x] Add a link to the AppMenu and classic Tools menu
- [x] Add/update icons

janekptacijarabaci:
I leave it to you, or someone else :-)

Lootyhoof:
Is this the permissions manager in `about:permissions`, the one in `Page Info`, or both?

wolfbeast:
The permissions manager is the in-content all-site manager in `about:permissions`

wolfbeast:
A quick look shows that it needs a title-heading, and less icons that are mere `?` icons. It could probably do with having its own, more elegant, icon set. The ones used now are very early-2000s. Not 2-tone/flat, though.

Categorization of the actual options and aligning controls would be a good improvement too.

mattatobin:
Personally I would like to see this become more like SeaMonkey's Data Manager and spawned in a window.. In-content is terrible.. Though, that is my opinion.

wolfbeast:
There has never been a XUL version of this so that would require to build something new from scratch entirely. That's really not very useful to invest time in right now. And in this case, I do think in-content at least works for its purpose.

wolfbeast:
Actually, the in-content version of this is XUL -- I noticed when I got around to tinkering with it.
At the very least I'll add a title and hook it up to the AppMenu and classic menu. Icon restyling can be done later.

wolfbeast:
Permissions manager is hooked up now and has a title to the page.
@JustOff - new label created for you to keep track of issues/PRs with string changes. Should make it easier for you.

wolfbeast:
Basic icon updates done. It should be in an acceptable state now.

JustOff:
> @JustOff - new label created for you to keep track of issues/PRs with string changes. Should make it easier for you.

Thanks!

> It should be in an acceptable state now.

I think that `about:permissions` page is too dangerous for common users. Mostly because two things are very confusing: 1) cookies correspond to strict urls, not to domains as on single page permissions 2) there are no easy way to find which permissions were changed on which pages. 

How about to hide the links to `about:permissions` from menu until special preference is set in `about:config`?

wolfbeast:
Why is it too dangerous, exactly? I don't understand what the risk is.
[...]

PR #1027, opened 2017-04-16
Permissions: Cookies correspond to domain, difference between the "default value" and an "actual value" (improvements)

janekptacijarabaci:
Ad https://github.com/MoonchildProductions/Pale-Moon/issues/823#issuecomment-294313936

> 1) Cookies correspond to strict urls, not to domains as on single page permissions
> 2) There are no easy way to find which permissions were changed on which pages

@JustOff
If that's how you meant (partially)...

While it is true that:
> There are no easy way to find which permissions were changed on which pages

...difference between the "default value" and an "actual value" - it may be displayed.

---

### DIFF
--- a/browser/themes/linux/preferences/aboutPermissions.css
+++ b/browser/themes/linux/preferences/aboutPermissions.css
@@ -115,6 +115,13 @@
   font-size: 125%;
   margin-bottom: 0;
   font-weight: bold;
+  margin-right: 0;
+  padding-right: 0;
+}
+
+.pref-default {
+  margin-left: 1ex;
+  padding-left: 0;
 }
 
 .pref-menulist {
@@ -130,4 +137,13 @@
 .plugins-vulnerable {
   margin-left: 0;
   padding-left: 0;
+  margin-right: 0;
+  padding-right: 0;
+}
+
+.plugins-default {
+  margin-left: 1ex;
+  padding-left: 0;
+  margin-right: 2ex;
+  padding-right: 0;
 }
--- a/browser/themes/linux/preferences/aboutPermissions.css
+++ b/browser/themes/linux/preferences/aboutPermissions.css
@@ -120,7 +120,7 @@
 }
 
 .pref-default {
-  margin-left: 1ex;
+  margin-left: 0.5em;
   padding-left: 0;
 }
 
@@ -142,8 +142,8 @@
 }
 
 .plugins-default {
-  margin-left: 1ex;
+  margin-left: 0.5em;
   padding-left: 0;
-  margin-right: 2ex;
+  margin-right: 1em;
   padding-right: 0;
 }
--- a/browser/themes/osx/preferences/aboutPermissions.css
+++ b/browser/themes/osx/preferences/aboutPermissions.css
@@ -123,7 +123,7 @@
 }
 
 .pref-default {
-  margin-left: 1ex;
+  margin-left: 0.5em;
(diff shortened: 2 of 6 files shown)


browser/themes/osx/preferences/aboutPermissions.css
```diff
@@ -118,6 +118,13 @@
   font-size: 125%;
   margin-bottom: 0;
   font-weight: bold;
+  margin-right: 0;
+  padding-right: 0;
+}
+
+.pref-default {
+  margin-left: 1ex;
```
wolfbeast: You should probably use `em` here, not `ex`...

browser/themes/linux/preferences/aboutPermissions.css
```diff
@@ -130,4 +137,13 @@
 .plugins-vulnerable {
   margin-left: 0;
   padding-left: 0;
+  margin-right: 0;
+  padding-right: 0;
+}
+
+.plugins-default {
+  margin-left: 1ex;
```
wolfbeast: You should probably use em here, not ex...

browser/themes/linux/preferences/aboutPermissions.css
```diff
@@ -115,6 +115,13 @@
   font-size: 125%;
   margin-bottom: 0;
   font-weight: bold;
+  margin-right: 0;
+  padding-right: 0;
+}
+
+.pref-default {
+  margin-left: 1ex;
```
wolfbeast: You should probably use em here, not ex...

browser/themes/osx/preferences/aboutPermissions.css
```diff
@@ -134,4 +141,13 @@
 .plugins-vulnerable {
   margin-left: 0;
   padding-left: 0;
+  margin-right: 0;
+  padding-right: 0;
+}
+
+.plugins-default {
+  margin-left: 1ex;
```
wolfbeast: ditto

browser/themes/windows/preferences/aboutPermissions.css
```diff
@@ -118,6 +118,13 @@
   font-size: 125%;
   margin-bottom: 0;
   font-weight: bold;
+  margin-right: 0;
+  padding-right: 0;
+}
+
+.pref-default {
+  margin-left: 1ex;
```
wolfbeast: ditto

browser/themes/windows/preferences/aboutPermissions.css
```diff
@@ -134,4 +141,13 @@
 .plugins-vulnerable {
   margin-left: 0;
   padding-left: 0;
+  margin-right: 0;
+  padding-right: 0;
+}
+
+.plugins-default {
+  margin-left: 1ex;
```
wolfbeast: ditto

wolfbeast:
I just built this and it's not good. Disabling the options for plugins that are set to default is NOT good because people will not be able to change the permissions, which is the whole point of having the permissions manager in place... What were you thinking?

janekptacijarabaci:
OK. You can change it according to your needs :-)
https://github.com/janekptacijarabaci/Pale-Moon/blob/248229018be44049493f2e3ecb16959704afaa20/browser/components/preferences/aboutPermissions.js#L549

wolfbeast:
My mistake -- I didn't really understand what you were doing and misinterpreted the output.

PR #1028, opened 2017-04-16
Introduce state machinery

trav90:
This PR introduces so-called state machinery for use by our media subsystem (see the new header files for a detailed description). TL;DR this state machinery solves a lot of racey conditions in our media code as a whole, and is a _major_ step towards no longer needing the decoder/reentrant monitor that's caused us issues. Additionally, it is needed by our upcoming MSE re-write/update, which will be much more racey and crash prone without it.

In addition to introducing the state machinery, this PR adjusts our threading/dispatching behavior, and fixes a couple crash points made visible by the state machinery.

I've been testing on Twitch, YouTube and Vimeo and it appears to be working as intended, but further testing (mainly on Windows) will be needed.

Note that #1026 MUST land before this. The state machinery expects async notifications, and _will_ crash without it.


wolfbeast:
Depends on #1026

PR #1029, opened 2017-04-16
Extend the "highlight all" option behavior in findbar

wsdfhjxc:
These changes are targeted at #364. 

Preference `accessibility.typeaheadfind.highlightallbydefault` indicates the default state of "highlight all" option after opening a findbar, i.e. `true` - enabled, `false` - disabled.

Preference `accessibility.typeaheadfind.highlightallremember` indicates whether the "highlight all" option state should be remembered after closing a findbar or not. If it's set to `true`, then the previous preference gets overwritten with the "highlight all" option state from recently closed findbar.

If both preferences are set to `false`, then original behavior is preserved.

Tested on Linux and observed no negative side-effects.

### DIFF
--- a/browser/base/content/browser.js
+++ b/browser/base/content/browser.js
@@ -3824,9 +3824,6 @@ var XULBrowserWindow = {
           // Close the Find toolbar if we're in old-style TAF mode
           gFindBar.close();
         }
-
-        // fix bug 253793 - turn off highlight when page changes
-        gFindBar.getElement("highlight").checked = false;
       }
     }
     UpdateBackForwardCommands(gBrowser.webNavigation);
--- a/browser/base/content/browser.js
+++ b/browser/base/content/browser.js
@@ -3824,6 +3824,9 @@ var XULBrowserWindow = {
           // Close the Find toolbar if we're in old-style TAF mode
           gFindBar.close();
         }
+
+        // fix bug 253793 - turn off highlight when page changes
+        gFindBar.getElement("highlight").checked = false;
       }
     }
     UpdateBackForwardCommands(gBrowser.webNavigation);
--- a/browser/base/content/browser.js
+++ b/browser/base/content/browser.js
@@ -3825,8 +3825,7 @@ var XULBrowserWindow = {
           gFindBar.close();
         }
         
-        if (!(gPrefService.getBoolPref("accessibility.typeaheadfind.highlightallremember") ||
-              gPrefService.getBoolPref("accessibility.typeaheadfind.highlightallbydefault"))) {
+        if (gPrefService.getIntPref("accessibility.typeaheadfind.highlightAllBehavior") == 0) {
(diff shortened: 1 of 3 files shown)


wolfbeast:
It's probably better to make it a single Int pref instead of 2 bools, especially considering one will overwrite the other.
e.g. `accessibility.typeaheadfind.highlight_behavior`:
0 = original behavior (don't highlight, don't remember)
1 = always highlight
2 = remember last state

wsdfhjxc:
> It's probably better to make it a single Int pref instead of 2 bools, especially considering one will overwrite the other.
> e.g. accessibility.typeaheadfind.highlight_behavior:
> 0 = original behavior (don't highlight, don't remember)
> 1 = always highlight
> 2 = remember last state

In that case, without external reference, the "highlight all" option state selected and left by the user wouldn't be remembered globally across browser restarts and after using findbars in multiple windows.

I made it a 4-value preference instead. The value of 0 means the default behavior, value of 1 means to always enable "highlight all" when a findbar is opened and any value greater than 1 means that the "highlight all" option state should be remembered. If the latter, then specific value of 2 or 3 is automatically set depending on the "highlight all" option state from the last closed findbar. In this case, it's used as a reference to determine the initial state, e.g. after browser restart.

I think it should cover that properly. If not, this can be ditched eventually, as there is already a nice extension which can do this and much more (FinderBar Tweak).

wolfbeast:
Ugh, that becomes muddy real quick. I think it's better to K.I.S.S. it instead then, and only have a pref to indicate preferred initial state, either highlight or not, and drop the whole "remember" bit.

JustOff:
> I made it a 4-value preference instead.

I read it twice, but still can't understand (not looking into the code) ;-)

> drop the whole "remember" bit

Or maybe return the two prefs?

wolfbeast:
No, the two prefs are just as confusing.

JustOff:
I'm just worried about not to get again [the problem we had on previous iteration](https://github.com/MoonchildProductions/Pale-Moon/pull/904#issuecomment-280913548).

wolfbeast:
How would an extra pref to indicate it should be remembered make any different to that?

wolfbeast:
I have a better idea.
How about we skip all this manual stuff and just let it remember the state, always, and store that in a pref?

That way users who want to generally highlight all can just check it and it will be the default from then on. And the other way around.

wolfbeast:
You know what, scratch it, the 2-bool option is going to be just fine. maximum control, as long as people understand what happens when they choose to "remember"

wolfbeast:
@wsdfhjxc if you can revert your last commit to go back to 2 booleans, I'll merge it in. Sorry about the indecision and trying to come up with something more efficient -- doesn't look like there is an easy option to do so, and your initial idea was best.

wolfbeast:
Or, I'll just do it myself, then.

PR #1030, opened 2017-04-16
Allow imagedocument.png to be themeable

FranklinDM:
Includes imagedocument.png to the base theme and allows it to be theme-able by custom themes.
(hasn't been tested yet)


wolfbeast:
@Lootyhoof This would probably require changes to all themes, correct?

Lootyhoof:
Correct. As this change appears to be related specifically to the favicon shown in the urlbar (though not the tab), when viewing a standalone image, the effects should be minimal (a missing favicon in the urlbar perhaps).

Lootyhoof:
As an aside, it looks like this icon is not included in each platform's `jar.mn` yet, so wouldn't be included in the processed build. These can be found at:

`/browser/themes/*/jar.mn`

FranklinDM:
Thanks. Sorry, I didn't really know much about these since I'm just new to this (can't clone the repo & test, data caps).

Lootyhoof:
Looks good to me!

PR #1035, opened 2017-04-17
Fixed virtualenv multilib support

JamesTheAwesomeDude:
Since Pale Moon rolls its own bundled virtualenv, it missed out on the critical fix at pypa/virtualenv#516

### DIFF
--- a/python/virtualenv/virtualenv.py
+++ b/python/virtualenv/virtualenv.py
@@ -1157,7 +1157,10 @@ def install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages, clear, sy
     else:
         prefix = sys.prefix
     mkdir(lib_dir)
-    fix_lib64(lib_dir, symlink)
+
+    # Account for libdir of explicit bitness
+    is_bitness_explicit, libdir_bitness = fix_libdir_bitness(lib_dir, symlink)
+
     stdlib_dirs = [os.path.dirname(os.__file__)]
     if is_win:
         stdlib_dirs.append(join(os.path.dirname(stdlib_dirs[0]), 'DLLs'))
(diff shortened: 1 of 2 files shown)


wolfbeast:
This is completely outside of my area.
@trav90 or @mattatobin -- can either of you review this before I merge it?

trav90:
My concerns:

If this is a critical fix, why has it been over three years and still not merged into upstream virtualenv?
Mozilla has a bug on file for this ([Bug 975742](https://bugzilla.mozilla.org/show_bug.cgi?id=975742)) where they link to an open issue on virtualenv's repo, with users reporting mixed results.
Comment 5 in the Gentoo bug linked in the commit message seems to indicate that this patch did not work when building FF 44 (which used virtualenv 12.0.5 like we do).
This seems to only affect Gentoo on certain architectures?

trav90:
Seeing as how there has been no movement on this here or at the upstream virtualenv repo for 2 weeks now, I'm going to close this. Feel free to re-submit after the issues are resolved.

JamesTheAwesomeDude:
There are no issues with this; I tested it myself.

The reason it hadn't been accepted into upstream virtuanev was that upstream had changed to become incompatible since then; however, Firefox's (and, by extension, Pale Moon's) in-tree virtualenv is still based off the version that works as-is (well, with very minimal modification—which I have done) with those patches.

Essentially, this simply removes all hardcoded "lib64" references and uses a more correct way of detecting the architecture-specific library directories.

I've done this as part of my attempts to do the legwork to get Pale Moon working on [x32](http://google.com/search?q=x32+abi); one of the first things needed was to get rid of this hacky workaround in virtualenv that only works on x86 and x86_64.

mattatobin:
@JamesTheAwesomeDude The Mozilla-style build system is a tricky thing.. How does your patch affect Windows, Non-Gentoo Linux distros, and potentially Mac?

JamesTheAwesomeDude:
@mattatobin The patch only affects the `lib64` situation, as it only even touches the functions to do with its detection, so it's a reasonable assumption that it would not have implications reaching beyond that.

It causes no breakage when building Pale Moon on other Linux distros, and, while I don't have Mac OS available to me, I can try to set up a Windows development environment to get some test data from it, if that would be desired/helpful towards getting this merged. (I see that AppVeyor supports Windows and TravisCI supports OSX; are those being considered at all for Pale Moon?)

The only known breakage is when setting up a PyPy virtualenv (which is outside of the realm of even sanity for the Mach build system anyway—cPython 2.7 specifically being the only supported runtime)

trav90:
I think verification on Windows should be a requirement before this can be merged in.

> The reason it hadn't been accepted into upstream virtuanev was that upstream had changed to become incompatible since then

So what will happen if and when we update our in-tree version of virtualenv from upstream? Would doing so break the fixes this patch makes?

wolfbeast:
Removing verification needed - nothing actionable has been presented.

PR #1039, opened 2017-04-19
Add support for posting C++11 lambdas to messageloop/threads

trav90:
As part of the media threading/dispatch work Mozilla did, they also began using lambdas in parts of the media code. This PR adds a wrapper that will allow us to also use lambdas in our media code. This PR also makes necessary changes to Mozilla's clang-plugin to account for lambdas.

Compiled/tested on Linux and doesn't appear to break anything. Note that I have _not_ tested the changes to the clang-plugin, since (unless I'm mistaken) it is not used on any platform we currently support.


wolfbeast:
Correct. none of the platforms we officially support are built with clang. I do think Mac is built with clang, but if there's an issue there, I'm sure Sugis can let us know.

PR #1040, opened 2017-04-19
More improvements for the state machinery + use it in even more places

trav90:
This PR eliminates the state machinery ::Holder setup and provides the WatchManager an owner thread, both of which improves the overall ergonomics of the state machinery. Additionally, this PR uses state mirroring for stuff in MediaDecoderStateMachine::SetStateMachineParameters, which  currently accounts for a good portion of the cross-thread calls in our media code.

Tested on Linux (YouTube, Vimeo, Twitch, etc.) and appears to be working as intended.


trav90:
Dependent on #1039 being merged first (otherwise the patches in this PR won't compile).

PR #1043, opened 2017-04-19
Permissions: Reload list of sites, use default, confirm before any action, permissions.places-sites-limit (improvements), style clean up

janekptacijarabaci:
Ad https://github.com/MoonchildProductions/Pale-Moon/issues/823#issuecomment-294831241

> `about:permissions` does not allow to set the permission for site into "use default" state

Added a button.

> Once set using `about:permissions` the plugin state can't be cleared from the database the same way.

By "cleared" do you mean disabled? For global: I don't know how to do it better. See:
http://xref.palemoon.org/palemoon-trunk/search?string=.userDisabled&find=AddonManager.jsm

> `Forget About This Site` and `Remove [All] Cookies` buttons immediately remove the corresponding data without any warnings or confirmation dialogs.

It was added.

> `about:permissions` does not provide the filter to display only sites with modified permissions that could help to easily find the culprit in case of any problems.

Added configurable parameter (`about:config`): `permissions.places-sites-limit`.

> `about:permissions` does not allow to set the permission for site if target value equals current default.

> `pageinfo>permissions` has several mistakes in control logic which prevent to set specific permissions in case when global permissions for all sites differ from default values

__I don't have time to look at it just now...__

---

### DIFF
--- a/browser/components/preferences/aboutPermissions.js
+++ b/browser/components/preferences/aboutPermissions.js
@@ -2,9 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-var Cc = Components.classes;
-var Ci = Components.interfaces;
-var Cu = Components.utils;
+let Cc = Components.classes;
+let Ci = Components.interfaces;
+let Cu = Components.utils;
 
 Cu.import("resource://gre/modules/Services.jsm");
 
@@ -14,15 +14,15 @@ Cu.import("resource://gre/modules/ForgetAboutSite.jsm");
 Cu.import("resource://gre/modules/NetUtil.jsm");
 Cu.import("resource://gre/modules/PluralForm.jsm");
 
-var gFaviconService = Cc["@mozilla.org/browser/favicon-service;1"].
+let gFaviconService = Cc["@mozilla.org/browser/favicon-service;1"].
                       getService(Ci.nsIFaviconService);
 
-var gPlacesDatabase = Cc["@mozilla.org/browser/nav-history-service;1"].
+let gPlacesDatabase = Cc["@mozilla.org/browser/nav-history-service;1"].
                       getService(Ci.nsPIPlacesDatabase).
                       DBConnection.
                       clone(true);
 
-var gSitesStmt = gPlacesDatabase.createAsyncStatement(
+let gSitesStmt = gPlacesDatabase.createAsyncStatement(
                  "SELECT get_unreversed_host(rev_host) AS host " +
                  "FROM moz_places " +
                  "WHERE rev_host > '.' " +
@@ -31,12 +31,12 @@ var gSitesStmt = gPlacesDatabase.createAsyncStatement(
                  "ORDER BY MAX(frecency) DESC " +
                  "LIMIT :limit");
 
-var gVisitStmt = gPlacesDatabase.createAsyncStatement(
+let gVisitStmt = gPlacesDatabase.createAsyncStatement(
                  "SELECT SUM(visit_count) AS count " +
                  "FROM moz_places " +
                  "WHERE rev_host = :rev_host");
 
-var gFlash = {
+let gFlash = {
   name: "Shockwave Flash",
   betterName: "Adobe Flash",
   type: "application/x-shockwave-flash",
@@ -47,7 +47,7 @@ var gFlash = {
  * to testPermission. This is based on what consumers use to test these
  * permissions.
  */
-var TEST_EXACT_PERM_TYPES = ["desktop-notification", "geo", "pointerLock"];
+const TEST_EXACT_PERM_TYPES = ["desktop-notification", "geo", "pointerLock"];
 
 /**
  * Site object represents a single site, uniquely identified by a host.
@@ -80,11 +80,11 @@ Site.prototype = {
     }
 
     // Try to find favicon for both URIs, but always prefer the https favicon.
-    gFaviconService.getFaviconURLForPage(this.httpsURI, function (aURI) {
+    gFaviconService.getFaviconURLForPage(this.httpsURI, function(aURI) {
       if (aURI) {
         invokeCallback(aURI);
       } else {
-        gFaviconService.getFaviconURLForPage(this.httpURI, function (aURI) {
+        gFaviconService.getFaviconURLForPage(this.httpURI, function(aURI) {
           if (aURI) {
             invokeCallback(aURI);
           }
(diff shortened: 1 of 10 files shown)


JustOff:
> Added a button.

In `pageinfo>permissions` set permission for site into "use default" state means clear corresponding record from `permissions.sqlite`, not set it to default value as you made.

> By "cleared" do you mean disabled?

See above.

> Added configurable parameter (`about:config`): `permissions.places-sites-limit`.

You mean set it to 0 for view only sites with modified permissions? Then this trick works.

PS: If I would have decided to work on `about:permissions` task I probably just use the view and code for single site from `pageinfo>permissions` as it has almost all necessary (including default state handling) and works almost fine. But since I'm not, I can only advise to do so.

janekptacijarabaci:
Thank you for the feedback.

> You mean set it to 0 for view only sites with modified permissions?

Yes.

browser/locales/en-US/chrome/browser/preferences/aboutPermissions.properties
```diff
@@ -2,6 +2,10 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+# LOCALIZATION NOTE (permissionsConfirm): Confirm before any action.
+# See: http://developer.mozilla.org/en/docs/Localization_and_Plurals
+permissionsConfirm=Are you sure you want to continue?
```
wolfbeast: Very/too vague... continue with what?

browser/locales/en-US/chrome/browser/preferences/aboutPermissions.dtd
```diff
@@ -15,8 +15,12 @@
 
 <!ENTITY header.defaults                 "Default Permissions for All Sites">
 
+<!ENTITY permissions.sitesReload         "Reload list of sites">
+
 <!ENTITY permissions.forgetSite          "Forget About This Site">
 
+<!ENTITY permission.default              "Set a default value">
```
wolfbeast: Set a default value for what? What is the default value? Very vague...

If this is the button label, IIUC, then it should just be "Use Default" or similar.

browser/components/preferences/aboutPermissions.js
```diff
@@ -2,44 +2,52 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-let Ci = Components.interfaces;
-let Cc = Components.classes;
-let Cu = Components.utils;
+var Cc = Components.classes;
+var Ci = Components.interfaces;
+var Cu = Components.utils;
```
wolfbeast: Why are you changing all these `let` keywords to `var`??

browser/components/preferences/aboutPermissions.js
```diff
@@ -117,7 +125,7 @@ Site.prototype = {
    *
    * @param aType
    *        The permission type string stored in permission manager.
-   *        e.g. "cookie", "geo", "indexedDB", "popup", "image"
+   *        e.g. "image", "popup", "cookie", "geo", "indexedDB"
```
wolfbeast: Why are you changing the order here?

browser/components/preferences/aboutPermissions.js
```diff
@@ -156,7 +164,7 @@ Site.prototype = {
...
```
wolfbeast: Ditto, for order?

browser/components/preferences/aboutPermissions.js
```diff
@@ -185,7 +193,7 @@ Site.prototype = {
...
```
wolfbeast: ditto, for order?

browser/components/preferences/aboutPermissions.js
```diff
@@ -95,7 +103,7 @@ Site.prototype = {
     let rev_host = this.host.split("").reverse().join("") + ".";
     gVisitStmt.params.rev_host = rev_host;
     gVisitStmt.executeAsync({
-      handleResult: function(aResults) {
+      handleResult: function (aResults) {
```
wolfbeast: Why are you adding spaces between function and parentheses...?
common coding style is `function()` not `function ()` -- please don't make unnecessary and counter-practice style changes.

(also for all others below where you did this)

browser/components/preferences/aboutPermissions.js
```diff
@@ -646,36 +685,40 @@ let AboutPermissions = {
   },
 
   observe: function (aSubject, aTopic, aData) {
-    switch(aTopic) {
+    switch (aTopic) {
```
wolfbeast: Same issue as with `function()` -- common style = no space, please don't change this.

JustOff:
Btw, are you intentionally use different font style for 'Use default' buttons?

palemoon_2017-04-19_22-27-14

janekptacijarabaci:
@JustOff 
> are you intentionally use different font style for 'Use default' buttons?

It's only the smaller font. Do you have a better idea? Maybe another font style?! I don't know...

JustOff:
> I don't know...

Ok, never mind.

wolfbeast:
There's still the question "Are you sure you want to continue?" -- when is this used? It's too vague.

wolfbeast:
Please remove the confirmation dialog boxes. It's unnecessary and just annoying.

janekptacijarabaci:
@wolfbeast 
Ad "aboutPermisions.xul and (thrice):
https://github.com/MoonchildProductions/Pale-Moon/pull/1043/files#diff-9b3cb3249f3a1f077eac471d10b882baR64

Please replace it for what you think is best.

JustOff:
> Please remove the confirmation dialog boxes. It's unnecessary and just annoying.

Sorry for insisting, but I think it's very dangerous decision. We probably may just remove 'Remove Cookies' buttons at all 'cause such action can be done using 'Manage Cookies' either one by one or completely. And as for 'Forget About This Site' button I'm sure the confirmation dialog should be there.

janekptacijarabaci:
> Please remove the confirmation dialog boxes.

Unfortunately, I can't do it now.

wolfbeast:
@JustOff we don't ask for permissions in other places where you have these options, there is no reason whatsoever to suddenly introduce it here, which is supposed to be an interface for quick management of site permissions and per-site settings.

None of these settings are dangerous to use.

@janekptacijarabaci All right, I'll merge it and remove it myself.

wolfbeast:
@JustOff FYI: this is going to be the last accepted commit with locale changes for 27.3.

mattatobin:
If they can't in time it would be best to make sure untraslated english strings are at least existant for all active langpacks by release day.

JustOff:
@mattatobin 
That will be done automatically next time I build the project on CrowdIn.

PR #1049, opened 2017-04-22
Cookies - sub1.example.com vs. sub2.sub1.example.com: Forget About site (+ improvements, follow up); Permissions (+ fix comments, follow up)

janekptacijarabaci:
## 1)

Ad https://forum.palemoon.org/viewtopic.php?f=50&p=112386#p112386

> The `getCookiesFromHost` method will only match `sub1.example.com` and `example.com` if aDomain = 
`sub1.example.com`. It will not match `sub2.sub1.example.com`.

## 2)
Ad https://github.com/MoonchildProductions/Pale-Moon/issues/579#issuecomment-257860811

> When I want to perform that command on a site that has a large number of entries (as shown on the screenshot; I chose CNN.com for this example) in the History menu, two things happen:
> - It only deletes one or a couple of entries instead of all of them.
> - The Forget About This Site and Delete This Page commands become unresponsive. When I try to apply them to any site in the History menu, they simply won't work.
>> I don't think it works as it should with 27b3. I just tried it on my desktop and on a portable x86 on my laptop and the issue is still there.

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=1247201

---

## 3)

__3a)__
Fix Permissions - if the master password is set.

__3b)__
Follow up:
The indexOf() method is case sensitive!
`User canceled Master Password entry` !== `User canceled master password entry`

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=1358772

---

### DIFF
--- a/browser/components/places/content/controller.js
+++ b/browser/components/places/content/controller.js
@@ -228,7 +228,8 @@ PlacesController.prototype = {
       }
       else
         host = NetUtil.newURI(this._view.selectedNode.uri).host;
-      ForgetAboutSite.removeDataFromDomain(host);
+      ForgetAboutSite.removeDataFromDomain(host)
+                     .catch(Components.utils.reportError);
       break;
     case "cmd_selectAll":
       this.selectAll();
--- a/browser/components/preferences/aboutPermissions.js
+++ b/browser/components/preferences/aboutPermissions.js
@@ -224,21 +224,19 @@ Site.prototype = {
   },
 
   /**
-   * Gets cookies stored for the site. This does not return cookies stored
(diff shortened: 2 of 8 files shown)


janekptacijarabaci:
@wolfbeast 
I propose to consider this as `Uplift Wanted` (27.3.0).

wolfbeast:
Too late for 27.3.0 - already in RelEng stage.

janekptacijarabaci:
Is it too late for 27.3.1 (if that will be the case)?

wolfbeast:
I don't know if there will be a point release. if so, I'll keep it in mind.

PR #1054, opened 2017-04-24
Misc. media fixes

trav90:
This PR contains several small media fixes I came across while researching improving our MSE code. Changes:

Fixes a bug/prevents out-of-bounds memory access in libstagefright.
Makes small adjustment to improve A/V sync in some situations.
Fixes a small calculation bug in MediaDecoder::UpdateStreamBlockingForStateMachinePlaying().
Prevents a promise being disconnected prematurely if it is still waiting on calls from another thread (only affects MSE).
Fixes a data eviction issue that would leave MP4 + MSE streams unplayable or cause a stall when changing resolutions.

Tested on Linux and appears to be working as intended.

### DIFF
--- a/dom/media/MediaDecoder.cpp
+++ b/dom/media/MediaDecoder.cpp
@@ -477,8 +477,7 @@ void MediaDecoder::UpdateStreamBlockingForStateMachinePlaying()
     return;
   }
   bool blockForStateMachineNotPlaying =
-    mDecoderStateMachine && !mDecoderStateMachine->IsPlaying() &&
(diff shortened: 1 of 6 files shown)


wolfbeast:
Looks like this fixes playback of http://beta.medici.tv/en/operas/the-magic-flute-mozart-salzburg-festival-pierre-audi-riccardo-muti/ ...?

JustOff:
> Looks like this fixes playback

Can't confirm this.

wolfbeast:
Ah, with this in, it seems to fall back to flash. That's why.

Issue #1033, opened 2017-04-17
Improve our support for MSE and make it spec-compliant


trav90:
I'm creating this to track the progress of updating our current MSE code to be more spec compliant and provide better MSE support. Things that need done (not necessarily in this order):

- [x] Implement state machinery to reduce raceyness of our media code (both current and new).
- [x] Research and make necessary changes (if any) to Stagefright/MoofParser to work with new code.
- [x] Research and make necessary changes to the MediaFormatReader and our decoders (if needed).
- [x] Abstact the TrackBuffer interface to make it easier to update the surrounding code.
- [x] Make MediaPromises work together with lambdas.
- [x] Implement Completion MediaPromises and MediaPromise::All
- [x] Rewrite our TrackBuffer/SourceBuffer code according to spec.
- [x] Add a new (mostly) spec-compliant MediaSourceDemuxer object (behind a pref) that will work together with the MediaFormatReader. This will be the bulk of our new MSE-specific code.
- [x] Research if any additional areas of our MSE code need to be aligned to the spec.
- [x] Optimize new architecture to prevent performance regressions.
- [x] Fix remaining regressions (such as stalls when changing resolutions, a/v desync issues, etc).
- [x] Enable new MediaSourceDemuxer by default.

Targeting the 27.3.0 release, but may have to be pushed back depending on amount of time needed to get all this working.

trav90:
Point 1 was fixed by #1028. Point 4 was completed by #916.

wolfbeast:
I'm going to bump remaining open issues to 27.4 -- 27.3 needs to be published next week somewhere due to some sec issue deadlines; we'll just have to consider this an ongoing effort for now.

trav90:
Agreed. Based on the research I've been doing overnight, I think getting this done for 27.4 is doable and should be the target.

trav90:
Added point 5 to list above. While not directly needed, by allowing MediaPromise to use lambdas there will be less reverse-engineering required for the actual MSE patches.

wolfbeast:
Marking point 5 done since I've tested it on Windows (BBC news, YT, Twitch, Vimeo) with #1066 in, and seems to play fine.

One glaring issue we still have is twitch streams having terrible video/audio de-sync.

wolfbeast:
Also, it's not entirely clear for testing if I need to flip a pref or not for testing... If so, which one?

trav90:
Marking point 2 complete also (#1056).

[...]

PR #1056, opened 2017-04-26
Re-land "Properly handle MP4 time offset in MoofParser"

trav90:
This originally landed as part of #707, but was backed out in #838 for causing #831. However, it turns out that this is needed after all (according to Mozilla comments on BMO) as it will be used/is needed by our new MSE code.

Also included in this PR is a new test file. In 27.3.0 and current trunk, there is a noticeable audio/video desync when playing the new bipbop-lateaudio.mp4 file added here. With this PR merged, the desync issue for that file is fixed.

I've tested this pretty extensively on Linux, and all appears to be working as intended. However, this previously introduced a regression on Windows, which I have no way to test so am going to need someone to verify for me. The state of our media code has changed considerably since January, so I'm hoping that this will be similar to #1026 in that it now works without introducing any regressions.

If this still causes regressions, I'll try to determine the cause or see if it's possible to work around.


trav90:
Tag #1033, point 2.

JustOff:
At least Dailymail video (#831) works fine for me with (and without) this PR on Win7.

trav90:
Since we've gotten no further reports, considering this working.

Issue #1059, opened 2017-04-27
comit 8b3e92a broke site icetrade.by


Feodor2:
regarding https://forum.palemoon.org/viewtopic.php?f=12&t=15299

These code pile do affect
`  if (aType == MIN_ISIZE &&
     (((styleISize.HasPercent() || styleMaxISize.HasPercent()) &&
       aFrame->IsFrameOfType(nsIFrame::eReplacedSizing)) ||
     (styleISize.HasPercent() &&
      aFrame->GetType() == nsGkAtoms::textInputFrame))) {
`
After i restored it to previous, the site problem was gone. Notice that there is no problem with other browsers. It is to find more sites with similar elements to test.

JustOff:
See #786.

JustOff:
Last bad: 2016-02-24
First good: 2016-02-25
Pushlog: https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=5b2baa5e9356644a7ed0b73e422eaff62e159ffb&tochange=918df3a0bc1c4d07299e4f66274a7da923534577

JustOff:
Probably https://bugzilla.mozilla.org/show_bug.cgi?id=1250342

janekptacijarabaci:
@JustOff 

:+1:

JustOff:
@janekptacijarabaci 
It seems my initial guess was wrong and actually the fix is https://bugzilla.mozilla.org/show_bug.cgi?id=1247929, but since you ported  both of them it solves the case. Nice shot!

PR #786, opened 2016-12-25
CSS: Percentage max-width does not seem to affect contributions to intrinsic min-width

janekptacijarabaci:
Ad https://forum.palemoon.org/viewtopic.php?f=26&p=102056

---

Last bad: Firefox 46.0a2 (2016-02-09)
Built from https://hg.mozilla.org/releases/mozilla-aurora/rev/6a9b6a1e24545e0836150269df83bc53bea0db2d

First good: Firefox 46.0a2 (2016-02-10)
Built from https://hg.mozilla.org/releases/mozilla-aurora/rev/274c4d8b82caf0d7166b914e6772a6d1c4ca167b

Pushlog
http://hg.mozilla.org/releases/mozilla-aurora/pushloghtml?fromchange=6a9b6a1e24545e0836150269df83bc53bea0db2d&tochange=274c4d8b82caf0d7166b914e6772a6d1c4ca167b

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=823483
See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=1174450
(https://hg.mozilla.org/mozilla-central/diff/06f474c578d6/layout/base/nsLayoutUtils.cpp)

---

I've created the new build (x64) and tested.

__But please check it in detail (all options).__

PR #1060, opened 2017-04-27
CSS: text-align: true -> text-align: unsafe; sizing problems with some form fields like <input type="number">

janekptacijarabaci:
# 1)

Ad https://github.com/MoonchildProductions/Pale-Moon/issues/1059
(https://forum.palemoon.org/viewtopic.php?f=12&t=15299)

# 2)

The table layout algorithm has sizing-problems with some form fields like \<input type="number"\>

---

An example:
http://www.icetrade.by/
(Advanced search - Расширенный поиск)

An example:
https://jsfiddle.net/o1tkax4t/
https://bug1247929.bmoattachments.org/attachment.cgi?id=8719124

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1250342
https://bugzilla.mozilla.org/show_bug.cgi?id=1247929

---

I've created the new build (x32, Windows) and tested (+ https://github.com/MoonchildProductions/Pale-Moon/pull/786: http://www.corvetteforum.com/forums/c3-tech-performance/3909764-what-is-this-hole-under-fuel-pump.html).


wolfbeast:
This resolves #1059

PR #1063, opened 2017-04-28
Reset general.useragent.locale if gBrowser isn't defined on start

### LINKED ISSUE #718: Reset general.useragent.locale on upgrade?
Given the the number of problems occurring during the upgrade from 26 to 27 may be it makes sense to simply reset general.useragent.locale to en-US?

JustOff:
Trying to auto recover browser if it doesn't start due to incompatible or outdated language pack. The fuse code is injected into padlock.xul/js because it has no localized strings and works regardless of possible browser.xul errors.

Resolves #718.

### DIFF
--- a/browser/base/content/padlock.js
+++ b/browser/base/content/padlock.js
@@ -122,6 +122,12 @@ var padlock_PadLock =
   },
   prefbranch : null,
   onLoad: function() {
+    if (typeof gBrowser === "undefined") {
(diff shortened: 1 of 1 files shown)


wolfbeast:
Since the code it in a rather specific and non-logical location (unrelated to the padlock), can you add a comment block to it explaining why it's there and what it does?

wolfbeast:
LGTM!

wolfbeast:
One thing that should probably be done is inform the user and/or automatically restart after the language has been reset.

PR #1066, opened 2017-04-29
Make MediaPromises work with lambdas

trav90:
This PR makes our MediaPromise implementation work with lambdas. Additionally (thanks to both lambdas and AbtractThread), some code in the MDSM has been cleaned up.

Our new MSE code will be using lambdas, so getting this working first will reduce some of the reverse-engineering that will be needed while porting patches from Mozilla.

Tested multiple video sites/formats etc. on Linux and appears to be working as intended.


trav90:
#1033, point 5.

PR #1068, opened 2017-04-29
Smart browser recovery in case of incompatible language pack and not only

JustOff:
Follow-up to #1063 and https://github.com/MoonchildProductions/Pale-Moon/commit/7988657bf5d5bc7fa7f7ac84acdacf6826ac552b:

* allow to choose between locale reset and restart in safe mode
* allow to see actual error
* try to catch not only LP error, but possible others, like from bad add-ons etc

### DIFF
--- a/browser/base/content/padlock.js
+++ b/browser/base/content/padlock.js
@@ -124,16 +124,46 @@ var padlock_PadLock =
   onLoad: function() {
     // gBrowser should be always defined at this point, but if it is not then most likely 
     // it is due to incompatible or outdated language pack is installed and selected.
-    // In this case we reset "general.useragent.locale" to try to recover browser startup.
     // This fuse code is placed here because padlock.xul has no localized strings and 
     // therefore works regardless of possible browser.xul errors.
     if (typeof gBrowser === "undefined") {
-      var prefBranch = Cc["@mozilla.org/preferences-service;1"].getService(Ci.nsIPrefBranch);
-      if (prefBranch.prefHasUserValue("general.useragent.locale")) {
-        prefBranch.clearUserPref("general.useragent.locale");
-        a=Ci.nsIAppStartup,Cc["@mozilla.org/toolkit/app-startup;1"].getService(a).quit(a.eRestart | a.eAttemptQuit);
+      var localePref = "general.useragent.locale", a = Ci.nsIAppStartup;
+      // Catch any possible errors to be safe.
+      try {
+        var prefBranch = Cc["@mozilla.org/preferences-service;1"].getService(Ci.nsIPrefBranch),
+            prompts = Cc["@mozilla.org/embedcomp/prompt-service;1"].getService(Ci.nsIPromptService);
+        // If "general.useragent.locale" is defined then we will prompt to reset it or restart in safe mode.
+        if (prefBranch.prefHasUserValue(localePref)) {
+          // Save and reset current locale to avoid possible fail in nsIPromptService.
+          var lang = prefBranch.getCharPref(localePref);
+          prefBranch.clearUserPref(localePref);
+          // Display prompt dialog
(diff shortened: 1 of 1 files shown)


wolfbeast:
Needs a bit of re-wording in the prompts, but I'll take care of that.

wolfbeast:
I worked on this and after a few hours of tinkering and trying variations, I've backed it out. It's much too complex, the "select()" can't have multi-line explanations above the selection, and there is severe risk of endless restart loops in the way it's implemented in this PR which **must** be avoided **at all cost**.

I've added an `alert()` to inform the user what is going on, but otherwise reverted to a straight-forward one-shot check and recovery fix. If the browser is severely broken because of something else, then it's likely that a Safe Mode restart won't magically recover it (and potentially won't prompt the user in that case), and it'll be better in that case to let normal error handling take place.

JustOff:
> there is severe risk of endless restart loops in the way it's implemented in this PR which must be avoided at all cost.

Sorry, my fault, but I could fix this with a simple check if browser is already running in safe mode.

> I've added an alert() to inform the user what is going on, but otherwise reverted to a straight-forward one-shot check and recovery fix.

I would prefer at least keep option to restart in safe mode for LP-related issues, but you decide.

wolfbeast:
> I could fix this with a simple check if browser is already running in safe mode.

That actually won't help, because if you get into the situation where the browser doesn't initialize in safe mode, then there is no other recovery option and the browser is completely broken. So having a code path for that is rather pointless.

> I would prefer at least keep option to restart in safe mode for LP-related issues

As I said before, if it's langpack-related, we have a known-good configuration to use with minimal interruption. Safe Mode will discard the browsing session unless manually loaded which will risk data loss.
I appreciate wanting to give maximum choice to users what to do, but this is disaster recovery -- what we should do is make sure to get back to a working state after which the user can always take further steps as-needed.

mattatobin:
Here is a question.. Why don't we simply make maxVersion be X.x.* instead of X.* and simply reissue them on major versions even if they aren't changed? Shouldn't it disable langpacks if not exactly compatible because they are always strict?

mattatobin:
From my perspective, doing some crazy insane check in padlock.js, a purely Pale Moon Application component, seems exceedingly hackish and totally the wrong approach as it will not address the core issues here. Which is why exactly langpacks aren't being updated and also seems to relate to a release engineering issue that assumes a Mozilla release scheme where none exists.

Here is a question.. Why don't we simply make maxVersion be X.x.* instead of X.* and simply reissue them on major versions even if they aren't changed? Shouldn't it disable langpacks if not exactly compatible because they are always strict?

Additionally, if the checker str8 up determines it isn't compatible won't it force an AUS check and download the update and hopefully resolve all current issues or at least disable the langpack which would cause the application to fall back to en-US regardless of prefs and not bust the browser?

If it is about the workload I am sure I could come up with some script that could accelerate the process and even write out the php array for Phoebus for easy copypasta.

mattatobin:
Assuming this is more of a problem when add-on updates are disabled or manual.. Perhaps we should alter XPInstall and the update checking routine to ignore user preferences for langpacks specifically as they are a critical aspect to application functioning.

If that assumption is incorrect then we should be looking at XPInstall and the update checking routine for a solution regardless not a grafted application scope js file.

JustOff:
> Why don't we simply make maxVersion be X.x.* instead of X.* and simply reissue them on major versions even if they aren't changed? Shouldn't it disable langpacks if not exactly compatible because they are always strict?

Good question and I think we should implement this suggestion as an addition to other measures that were taken here. @wolfbeast, what do you say?

wolfbeast:
Sounds good. We're clearly having changes often enough in string for major updates to make them less wallpaper-y

PR #1075, opened 2017-05-03
[Windows] Allow Atomic to hold 64-bit types everywhere

trav90:
Currently, we only support 32-bit Atomics types on 32-bit Windows. This PR allows us to support 64-bit types on 32-bit Windows as well.

I believe that this was the cause of the build bustage in #1017 (which used a 64-bit Atomic). Additionally, our new MSE code will be using 64-bit Atomics as well, so #1033 is dependent on this as well.

Unfortunately I have no way to test this, so am adding the `Verification Needed` label. For more in-depth information and the discussion of this at Mozilla, see [BMO 1155864](https://bugzilla.mozilla.org/show_bug.cgi?id=1155864).

### DIFF
--- a/mfbt/Atomics.h
+++ b/mfbt/Atomics.h
@@ -895,10 +895,8 @@ namespace detail {
 template<typename T, MemoryOrdering Order>
 class AtomicBase
 {
-  // We only support 32-bit types on 32-bit Windows, which constrains our
-  // implementation elsewhere.  But we support pointer-sized types everywhere.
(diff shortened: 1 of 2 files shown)


JustOff:
I confirm the successful win32 build with this PR and [Fix data race in AudioSink::mWritten](https://github.com/MoonchildProductions/Pale-Moon/pull/1017/commits/55bb01f681f9808e0aedfd39fa63f608379f5a39) re-applied.

PR #1080, opened 2017-05-06
Fix when resetting toolbars and controls in Safe Mode prompt

FranklinDM:
If you check the **Reset toolbars and controls** button and apply the changes in PM's safe mode prompt, it will only clear _localstore.rdf_, which no longer holds toolbar and window size/position settings since these were moved to _xulstore.json_.

This was also reported in the [forum](https://forum.palemoon.org/viewtopic.php?f=37&t=15658).

Please add `verification needed` label since I was only able to test this by editing omni.ja (My comp. isn't dev. class) and for me to prevent committing the mistake I did in my last PR again.

### DIFF
--- a/browser/base/content/safeMode.js
+++ b/browser/base/content/safeMode.js
@@ -81,7 +81,7 @@ function deleteLocalstore() {
   xulstoreFile.append("xulstore.json");
   try {
     xulstoreFile.remove(false);
-    if (!localstoreFile.exists()) {
+    if (localstoreFile.exists()) {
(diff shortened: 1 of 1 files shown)


janekptacijarabaci:
Should it be that way?
```
if (!localstoreFile.exists()) {
```
Maybe it should be:
```
if (localstoreFile.exists()) {
```

FranklinDM:
Fixed the typo 👍

janekptacijarabaci:
Confirmed.

PR #1081, opened 2017-05-06
Add TrackBuffersManager object

trav90:
This massive PR is a big step towards spec-compliant MSE code. Changes (and what areas are affected):

- Add abilities to MoofParser to indicate if a media segment is complete. (Regular MP4 & MSE + MP4)
- Make changes to IntervalSet and TimeUnit. (Entire media subsystem)
- Split AppendData between a synchronous part (adding to the Input Buffer) and the Processing part (MSE only)
- Make RangeRemoval return a promise rather than messing with tasks scheduled on the main thread. (MSE only)
- Add ability to retrieve init range to ContainerParser. (MSE only)
- Check MoofParser index before demuxing. (Regular MP4 & MSE + MP4)
- Remove "Diamond Problem" with MediaDecoder inheritance (see commit comment). (media subsystem)
- Add TrackBuffersManager object (+ various fixes; see individual commits). This is essentially a rewrite of our current trackbuffer/sourcebuffer code that is closer to spec. Note that this is still largly unused (will be hooked up and used to a greater extent in a future PR). (MSE only)
- Add ContainerParser::FirstCompleteMediaHeader() method. Currently, ContainerParser doesn't return true if it has a media segment, but true if at least one sample is present with a new moof. We need to be able to detect the actual media segment header (moof). (MSE + MP4)
- Do not evict data until a full init segment has been received. (MSE + MP4)
- Properly search for the required MP4 Atoms rather than make assumptions. (MSE + MP4)
- A few minor fixes to libstagefrights' ByteReader. (Regular MP4 & MSE + MP4)

I've tested this on Linux and appears to be working as intended (though more thorough testing will continue being done). Needs verification on Windows.

Tag #1033, point 7.

### DIFF
--- a/dom/media/AbstractMediaDecoder.h
+++ b/dom/media/AbstractMediaDecoder.h
@@ -41,7 +41,7 @@ enum class MediaDecoderEventVisibility : int8_t {
  * The AbstractMediaDecoder class describes the public interface for a media decoder
  * and is used by the MediaReader classes.
  */
-class AbstractMediaDecoder : public nsISupports
+class AbstractMediaDecoder : public nsIObserver
 {
 public:
   // Returns the monitor for other threads to synchronise access to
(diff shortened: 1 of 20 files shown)


wolfbeast:
I've done some very casual testing on a build with this merged in and didn't see any problems. The state seems to be very similar to what it was before from a user's perspective.

trav90:
That is the goal at this stage of the implementation. Thanks for the Windows verification!

PR #1090, opened 2017-05-09
Fix -Wreorder GCC warning

trav90:
Just what it says in the tin.


wolfbeast:
Not sure why gcc complains about it; not like one depends on the other in the initialization list...

trav90:
I'm not sure either. GCC just like to complain I guess. I noticed this warning while fixing similar warnings in some new code I'm working on, so thought I'd shut up the compiler here too while I was at it.

PR #1091, opened 2017-05-10
Add MediaSourceDemuxer

trav90:
Here's another big one! This PR introduces the MediaSourceDemuxer, which (together with the TrackBuffersManager in #1081) makes up the bulk or our new MSE code. In addition to landing that, this PR also makes several changes to our existing MSE support (which in turn will also be used by the new architecture when it is active):

- Adds the MediaSourceDemuxer object (our new MSE!) behind a pref which is disabled by default. This object uses the TrackBuffersManager to a much greater extent than we do currently.
- Adds an IsThreadSafe method because the media source demuxer can't support cloning.
- Allow MediaFormatReader to handle errors when skipping to the next keyframe.
- Don't compute the start time for MSE.
- Force the buffered range to update after reaching EOS.
- Ensure that the decoder doesn't stop decoding if there is still data available.
- Only allow seeking if we have the target data.
- Allow the MediaFormatReader to detect changes in stream content and recreate the decoder as needed.

Note that this is only a base implementation of our new MSE code. There are still many bugs and spec-compliance issues that need researched and fixed before this can start being used in place of our existing implementation. It is still very much a WIP, and should be treated as such.

As with #1081, This mostly needs testing (at this stage) to ensure that our existing implementation has not been broken. With that said, note that flipping the media.mediasource.format-reader pref _will_ cause playback regressions and (in some cases) things may not play at all. So unless you want to see all the colorful ways this doesn't quite work correctly (yet), leave it disabled!

Spent a few hours with it on Linux and appears to be working (though will continue testing). Needs verification on Windows.

### DIFF
--- a/dom/media/Intervals.h
+++ b/dom/media/Intervals.h
@@ -129,6 +129,11 @@ public:
     return mStart <= aX && aX < mEnd;
   }
 
+  bool ContainsWithStrictEnd(const T& aX) const
+  {
+    return mStart - mFuzz <= aX && aX < mEnd;
+  }
+
   bool Contains(const SelfType& aOther) const
   {
     return (mStart - mFuzz <= aOther.mStart + aOther.mFuzz) &&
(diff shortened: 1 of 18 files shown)


wolfbeast:
Preliminary testing with this built shows no regressions on windows. If anything, live streaming seems a bit more robust now.

One observation: flipping the pref, twitch halts at the exact same moment the current implementation de-syncs with a stutter. So that's a good sign that the new implementation might actually solve that problem when it's finished. Fingers crossed ;)

trav90:
> Preliminary testing with this built shows no regressions on windows. If anything, live streaming seems a bit more robust now.

:+1: 

> flipping the pref, twitch halts at the exact same moment the current implementation de-syncs with a stutter

I've seen the same behavior on Linux. I too think it's a good sign. We're (finally) getting there I think!

trav90:
Tag #1033, point 8.

PR #1099, opened 2017-05-15
Media fixes

trav90:
This PR fixes several small issues with the MediaFormatReader not correctly draining its decoder or not draining it at the correct time (which could potentially lead to stalls, dropped frames, or a/v desync in some cases). Additionally, this PR fixes an issue when we try to play a video with negative start times and also fixes a couple assertion failures.

While these fixes are mostly for our new MSE (#1033), our current code should benefit from these fixes as well. Tested and appears to be working on Linux.

### DIFF
--- a/dom/media/Intervals.h
+++ b/dom/media/Intervals.h
@@ -157,6 +157,11 @@ public:
       (aOther.mStart - aOther.mFuzz < mEnd + mFuzz);
   }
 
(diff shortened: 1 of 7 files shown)


wolfbeast:
Merged, built on windows and done basic testing. No issues seen other than the well-known ones.
Using the new code, nothing notable has changed either. So, no regressions on Windows at least :)

PR #1102, opened 2017-05-18
MSE fixes

trav90:
And here are several more MSE fixes on the road to completing #1033. First, these changes affect both our current MSE implementation as well as our new architecture:

- Ensure that the MediaFormatReader is notified when new data is received (so it can fetch new samples from the media stream).
- Be slightly less aggressive about evicting data and increase the eviction threshold to 100MB (fixes playback stalls in some situations).
- If eviction fails for some reason, throw an error.

And these changes only apply to our new code (i.e. only when the media.mediasource.format-reader pref is set to true):

- Fix an issue where new data would be inserted at the wrong position (in some circumstances) when appending data to the source buffer.
- Fix an issue where calling sourcebuffer.remove() could leave the trackbuffers in an invalid and unplayable state.
- Update the insertion index on the fly (preventing the entire stream from being reparsed).
- Various fixes related to frames being handled incorrectly.
- Various fixes related to buffered/cached data.
- At this stage, only MP4 is supported. Throw an error instead of asserting if other formats (such as WebM) are passed to the TrackBuffersManager.

I've done light testing (have been unable to test as thoroughly as usual due to intermittent lack of internet service again), and everything appears to be working as intended. I did not see any regressions with either our old or new code.

While our new code still has issues (most notable being a playback hang when changing resolutions requiring a page refresh), this PR does appear to fix the visible video distortion when on YouTube.

### DIFF
--- a/dom/media/Intervals.h
+++ b/dom/media/Intervals.h
@@ -619,6 +619,27 @@ public:
     return NoIndex;
   }
 
+  // Methods for range-based for loops.
+  typename ContainerType::iterator begin()
+  {
+    return mIntervals.begin();
+  }
+
+  typename ContainerType::const_iterator begin() const
+  {
+    return mIntervals.begin();
+  }
+
+  typename ContainerType::iterator end()
+  {
+    return mIntervals.end();
+  }
+
+  typename ContainerType::const_iterator end() const
+  {
+    return mIntervals.end();
+  }
+
 protected:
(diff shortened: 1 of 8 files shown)


wolfbeast:
Looks like we're making progress. Twitch still desyncs, but if you stop and restart the live stream, it will keep sync by the looks of it.

PR #1105, opened 2017-05-19
Canvas - CanvasRenderingContext2D.drawImage can throw 0x80040111 (NS_ERROR_NOT_AVAILABLE)

janekptacijarabaci:
__Steps to reproduce (an example)__
https://bugzilla.mozilla.org/attachment.cgi?id=549180

Actual Results:
```
[Exception... "Component is not available" nsresult: "
0x80040111 (NS_ERROR_NOT_AVAILABLE)" location: "
JS frame :: https://bug405761.bmoattachments.org/attachment.cgi?id=549180
:: window.onload :: line 7" data: no]
```

Expected Results:
```
success
```

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1297211
https://bugzilla.mozilla.org/show_bug.cgi?id=1162282

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=1328005
(I will check it continuously)

---


wolfbeast:
I'm not sure if I like the assumption that a 0x0 image is available (and can therefore manipulate hidden objects on a page). But I guess for browser parity's sake...

PR #1108, opened 2017-05-20
Fixes for new MSE architecture

trav90:
Three small MSE fixes that only apply to the new architecture:

- Fix a potential crash by preventing calls to SourceBuffer outside of the main thread.
- Fix an issue where frames could be incorrectly removed in some situations (this fix brings us closer to spec).
- Correctly handle encryption if the MSE stream has multiple init segments.

### DIFF
--- a/dom/media/MediaInfo.h
+++ b/dom/media/MediaInfo.h
@@ -287,6 +287,11 @@ public:
 
 class EncryptionInfo {
 public:
+  EncryptionInfo()
+    : mEncrypted(false)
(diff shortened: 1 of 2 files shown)


wolfbeast:
Is this encryption stuff necessary even if we don't support EME?

trav90:
I included it because we already have some encryption stuff for our current MSE architecture - (see commit 5e7416d). That and it is a small fix with a low risk of regressions.

Issue #1112, opened 2017-05-25
IndexedDB "Always Ask" option sometimes allows without asking.


janekptacijarabaci:
Ad https://forum.palemoon.org/viewtopic.php?f=3&p=114654#p114654

In my opinion it is a problem.
Even:
https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria

__Steps to reproduce__
Go to https://demo.agektmr.com/storage/
(but e.g.: http://clb.demon.fi/dump/html5_popups/?persistent - OK)
Page Info - Permissions - Maintain Offline Storage -> `Always Ask` (or `Block`)

## `Always Ask`

Pale Moon 26:
![1](https://cloud.githubusercontent.com/assets/2373486/26440308/fdb0a57e-412b-11e7-98a3-9ebdc318a5d2.png)

Pale Moon 27 (and Firefox - a last version):
`[No prompt will appear]`

## `Block`

Pale Moon 26:
![2](https://cloud.githubusercontent.com/assets/2373486/26440778/24a8347e-412e-11e7-97c7-bcd5a197818e.png)

Pale Moon 27 (and Firefox - a last version):
![3](https://cloud.githubusercontent.com/assets/2373486/26440872/84ef6fdc-412e-11e7-85e0-72bfd106fe7d.png)

__AFAIK:__

After this change...:
https://bugzilla.mozilla.org/show_bug.cgi?id=994190

...this observer:
https://github.com/MoonchildProductions/Pale-Moon/blob/27.3.0_Release/browser/base/content/browser.js#L5651
should be called for `indexedDB-permissions-prompt`

See also:
http://xref.palemoon.org/palemoon-trunk/search?string=kPermissionPromptTopic
http://xref.palemoon.org/palemoon-trunk/search?string=PromptIfNeeded

But that will not happen (if it's not `persistent`).

IMHO: The user should always decide (for `Always Ask`), no matter what type it is.

wolfbeast:
no, no, no, no.

I'm not going to break web compatibility when it was just fixed. If you want prompts, you have to use the master switch, as I explained already, in options.
![image1](https://cloud.githubusercontent.com/assets/3359132/26445977/7b052778-4142-11e7-943f-e3032b7be9c4.png)

janekptacijarabaci:
Here I am confused.

Files (e.g. in `storage\default`) are still being created (without any prompts).
[...]

PR #1113, opened 2017-05-25
Page Info - indexedDB - a remnant from ancient times (reversed value)

janekptacijarabaci:
See also https://github.com/MoonchildProductions/Pale-Moon/issues/1112 (but this problem remains)

---

### DIFF
--- a/browser/base/content/pageinfo/pageInfo.xul
+++ b/browser/base/content/pageinfo/pageInfo.xul
@@ -377,9 +377,8 @@
             <checkbox id="indexedDBDef" command="cmd_indexedDBDef" label="&permUseDefault;"/>
             <spacer flex="1"/>
             <radiogroup id="indexedDBRadioGroup" orient="horizontal">
-              <!-- Ask and Allow are purposefully reversed here! -->
-              <radio id="indexedDB#1" command="cmd_indexedDBToggle" label="&permAskAlways;"/>
-              <radio id="indexedDB#0" command="cmd_indexedDBToggle" label="&permAllow;"/>
+              <radio id="indexedDB#0" command="cmd_indexedDBToggle" label="&permAskAlways;"/>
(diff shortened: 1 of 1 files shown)


wolfbeast:
I'm confused here. what exactly are you doing?

janekptacijarabaci:
I don't see anything older:
https://github.com/MoonchildProductions/Pale-Moon/blob/24.6_RelBranch/browser/base/content/pageinfo/pageInfo.xul#L378

I don't know why it was there.

http://xref.palemoon.org/mozilla-esr24/source/browser/base/content/pageinfo/pageInfo.xul#378
vs.
https://hg.mozilla.org/releases/mozilla-esr24/annotate/tip/browser/base/content/pageinfo/pageInfo.xul#l402

IMHO: Now it's bad.

wolfbeast:
I still don't understand what you're doing here "Now it's bad" doesn't help me to understand. If the order is visually swapped, then it would not look nice in the info window.

janekptacijarabaci:
> If the order is visually swapped

But it is not, there would be a reversal of **values**:

Now:


After:

wolfbeast:
Ahh I see.

wolfbeast:
Begs the question... are you sure this corrects something wrong? because the mercurial link is only a 2-state so it was changed somewhere. Are you sure you're not breaking something here?

janekptacijarabaci:
The mercurial :-/ link is wrong (I'm sorry):
https://hg.mozilla.org/releases/mozilla-esr24/file (a v14)
https://hg.mozilla.org/releases/mozilla-esr24/file/8a82039ef232 (a last v24)

However...

Here it is added:
https://hg.mozilla.org/mozilla-central/rev/e7ccacd50d63
a here it was removed: 
https://hg.mozilla.org/mozilla-central/rev/8bcb8d073908

IMHO: It has three possible values (for a very long time).

An example:
http://clb.demon.fi/dump/html5_popups/?persistent
(there is a difference between `Always ask` and `Allow`)

PR #1114, opened 2017-05-26
[minor fix] HTML - input - dynamic maxlength didn't work

janekptacijarabaci:
An example:
https://jsfiddle.net/n44cbhgd/2/

__Steps to reproduce:__
- click b1
- input: "12" into the textbox
- click b2
- click b3
- input something into the textbox, it should be 4 char max but we can't input more than 2

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1352799

---


janekptacijarabaci:
@wolfbeast
You add the label `UXP-task`, please.

PR #1116, opened 2017-05-27
MSE performance enhancements/fixes

trav90:
This PR is a collection of performance improvements when playing MSE media. Specific changes:

- Properly release MediaResource if it is not running on the main thread (fixes a memory leak introduced into Mozilla code in early 2014!).
- Make SourceBufferResource destructor virtual (fixes memory leak).
- Remove cycle between SourceBuffer and TrackBuffersManager and use a ref-counted attribute holder to store those arguments instead (fixes memory leak and potential shutdown crash).
- Use actual allocated size rather than logical size when calculating eviction rate (should lower memory usage as a whole when playing MSE streams).

The above changes apply to both our current and new MSE architecture. The following only apply to our new code:

- Process the entire media segment at a time instead of frame by frame. This significantly reduces CPU usage (by almost 50% on my system!) when playing MSE streams with the new format reader.
- Move some debug code to behind #IF DEBUG's.
- Properly shutdown demuxers.
- Apply fuzz factor when adjusting buffered time ranges (fixes a memory leak as well as a potential crash point).

As mentioned above, using the new format reader with these patches results in nearly 50% less CPU usage and the browser uses around 200MB less RAM playing 720p YouTube videos (on my system, this makes both MSE architectures about the same performance-wise).

With our current MSE code, the CPU usage remains about the same as without these patches, but memory usage is around 75MB less (tested on YouTube and Twitch).

### DIFF
--- a/dom/media/Intervals.h
+++ b/dom/media/Intervals.h
@@ -426,6 +426,14 @@ public:
     return Intersection(intervals);
   }
 
+  SelfType& operator-= (const SelfType& aIntervals)
+  {
+    for (const auto& interval : aIntervals.mIntervals) {
+      *this -= interval;
+    }
+    return *this;
+  }
+
   SelfType operator- (const ElemType& aInterval)
   {
     SelfType intervals(*this);
--- a/dom/media/MediaData.h
+++ b/dom/media/MediaData.h
@@ -380,6 +380,10 @@ public:
   const uint8_t* Data() const { return mData; }
   // Size of buffer.
   size_t Size() const { return mSize; }
+  size_t ComputedSizeOfIncludingThis() const
+  {
+    return sizeof(*this) + mCapacity;
+  }
 
   const CryptoSample& mCrypto;
   nsRefPtr<MediaByteBuffer> mExtraData;
--- a/dom/media/MediaResource.cpp
+++ b/dom/media/MediaResource.cpp
@@ -53,20 +53,14 @@ namespace mozilla {
 void
 MediaResource::Destroy()
 {
-  // If we're being destroyed on a non-main thread, we AddRef again and
-  // use a proxy to release the MediaResource on the main thread, where
-  // the MediaResource is deleted. This ensures we only delete the
-  // MediaResource on the main thread.
-  if (!NS_IsMainThread()) {
-    nsCOMPtr<nsIThread> mainThread = do_GetMainThread();
-    NS_ENSURE_TRUE_VOID(mainThread);
-    nsRefPtr<MediaResource> doomed(this);
-    if (NS_FAILED(NS_ProxyRelease(mainThread, doomed, true))) {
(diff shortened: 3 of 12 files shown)


trav90:
Tage #1033

trav90:
Didn't mean to close. Clicked wrong button. :flushed:

wolfbeast:
Enabling the new format reader, everything seems to work; including twitch without sync issues...

wolfbeast:
Issue on shutdown after playing on several video sites: process doesn't shut down and keeps increasing memory allocation until it OOM crashes. I'll see if I can get a stack for it.

wolfbeast:
I'm published an unstable version with this merged, will have to see how things go.

wolfbeast:
Looks like the old code with the most recent changes makes twitch unwatchable (constant chokes). We may need to flip the new reader on by default even if it may not be 100% finished yet?

trav90:
> Aand.. of course I can't reproduce it with a debugger attached. Typical.
> Might be a race somewhere then, timing-critical.

Of course. I haven't experienced any crashes on Linux (yet anyway), so please do let me know if it continues.

> Looks like the old code with the most recent changes makes twitch unwatchable (constant chokes).

This is unrelated to this PR (same thing happens in -release). I just checked and it is because Twitch made yet _another_ change to their player since I was testing yesterday and they are now at backend version 1.8.5-7 according to the video stats (seriously Twitch, is it really necessary to change your player every other week? :angry: ).

trav90:
> I've published an unstable version with this merged, will have to see how things go.

:+1: 

> We may need to flip the new reader on by default even if it may not be 100% finished yet?

If we want a working Twitch we may have to. One thing to note however, is at this stage webm does not work with the new MSE yet, so enabling the new format reader effectively disables webm + MSE completely (which it's disabled by default but I know that some people still use it anyway).

wolfbeast:
I'm not entirely sure what the best choice is then. People want to use webm mostly for 4k video. We should probably keep it off even if it kills twitch for the moment.

I was able to reproduce the shutdown hang but because of Mozilla's "helpful" watchdog timer, it just crashes after x amount of time:
````
    // Shutdown is apparently dead. Crash the process.
    MOZ_CRASH("Shutdown too long, probably frozen, causing a crash.");
````
As a result there is no stack, so I have not been able to pinpoint where this shutdown hang occurs.

trav90:
`It all depends on which of the two sides you're likely to get to first ;)`

Well, so far all of my efforts have been towards getting MP4 support up to par (both with and without MSE), so at this point I'd rather finish that before moving on to webm. Even if we do enable the new format reader by default the people who have to have 4k will always have the option to disable the new format reader which (obviously) will restore previous functionality.

> Maybe this occurs when the format reader is switched off/on a few times in the same session.

I have a hunch this is the root cause, and will try to reproduce.

Issue #1120, opened 2017-05-28
Crash in ClientLayerManager::GetBackendName with Imagus 0.9.8.52 installed


wolfbeast:
Stack:
````
 	mozalloc.dll!mozalloc_abort(const char * const msg) Line 37	C++
 	xul.dll!NS_DebugBreak(unsigned int aSeverity, const char * aStr, const char * aExpr, const char * aFile, int aLine) Line 396	C++
>	xul.dll!mozilla::layers::ClientLayerManager::GetBackendName(nsAString_internal & aName) Line 767	C++
 	xul.dll!nsDOMWindowUtils::GetLayerManagerType(nsAString_internal & aType) Line 2479	C++
 	xul.dll!NS_InvokeByIndex(nsISupports * that, unsigned int methodIndex, unsigned int paramCount, nsXPTCVariant * params) Line 71	C++
 	xul.dll!XPCWrappedNative::CallMethod(XPCCallContext & ccx, XPCWrappedNative::CallMode mode) Line 1414	C++
 	xul.dll!XPC_WN_GetterSetter(JSContext * cx, unsigned int argc, JS::Value * vp) Line 1180	C++
 	mozjs.dll!js::Invoke(JSContext * cx, const JS::CallArgs & args, js::MaybeConstruct construct) Line 491	C++
````

Deliberate crash breakpoint in `ClientLayerManager::GetBackendName` for having an unknown layer manager back-end.
The extension apparently does something that makes the layer manager back-end come back with an unhandled value (possibly `LayersBackend::LAYERS_NONE`? That one is an enum value but missing from the GetBackendName case) which hits a crash because the troubleshooting page wants to get the name to enumerate accelerated/unaccelerated windows.

Extension is unfortunately minified js (why?!) so it's not possible to compare code/functionality between the last known good version (.49) and the crashing version (.52)

janekptacijarabaci:
See [bug 1218364](https://bugzilla.mozilla.org/show_bug.cgi?id=1218364)
(I don't have time to try)

Deathamns:
Imagus dev. My other extension has the same `bootstrap.js` with the same crash. [These were the changes](https://github.com/Deathamns/Viewhance/commit/f71162a4b3e069e27ec29814ff3a7110e9ddab49) after the crash started.
I've got reports for Pale Moon 26.5.0 that it can't even start the browser (but I understand that is legacy version, so I'll remove support for it).
[...]

PR #1121, opened 2017-05-28
Crash in ClientLayerManager::GetBackendName with Imagus 0.9.8.50+

janekptacijarabaci:
Ad #1120 

The problematic function is:
`appShell.createWindowlessBrowser.loadURI()` (in `bootstrap.js`)

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1218364

---

### DIFF
--- a/gfx/layers/client/ClientLayerManager.cpp
+++ b/gfx/layers/client/ClientLayerManager.cpp
@@ -750,6 +750,7 @@ void
 ClientLayerManager::GetBackendName(nsAString& aName)
 {
   switch (mForwarder->GetCompositorBackendType()) {
+    case LayersBackend::LAYERS_NONE: aName.AssignLiteral("None"); return;
     case LayersBackend::LAYERS_BASIC: aName.AssignLiteral("Basic"); return;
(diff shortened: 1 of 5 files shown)


wolfbeast:
Thanks but you said you had no time, worked on it while already assigned, and unfortunately your solution is also incomplete (there's more enum states, and anothe place in troubleshoot.jsm where this is an issue). Not everything can be ported directly -- I hope you realize that.

janekptacijarabaci:
> I hope you realize that.

Yes, the patch was imperfect, and this proposal was not accepted.

We have already talked about the flag...

wolfbeast:
... I was saying you could (and should) be more careful when porting patches to make sure the differences in our code are taken into account, including code paths we have and Firefox does not, or you'll miss paths and you'll end up potentially creating more bugs than you solve.

janekptacijarabaci:
Yes, it happened. I'm sorry.
But nobody is perfect. And will never be.
Many times before... - I had to make many adjustments (compared to the original code).

But/And a double check is always good.

PR #1122, opened 2017-05-28
Another collection of MSE fixes

trav90:
And here's another collection of (new architecture) MSE fixes:

- Ensure that progress event is fired upon successful appendBuffer (fixes playback that doesn't start in some players such as Vimeo).
- Set mediasource duration to group end timestamp (bring us closer to spec).
- Don't unnecessarily drain decoders when there's no frames pending. (also fixes some streams that won't start playing).
- Allow both first sample flags flag and sample flags simultaneously (yet another fix for streams that don't play).
- Only use init segment if known to be valid.
- Fix an issue where if a single appendBuffer is made and it contains multiple init segments; then loadedmetadata wasn't fired.
- Reduce likelihood of video distortion when seeking.
- Properly handle multiple init segments.
- Ensure all tracks are parsed (also affects non-MSE fragmented MP4).
- Activate autoplay when playing a media source (possible sec issue? See Mozilla bug 1189204).

Been testing this for a while with the new format reader and am happy to report that Twitch streams seem to play perfectly, I can change resolutions on YouTube without playback stopping and requiring a page refresh, and Vimeo HD videos now play as well! As long as this doesn't cause any adverse affects on other platforms this is a major hurdle that can be put behind us!

Tag #1033.

### DIFF
--- a/dom/html/HTMLMediaElement.cpp
+++ b/dom/html/HTMLMediaElement.cpp
@@ -3686,12 +3686,14 @@ bool HTMLMediaElement::CanActivateAutoplay()
 {
   // For stream inputs, we activate autoplay on HAVE_NOTHING because
   // this element itself might be blocking the stream from making progress by
-  // being paused.
+  // being paused. We also activate autoplay when playing a media source since
+  // the data download is controlled by the script and there is no way to
+  // evaluate MediaDecoder::CanPlayThrough().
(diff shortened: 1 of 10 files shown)


wolfbeast:
Eagerly merging this in to build and test! :confetti_ball:

PR #1124, opened 2017-05-29
Crash/hang fixes

trav90:
This PR is a collection of crash and hang fixes for our media code (both MSE and regular media). I've been testing these patches on YouTube, Twitch, and Vimeo and have not seen any hangs or cashes (hopefully this resolves our shutdown crash that's been reported). Changes:

- Clear mDecoder in HTMLMediaElement::NetworkError() to prevent it from dispatching a task to a task queue that has already been shut down (potential shutdown hang or crash).
- Change exception code when MediaSource.addSourceBuffer() fails due to user disabling MP4 (Mozilla bug 1160441).
- Remove potentially racey (and unnecessary) code from new MSE.
- Ensure buffered index is up to date for MP4 (Mozilla bug 1180403).
- Fix up and add a bit more logging to the MediaFormatReader.
- Ensure that the MediaFormatReader correctly sets the drainComplete flag on error (fixes potential shutdown hang).

Tag #1033.

### DIFF
--- a/dom/html/HTMLMediaElement.cpp
+++ b/dom/html/HTMLMediaElement.cpp
@@ -3226,6 +3226,9 @@ void HTMLMediaElement::FirstFrameLoaded()
 
 void HTMLMediaElement::NetworkError()
 {
+  if (mDecoder) {
+    ShutdownDecoder();
+  }
   Error(nsIDOMMediaError::MEDIA_ERR_NETWORK);
 }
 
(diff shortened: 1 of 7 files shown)


wolfbeast:
I think especially the point:

> Ensure that the MediaFormatReader correctly sets the drainComplete flag on error (fixes potential shutdown hang).

might be relevant, because I did notice a ton of JS errors being thrown by the player if shutting down while playing and the hang occurring.
Merging, building and praying ;)

trav90:
> I think especially the point:

Agreed.

> I did notice a ton of JS errors being thrown by the player if shutting down while playing and the hang occurring.

Which player was this? I've still had trouble reproducing on Linux, so knowing which player was in use when it happened on Windows might help (assuming that this PR doesn't fix it of course).

wolfbeast:
I think it was the one on Twitch, but not sure.

PR #1125, opened 2017-05-31
[minor fix] HTML - select - changing the value of a select to a value that matches none of the options

janekptacijarabaci:
__Steps to reproduce__

Go to: http://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_select_value
Change this line:
`document.getElementById("mySelect").value = "banana";`
with this one:
`document.getElementById("mySelect").value = "bananas";`

Actual results:
The selected value still the first value. 

Expected results:
This option should be blank.

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1203668

---


wolfbeast:
:roll_eyes: You'd expect people in charge of scripting to do this right XD

PR #1129, opened 2017-05-31
[suggestion] JS - Date - parse - new Date(<mm/dd/yy>) behaves differently in some browsers

janekptacijarabaci:
__An example:__
https://bugzilla.mozilla.org/attachment.cgi?id=8742016

__Actual results (just like it is in: MSIE 11, EDGE):__
Mon Apr 16 1917 00:00:00 GMT+0200

__Expected results (just like it is in: Firefox, Safari, Chrome):__
Mon Apr 16 2017 00:00:00 GMT+0200
(IMHO: it makes sense)

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1265136

---


mattatobin:
Can you please explain the code changes and the cutoffs for the 1900/2000 for two digit dates?

What happens if i do the following years:
11/21/98
10/28/89
5/2/72
8/2/64
11/5/55
and so on.. When does it assume 19XX and when does it assume 20XX?

janekptacijarabaci:
You see https://github.com/MoonchildProductions/Pale-Moon/pull/1129/files#diff-dabbeb02889d1c69b5fa2cfe8ab4ec02R1092:

```
"4/16/17: Sun Apr 16 2017 00:00:00 GMT+0200"
"11/21/98: Sat Nov 21 1998 00:00:00 GMT+0100"
"10/28/89: Sat Oct 28 1989 00:00:00 GMT+0200"
"5/2/72: Tue May 02 1972 00:00:00 GMT+0200"
"8/2/64: Sun Aug 02 1964 00:00:00 GMT+0200"
"11/5/55: Sat Nov 05 1955 00:00:00 GMT+0100"
"12/31/49: Fri Dec 31 2049 00:00:00 GMT+0100"
"1/1/50: Sun Jan 01 1950 00:00:00 GMT+0100"
```

mattatobin:
One last question.. Does the code use the current year as a basis to decide between 19XX and 20XX?

wolfbeast:
@mattatobin the cutoff year is 50.
````C++
+    if (year < 50)
+        year += 2000;
+    else if (year >= 50 && year < 100)
+        year += 1900;
````
so 
0-49 => 2000-2049
50-99 => 1950-1999
100+ => as-is (full year AD)

PR #1144, opened 2017-06-04
[minor fix, suggestion] Pageinfo - General - the site's security information (regression)

janekptacijarabaci:
vs.

or

See also Pale Moon 26.5.0 (and SeaMonkey).

---

If the proposal is accepted:

__You add the label `Regression`, please.__

__You add the label `UXP-task`, please.__
(or is it not needed for `/browser`?)

---


wolfbeast:
UXP-task is only for platform issues, not for front-end, so no, anything in /browser will not need the UXP-task label.

mattatobin:
Unless it is both platform and browser..

wolfbeast:
> anything in /browser will not need the UXP-task label.

If it's got code changes that are not in `/browser`, then that doesn't apply, now does it?

Issue #1147, opened 2017-06-06
feature request - filterable network monitor output


leedoyle:
I think many web developers and power users would enjoy being able to filter the output of the network monitor in a more convenient way. Afaik, to implement this feature one would only have to port the code from Mozilla FF.

Attached is my suggestion (arrow **2**)
![devtoolsdream](https://cloud.githubusercontent.com/assets/25978190/26825429/09a89ec4-4ac6-11e7-9e53-9a309bd54365.png)

janekptacijarabaci:
The `Clear` button is at the bottom right.

leedoyle:
Thank you, I've amended the issue.

PR #1148, opened 2017-06-06
DevTools - Network - Filter URLs (+ a better URLs in File column)

janekptacijarabaci:
Ad #1147 

## 1)

Added `Filter URLs`

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=892229
(without a shortcut key - IMHO it is not a good idea)

## 2)

A better URLs in `File` column:

vs.

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1162677

---

__You add the label `Devtools`, please.__
__You add the label `String changes`, please.__

---

### DIFF
--- a/toolkit/devtools/netmonitor/netmonitor-view.js
+++ b/toolkit/devtools/netmonitor/netmonitor-view.js
@@ -59,6 +59,7 @@ const GENERIC_VARIABLES_VIEW_SETTINGS = {
   eval: () => {}
 };
 const NETWORK_ANALYSIS_PIE_CHART_DIAMETER = 200; // px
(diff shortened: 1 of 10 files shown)


leedoyle:
Thanks for your commit. However, I'm not sure all users will find it more convenient than the current network monitor interface, which, even though more complicated, can display both filtered and unfiltered output at the same time. 
The use case where one would need them both seems to be very specific, though, so your implementation may be a reasonable choice for most users.

Lootyhoof:
If merged, this should also use the label `Theme changes`.

wolfbeast:
Looks like a very clean patch and something users have asked for, so merging this in.

PR #1158, opened 2017-06-09
Import login data from Google Chrome on Windows

JustOff:
Based on bugs [707044](https://bugzilla.mozilla.org/show_bug.cgi?id=707044) and [682069](https://bugzilla.mozilla.org/show_bug.cgi?id=682069).

Import passwords from Chrome `Login Data` file. Currently it requires that Chrome be closed during import. Lock-independent migration introduced in [1285041](https://bugzilla.mozilla.org/show_bug.cgi?id=1285041) needs more investigation before implement.

### DIFF
--- a/browser/components/migration/ChromeProfileMigrator.js
+++ b/browser/components/migration/ChromeProfileMigrator.js
@@ -364,14 +364,15 @@ function GetWindowsPasswordsResource(aProfileFolder) {
         password_element, password_value, signon_realm, scheme, date_created,
         times_used FROM logins WHERE blacklisted_by_user = 0`);
       let crypto = new OSCrypto();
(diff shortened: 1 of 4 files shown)


browser/components/migration/ChromeProfileMigrator.js
```diff
@@ -6,16 +6,19 @@
 
 "use strict";
 
-const Cc = Components.classes;
-const Ci = Components.interfaces;
-const Cu = Components.utils;
-const Cr = Components.results;
+const { classes: Cc, interfaces: Ci, results: Cr, utils: Cu } = Components;
```
wolfbeast: Please don't use this. It's much more difficult to read with its backwards assignment.

JustOff: Ok, fixed. I'll try to remember this for the future.

Issue #1157, opened 2017-06-09
DevTools Inspector tooltip get element width/height wrong when browser zoomed


zcyzcy88:
Firefox isomorphic: https://bugzilla.mozilla.org/show_bug.cgi?id=1341580

I don't talk about image tooltip(white), I mean the element tooltip(black).

### Actual results
The value showed on the tooltip is considered DPI, but not considered zoom rate.
When I zoom in, the value will bigger.

<img width="1280" alt="2017-06-09 19_58_35-greenshot" src="https://user-images.githubusercontent.com/15178410/26974512-4172ac02-4d4e-11e7-843f-2e964f07786b.png">
<img width="1280" alt="2017-06-09 19_58_57-greenshot" src="https://user-images.githubusercontent.com/15178410/26974511-41715f50-4d4e-11e7-93f5-2c375525d539.png">

### Expected results:
Always same as `getComputedStyle()`

wolfbeast:
Can you please verify in the current unstable that it's fixed?

janekptacijarabaci:
@wolfbeast 
Ad #1154 and #1160
I'm sorry, but these changes are not included in this update (last unstable version).
You see also this file: omni.ja in the Pale Moon folder.

wolfbeast:
@janekptacijarabaci They should be (build of 2017-06-13)

wolfbeast:
Thanks for the crickets. I've verified this myself as working.

PR #1160, opened 2017-06-09
DevTools - Inspector - tooltip get element width/height wrong when browser zoomed

janekptacijarabaci:
Ad #1157 

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1341580

---

### DIFF
--- a/toolkit/devtools/server/actors/highlighter.js
+++ b/toolkit/devtools/server/actors/highlighter.js
@@ -1369,8 +1369,14 @@ BoxModelHighlighter.prototype = Heritage.extend(AutoRefreshHighlighter.prototype
       pseudos += ":" + pseudo;
     }
 
-    let rect = this._getOuterQuad("border").bounds;
-    let dim = Math.ceil(rect.width) + " \u00D7 " + Math.ceil(rect.height);
+    // We want to display the original `width` and `height`, instead of the ones affected
+    // by any zoom. Since the infobar can be displayed also for text nodes, we can't
+    // access the computed style for that, and this is why we recalculate them here.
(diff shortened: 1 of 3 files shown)


toolkit/devtools/server/actors/highlighter.js
```diff
@@ -1369,8 +1369,14 @@ BoxModelHighlighter.prototype = Heritage.extend(AutoRefreshHighlighter.prototype
       pseudos += ":" + pseudo;
     }
 
-    let rect = this._getOuterQuad("border").bounds;
-    let dim = Math.ceil(rect.width) + " \u00D7 " + Math.ceil(rect.height);
+    // We want to display the original `width` and `height`, instead of the ones affected
+    // by any zoom. Since the infobar can be displayed also for text nodes, we can't
+    // access the computed style for that, and this is why we recalculate them here.
+    let zoom = LayoutHelpers.getCurrentZoom(node);
```
wolfbeast: I'm wondering: is this correct? You're not looking at the zoom value for the actual window here which is the issue.

janekptacijarabaci: I don't know a specific example - in which case it would be bad.
You see also https://github.com/MoonchildProductions/Pale-Moon/pull/1162

wolfbeast: In the end I guess it doesn't matter since node zoom will be equal to window zoom, so.. it's fine :)

janekptacijarabaci:
@wolfbeast
You add the label `UXP-task`, please.

PR #1171, opened 2017-06-17
Remove cycle between TrackBuffersManager and MediaSourceDemuxer

trav90:
This fixes a potential never-ending cycle between TrackBuffersManage and the MediaSourceDemuxer that prevents the appropriate destructor from being called. This particularly happens when multiple streams have been played in a single session. I believe this is the cause of the shutdown hang we've seen with the new MSE code.

I've tested using my normal steps to reproduce the shutdown hang and so far have not experienced it.


trav90:
Tag #1033.

wolfbeast:
Sometimes it's those on-liners that hold you up a long time...
I'll give it some good thrashing myself, see if that hangs are all solved.

trav90:
I've not experienced a single shutdown hang on Linux over the past 3 days. Have any hangs happened on Windows?

wolfbeast:
No further hangs experienced, we can consider this good and well resolved.

trav90:
:tada:

PR #1173, opened 2017-06-17
Fix unsafe "instanceof" negations

janekptacijarabaci:
See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1322954

An example (Scratchpad):
```
var obj = {
  "a": 1,
};
if (!"b" in obj) {
  alert("good");
} else {
  alert("bad");
}
```

---

I've created the new build (x32, Windows).

### DIFF
--- a/mobile/android/chrome/content/InputWidgetHelper.js
+++ b/mobile/android/chrome/content/InputWidgetHelper.js
@@ -61,7 +61,7 @@ var InputWidgetHelper = {
   },
 
   hasInputWidget: function(aElement) {
-    if (!aElement instanceof HTMLInputElement)
+    if (!(aElement instanceof HTMLInputElement))
       return false;
 
     let type = aElement.getAttribute('type');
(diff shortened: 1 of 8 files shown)


wolfbeast:
I'm sorry but your logic seems to be wrong.
`!"b"` is false. `false` is not part of `obj` therefore should be false. Basic ECMAScript logic.
If you want this to evaluate as I think you do, then you need `(!("b" in obj))`

adisib:
That is what he is saying, and why the commit is adding parens. It isn't a PR to change behavior, it is to make the code not be like the example.

wolfbeast:
Right, right. Not that that was clear from the PR description which seemed to be a change in behavior (which is an immediate no so I didn't look at the code change).

Lesson for the future: please be **VERY DESCRIPTIVE** when you offer a PR with no attached issue out of the blue.

janekptacijarabaci:
I'm sorry. Next time I copy it (from the first comment).

wolfbeast:
No, just make a clear description of what a PR does/is for/fixes/etc.
Copying comment 0 from that bug would really not have helped in this case, because it only describes a situation (and it would still not be clear what this PR would be doing).

janekptacijarabaci:
Well, in that case I don't know how to describe it.

PR #1180, opened 2017-06-19
Supports building with Clang and Libc++

alaviss:
I have ported some patches from gecko-dev to allow Pale Moon to be built with clang/libc++

Tested with clang 5.0.0

### DIFF
--- a/memory/mozalloc/throw_gcc.h
+++ b/memory/mozalloc/throw_gcc.h
@@ -25,91 +25,91 @@ namespace std {
 // doing this after careful review because we want to define our own
 // exception throwing semantics.  Don't try this at home!
 
-MOZ_NORETURN MOZ_ALWAYS_INLINE void
+MOZALLOC_EXPORT MOZ_NORETURN MOZ_ALWAYS_INLINE void
 __throw_bad_exception(void)
 {
     mozalloc_abort("fatal: STL threw bad_exception");
 }
 
-MOZ_NORETURN MOZ_ALWAYS_INLINE void
+MOZALLOC_EXPORT MOZ_NORETURN MOZ_ALWAYS_INLINE void
 __throw_bad_alloc(void)
 {
     mozalloc_abort("fatal: STL threw bad_alloc");
 }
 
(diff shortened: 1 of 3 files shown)


mattatobin:
And what are the effects of these changes to gcc and msvc?

alaviss:
I'm currently building with gcc, but I would need help in testing with msvc, as I don't use Windows

JustOff:
I just made win32 build with this changes, it seems to run as usual.

mattatobin:
Don't underestimate any change to jemalloc or to xpcom even the smallest thing can be disastrous even if it isn't evident at a cursory glance.

alaviss:
Builds and runs fine with gcc-4.9. I'm typing this from that build right now

Bfgeshka:
Tried current master with this patchset and build (still) fails on clang 4.0.

alaviss:
Can I have the build log?

Bfgeshka:
Here you go: http://paste.org.ru/?2p2q0f

Btw, build for gcc 6.3.0 now failed too. Build log would be in another time, but I can say that it happened in different place.

alaviss:
Try to build with USE="-shared-js"‎

Bfgeshka:
So far build works only for gcc 4 and 5 (not concidering clang 5 - it is so, so far in the future), which is bad because these patches are breaking more compability than bringing in. It needs to be reworked.

> Try to build with USE="-shared-js"‎

You cannot say that. Build must be able to finish successfully without dependency on specific config option set.

wolfbeast:
I'll be backing out this patch. Sounds like you will need to make this conditional code based on which compiler and which version is in use (using preprocessor macros). You cannot simply break gcc builds (official) to satisfy clang (new support).

alaviss:
I've pulled gcc 6.3.1 to test this, there's absolutely no build failures for gcc 6. I would be interested in build log showing build failure for gcc 6

mattatobin:
Just ifdef it with else being the previous code. It isn't complicated.

PR #1182, opened 2017-06-21
Copy as cURL - Fix unsupported curl option

adeshkp:
This is fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1253487

### DIFF
--- a/toolkit/devtools/shared/Curl.jsm
+++ b/toolkit/devtools/shared/Curl.jsm
@@ -110,8 +110,12 @@ this.Curl = {
     // Add http version.
     if (aData.httpVersion && aData.httpVersion != DEFAULT_HTTP_VERSION) {
       let version = aData.httpVersion.split("/")[1];
-      let major = version.split(".")[0];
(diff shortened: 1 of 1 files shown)


toolkit/devtools/shared/Curl.jsm
```diff
@@ -109,7 +109,9 @@ this.Curl = {
 
     // Add http version.
     if (aData.httpVersion && aData.httpVersion != DEFAULT_HTTP_VERSION) {
-      command.push("--" + aData.httpVersion.split("/")[1]);
+      let version = aData.httpVersion.split("/")[1];
+      let major = version.split(".")[0];
+      command.push("--http" + (major == "2" ? major : version));
```
wolfbeast: Please add a comment about what cURL accepts and how the logic satisfies it (full version vs. major only).
Also, what about HTTP/0.9?

adeshkp: I couldn't find anything related to curl support for HTTP/0.9. Also curl does not have an option for accepting HTTP/0.9 only requests.

PR #1184, opened 2017-06-22
Updates for MSE spec compliance

trav90:
A few more minor changes that makes our new MSE code more spec compliant:

- Maintain coded group continuity in sequence append mode when resetting parser state per spec.
- Restrict leeway to append windows start (brings us closer to spec).
- Don't allow call to appendBuffer if media element is in error per spec.

Tested on usual media sites.


wolfbeast:
Built and tested on windows with YT, Vimeo and Twitch, no issues seen.

trav90:
Tag #1033.

PR #1191, opened 2017-06-23
Change print_to_filename

adeshkp:
When printing to file on Linux systems, filename is filled with "mozilla.pdf".
This PR changes it to "palemoon.pdf"

### DIFF
--- a/widget/gtk/nsDeviceContextSpecG.cpp
+++ b/widget/gtk/nsDeviceContextSpecG.cpp
@@ -482,9 +482,9 @@ NS_IMETHODIMP nsPrinterEnumeratorGTK::InitPrintSettingsFromPrinter(const char16_
       path = PR_GetEnv("HOME");
   
     if (path)
-      filename = nsPrintfCString("%s/palemoon.pdf", path);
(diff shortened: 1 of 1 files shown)


mattatobin:
Ideally this should not use a filename that is Pale Moon dependent. Can you think of an appropriate generic name for this?

adeshkp:
I gave it some more thought and now I feel that filename should be _<page_title>.pdf_. But doing that may require significant changes and is something we might decide do later (or not). Currently filename value is being filled with `print.print_to_filename` which is set to "mozilla.pdf" on build time. If a user changes the filename while printing, subsequent print dialogs will be filled with this modified value which I, as a user, would not like at all because it means that I'll have change this every time based on the page I'm printing.

For now, I'm fine with "goanna.pdf". What about "webpage.pdf"? Too generic?

wolfbeast:
`goanna.pdf` sounds good to me. I want to avoid `<page title>.pdf` or similar content-controlled names because you may end up with an attack surface then (e.g. page title is `~/.initrc,` or somesuch, overwriting data files)

PR #1199, opened 2017-06-29
Build fixes for Linux/m68k

glaubitz:
Hi!

This pull request contains a couple of patches necessary to fix the build of PaleMoon on Linux/m68k.

Those patches are based on my patch set for Firefox upstream [1], I just backported them to the PaleMoon codebase.

All of these patches have already been acknowledged by Mozilla upstream [2]. However, they haven't been merged yet because Mozilla upstream does not agree with my approach to enforce alignment of some structs as required for Linux/m68k. I haven't included these alignment fixes in this pull request yet anyway.

I have more fixes for other patches lined up which I am happy to contribute to PaleMoon as well!

Thanks for consideration! 

Adrian

> [1] https://github.com/glaubitz/gecko-dev/commits/m68k
> [2] https://bugzilla.mozilla.org/show_bug.cgi?id=1325771

### DIFF
--- a/ipc/chromium/src/build/build_config.h
+++ b/ipc/chromium/src/build/build_config.h
@@ -81,6 +81,9 @@
 #elif defined(__ppc__) || defined(__powerpc__)
 #define ARCH_CPU_PPC 1
 #define ARCH_CPU_32_BITS 1
+#elif defined(__m68k__)
+#define ARCH_CPU_M68K 1
+#define ARCH_CPU_32_BITS 1
 #elif defined(__sparc64__)
 #define ARCH_CPU_SPARC 1
 #define ARCH_CPU_64_BITS 1
--- a/media/webrtc/trunk/build/build_config.h
+++ b/media/webrtc/trunk/build/build_config.h
@@ -129,6 +129,11 @@
 #define ARCH_CPU_PPC 1
 #define ARCH_CPU_32_BITS 1
 #define ARCH_CPU_BIG_ENDIAN 1
+#elif defined(__m68k__)
+#define ARCH_CPU_M68K_FAMILY 1
+#define ARCH_CPU_M68K 1
+#define ARCH_CPU_32_BITS 1
+#define ARCH_CPU_BIG_ENDIAN 1
 #elif defined(__sparc64__)
 #define ARCH_CPU_SPARC_FAMILY 1
 #define ARCH_CPU_SPARC 1
--- a/mfbt/tests/TestPair.cpp
+++ b/mfbt/tests/TestPair.cpp
@@ -29,14 +29,19 @@ using mozilla::Pair;
   static_assert(sizeof(name##_2) == (size), \
                 "Pair<" #T2 ", " #T1 "> has an unexpected size");
 
+static constexpr size_t sizemax(size_t a, size_t b)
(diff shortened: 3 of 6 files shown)


wolfbeast:
Looks like the alignment isn't the issue, but rather the fact that they want to (need to) have a Rust m68k compiler back-end to use, and won't merge unless it has a future in their current trunk builds (ff57+)... Which likely nobody at Mozilla is going to be writing, in all honesty.

I'm not merging this in right now because it touches much core code and I'd like to get into the release stage for 27.4 first before tackling integration of code for a new architecture that touches other gcc-dependent code.

glaubitz:
> Looks like the alignment isn't the issue, but rather the fact that they want to (need to) have a Rust m68k compiler back-end to use, and won't merge unless it has a future in their current trunk builds (ff57+)... Which likely nobody at Mozilla is going to be writing, in all honesty.

Nah, that wasn't really the issue back then. They were just bike-shedding about the best solution for the alignment stuff. If you follow the discussion - which is very long because the discussion got derailed a bit - you'll see that they didn't want to use "__attribute__(((aligned(N)))" but rather C++ alignas(). I tried to use the the latter but it fails if the alignment chosen is smaller than the native alignment which is not the case with the attribute variant. They merged my other patches for sparc64, SuperH and Alpha and these platforms don't have Rust support yet either.

> I'm not merging this in right now because it touches much core code and I'd like to get into the release stage for 27.4 first before tackling integration of code for a new architecture that touches other gcc-dependent code.

All patches except the last one add code code which is #ifdef'd for __m68k__, so there isn't really anything to worry about. The last one by @karcherm just makes one part of the test suite more flexible regarding alignment and this patch has been acked by Mozilla as is.

But I'm fine with waiting, I might send some additional pull requests in the mean time.

wolfbeast:
I skimmed over the discussion because there was a lot of political rope-pulling going on, and I didn't look at the code changes in detail yet; sorry if I misunderstood the actual issue.
If all code is properly `#ifdef`ed then there's not going to be much of an issue, agreed. I'm still going to be careful here and hold off for now to get our focused development for the next release that's needed. I'll get back to this afterwards :)

mattatobin:
I am wondering why we would want m68k support in the first place.. What modern systems and devices could possibly be using a 68k anyway.. You planning to backport to the Sega Genesis or something?

Also do note that we will be leaving this codebase behind eventually once the Unified XUL Platform codebase is up to snuff and ready for a port of the Pale Moon application.

Know it has never been the intention of Pale Moon or the Platform code to cater or even support ancient hardware or operating systems. While I appreciate the enthusiasm of your work.. I would really like some insight as to why this is or would be important and beneficial to us now and in the future.

glaubitz:
There are already FPGA-based 68k accelerators for Amiga and Atari out there with 600 MHz and more.

68000 has all the important features for modern software like compare-and-swap in hardware and compiler support is extremely good (full C++11 support, for example). There is a large community around the CPU which constantly work on improving both hard- and software support.

Other interesting architectures are Hitachi's SuperH which is currently being redesigned as an open source CPU called "J-Core" (see www.j-core.org).

So, a processor design being old doesn't automatically mean it's not being developed and used anymore.

PR #1202, opened 2017-06-30
Restore missing notification icons

FranklinDM:
This pull request restores the missing notification icons in the notification bar (appears on top of the status bar) by changing the selectors for the notification icons. Tested by editing omni.ja, please add the `verification needed` and `theme/ui` tags.

I have noticed in a lot of themes (including the default theme) that when some events happen such as I was unable to sync for a few days, the icon beside the notification text is missing. I compared the "notification.css" file to one of my themes and found that the selectors for the icons were different.

### DIFF
--- a/toolkit/themes/linux/global/notification.css
+++ b/toolkit/themes/linux/global/notification.css
@@ -44,15 +44,15 @@ notification[type="critical"] {
 
 /* Default icons for notifications */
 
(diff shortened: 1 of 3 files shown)


wolfbeast:
Not entirely sure when this regressed. Since this is all part of the toolkit, it should have ported across cleanly with Tycho, but may not have.

PR #1204, opened 2017-06-30
[minor fix] Some sites - throws an errors (aTab is null)

janekptacijarabaci:
Some sites - throws an errors in Browser Console:
```
TypeError: aTab is null
tabbrowser.xml:2291:10
```

See:
https://hg.mozilla.org/mozilla-central/rev/103ffeaa554d

See also:
#745 

__I can give some example later today... Do you need it, please?__

---


wolfbeast:
No example needed. basic null check.

PR #1205, opened 2017-06-30
Move some WebRTC files out of /dom/media

trav90:
Just what it says in the tin in preparation for my work this weekend. Also disables more WebRTC code we don't need (and previously were compiling anyway) unless --enable-webrtc is defined in mozconfig.

Tested with normal build config and no issues seen. Also made sure an enable-webrtc build compiles successfully, though I did not test any WebRTC functionality.


wolfbeast:
I haven't been able to test this on MSVC yet because we have an older build bustage issue, apparently; see #1206 for details.

wolfbeast:
Builds fine with our default config, no issues seen.

wolfbeast:
Fixed the build issues with webrtc, but functional testing is likely not going to happen on our code base because WebRTC is still prefixed with -moz that nobody really uses. Good enough for now though.

PR #1209, opened 2017-07-03
Make more of our media code use State machinery

trav90:
And here it is! The PR that should complete #1033!

TL;DR this PR hooks up several additional areas of our media code to the State Machinery (which was added in #1028). This resolves some race conditions in the media subsystem and allows us to do some cleanup and simplify portions of the media code. In my testing thus far this fixes both the problem with seeking on Vimeo and some YouTube live streams not starting.

More detailed specifics:

- Use StateMirroring for playback position.
- Use StateWatching for mPlayState in the MDSM.
- Remove ApplyToStateMachine and PLAY_STATE_SEEKING (no longer needed thanks to above point).
- Update TimeUnits to work properly with StateMirroring.
- Use StateMirroring for metadata duration, demuxer-estimated duration, actual observed duration, and MSE duration.
- Remove no longer used codepath for off-main-thread MSE duration change.
- Make MediaDecoder:DurationChanged work asynchronously.
- Cleanup SetDuration code.
- Drop support for Content-Duration. See the commit message for more detail, but essentially this simplifies the computation of media duration.

I've tested this pretty extensively on Linux, playing many YouTube videos, a dozen or so YT Live streams, several streams on Twitch and Vimeo, and several on various other sites around the web. So far I've yet to see any problems, and as mentioned above seeking worked on all the Vimeo vids I tried it on, and I've yet to have a YouTube live stream not start automatically.

Most of my testing has been with the default unstable config (MSE + MP4 using new format reader), but I've also done a bit of testing with the older MSE reader and MSE disabled completely.


wolfbeast:
Fingers crossed this is a full cross-platform fix (it should...!)

I'll build this and take today testing it, and if good, trying to break it ;-)
Even if there are some leftover issues that aren't breaking, it can be done in follow-up, since I really do need to get 27.4 out the door due to sec updates in it.

PR #1220, opened 2017-07-07
Do not apply add-on compatibility update in strict mode.

JustOff:
Do not allow to override add-on's `minVersion` & `maxVersion` from the update server data in the strict compatibility mode (i.e. when `addon.type` is not in `{extension, dictionary}` or `extensions.strictCompatibility == true`).

See also [issue at phoebus](https://github.com/Pale-Moon-Addons-Team/phoebus/issues/57#issuecomment-313670445).


wolfbeast:
This prevents a served compatibility update from being applied (if version is outside of the target range in the add-on itself) when strict compatibility is enforced, either by the extension's flags or because of the type of add-on (language pack, theme, etc.)

wolfbeast:
Uplifted to release in 0463b796f67ca55db8767c07df75fcfea30796ee

PR #1221, opened 2017-07-08
Use CSS variables in browser.css on Windows

Lootyhoof:
This introduces the use of CSS variables in `browser.css` on Windows. For the most part this involves converting several of the pre-processed variables into CSS-compatible ones, so that can we manipulate them on-the-fly. This is highly desired for us, so that we can work with different Windows versions (mainly, Windows 10) without having to style each selector individually for each change that needs to be made (as a notable example, see the section that used to be littered with `@customToolbarColor@` and `@customToolbarColorWin10@` statements). This can also be used to provide alternative styling when in `-moz-lwtheme` mode.

As a side-effect of these changes, this also provides enhanced customization for users. While the variables introduced are not optimized for this feature at present, I have added a few variables to demonstrate the technology and complement variables that we are using for their intended purpose. As a particularly wild example, with the current variables in this patch we would be able to produce the following using the styling featured within:

CSS Variable example

This allows for some basic manipulation of the UI without having to dig deep to find the correct selectors, as the variables do all the hard work for those who want to use CSS to change the theme styling. With that said, though, I'm not sure if this feature should be advertised currently - unless we want to focus on the customization aspects (in which case I can add more variables), it's not a complete solution, and only allows for very basic changes, while assuming users actually know the types of values to input for each variable.

I've also included some small, additional changes into this, also related to variables (but changing slightly how elements are displayed), pertaining mainly to text styling on transparent windows. In particular, `.toolbarbutton-text` (i.e. Icons & Text mode) on `#nav-bar` now has a `text-shadow` on Windows Vista/7 to increase legibility. On Windows 8/10, this now turns white on dark window frames, while on unfocused windows all text in the transparent window area becomes `ThreeDShadow` (as opposed to just the menubar). Additionally, as a result of the introduction of variables for the toolbar highlights, items in `#TabsToolbar` now take on a dark motif when in `-moz-lwtheme-brighttext` mode, like the rest of the UI, when hovered (as opposed to the default white motif before, which is still used in normal mode and `-moz-lwtheme-darktext`).

Please note that, as the title suggests, this is for Windows only at present, as I could not find a valid purpose for these on other platforms with their current purpose. If we want to include more of these variables for increased customization for users (and possibly extension developers too), I could certainly look to adding some for Linux and macOS too, as well as adding more to the current selection on Windows.

I have tested these changes on Windows 7/8/10, and no issues were found. Other than the small changes mentioned above, there should be no noticeable (to the user) changes to the theme.

Please add the labels `Theme/UI` and `Code Cleanup`. This will not affect third-party themes, though theme authors may benefit greatly from including these (or similar) depending on how their CSS is styled.

### DIFF
--- a/browser/themes/windows/browser.css
+++ b/browser/themes/windows/browser.css
@@ -9,20 +9,14 @@
 
 %include ../shared/browser.inc
 %filter substitution
-%define toolbarHighlight rgba(255,255,255,.5)
-%define selectedTabHighlight rgba(255,255,255,.7)
 %define toolbarShadowColor rgba(10%,10%,10%,.4)
 %define toolbarShadowOnTab linear-gradient(to top, rgba(10%,10%,10%,.4) 1px, transparent 1px)
-%define bgTabTexture linear-gradient(transparent, hsla(0,0%,45%,.1) 1px, hsla(0,0%,32%,.2) 80%, hsla(0,0%,0%,.2))
-%define bgTabTextureHover linear-gradient(hsla(0,0%,100%,.3) 1px, hsla(0,0%,75%,.2) 80%, hsla(0,0%,60%,.2))
 %define navbarTextboxCustomBorder border-color: rgba(0,0,0,.32);
 %define navbarLargeIcons #navigator-toolbox[iconsize=large][mode=icons] > #nav-bar
 %define forwardTransitionLength 150ms
 %define conditionalForwardWithUrlbar window:not([chromehidden~=toolbar]) #navigator-toolbox[iconsize=large][mode=icons] > :-moz-any(#nav-bar[currentset*="unified-back-forward-button"],#nav-bar:not([currentset])) > #unified-back-forward-button
 %define conditionalForwardWithUrlbar2 window:not([chromehidden~=toolbar]) #navigator-toolbox[iconsize=large][mode=icons] > :-moz-any(#nav-bar[currentset*="unified-back-forward-button"]:not([currentset*="unified-back-forward-button,urlbar-container"]),#nav-bar:not([currentset])) > #unified-back-forward-button
 %define conditionalForwardWithUrlbarWidth 27
-%define customToolbarColor hsl(210,75%,92%)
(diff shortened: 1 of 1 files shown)


wolfbeast:
I'm sure this will make themeists happy ;)

wolfbeast:
Noticed a warning after this commit. Not sure why, since it's properly defined, but maybe you can see what's up.

07:37:00.369 Property contained reference to invalid variable.  Error in parsing value for 'color'.  Falling back to 'inherit'.1 browser.css:3996:124275

offending code
```css
  #toolbar-menubar:not(:-moz-lwtheme),
  #TabsToolbar[tabsontop=true]:not(:-moz-lwtheme),
  #nav-bar[tabsontop=false]:not(:-moz-lwtheme),
  #nav-bar[tabsontop=false]:not(:-moz-lwtheme) .toolbarbutton-text,
  #nav-bar + #customToolbars + #PersonalToolbar:-moz-any([collapsed=true],[moz-collapsed=true]) + #TabsToolbar[tabsontop=false]:last-child:not(:-moz-lwtheme) {
    background-color: transparent !important;
    color: var(--window-text-color);
    border-left-style: none !important;
    border-right-style: none !important;
  }
```

Also, if at all possible, avoid using `!important` in browser css, because it makes it hard/impossible to override with user styles in that case. Re-order statements so they cascade properly?

Lootyhoof:
The `!important` code was already present before my patch, FYI. ;)

That said, I'll take a look. For that error, is that the error console producing that or something else?

wolfbeast:
The `!important` remark was a general remark :) Sometimes it's hard to avoid.

The warning is thrown in the browser console.

Lootyhoof:
Hmm...I've an idea what this might be, though I'm not able to reproduce myself. Under what circumstances are those messages produced?

wolfbeast:
I just started up a fresh build and looked at the browser console, and I was greeted with that warning; so.. pretty much somewhere on startup.

PR #1227, opened 2017-07-09
Extensions - loadSubScript - add support for blob

janekptacijarabaci:
__Steps to reproduce:__

1) Open Scratchpad
3) Insert the code:
```
Components.utils.import("resource://gre/modules/Services.jsm");

var ns = {};
var b = new Blob([
  "var Exported = 17;"
]);
var blobUrl = URL.createObjectURL(b);
try {
  Services.scriptloader.loadSubScript(blobUrl, ns);
  alert(ns.Exported);
} finally {
  URL.revokeObjectURL(blobUrl);
}
```
4) Run

__Actual results__

Throws an error:
```
Exception: Trying to load a non-local URI.: blob:null/...
```

__Expected results__

alert("17");

---

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1221148

---

You add the label `Enhancement` (?), `Extensions`, `UXP-task`, please.

---

### DIFF
--- a/js/xpconnect/idl/mozIJSSubScriptLoader.idl
+++ b/js/xpconnect/idl/mozIJSSubScriptLoader.idl
@@ -18,7 +18,7 @@ interface mozIJSSubScriptLoader : nsISupports
      * In JS, the signature looks like:
      * rv loadSubScript (url [, obj] [, charset]);
      * @param url the url of the sub-script, it MUST be either a file:,
-     *            resource:, or chrome: url, and MUST be local.
+     *            resource:, blob:, or chrome: url, and MUST be local.
      * @param obj an optional object to evaluate the script onto, it
      *            defaults to the global object of the caller.
      * @param charset optionally specifies the character encoding of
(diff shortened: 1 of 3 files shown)


wolfbeast:
Um... why do we need this, exactly? Being able to load blobs as scripts is potentially dangerous, so I'd rather not.
The related bug seems to be a crude workaround for e10s sandboxing which is N/A for either Pale Moon or UXP, anyway, considering that was implemented by Mozilla AFTER our fork-off point for UXP.

janekptacijarabaci:
If it is not needed... It would be (much) easier to use.
(I thought it was not decided)

wolfbeast:
... ? I really don't understand what you're trying to say.

janekptacijarabaci:
If this property (e10s and/or content sandboxing) will not be used (or it does not exist, and will not be added to Basilisk: xy), this PR is unnecessary in this case.

wolfbeast:
Thanks, that clarifies it.
the content sandboxing (for e10s) will not be added to Pale Moon and UXP because it's directing process security responsibility away from the code, and will cause many extensions to stop working properly (there is no reason we can't trust our own code or installed extension code with file system access).
As such, this (breaking) code change isn't needed. Closing.

Issue #1228, opened 2017-07-09
Windows 10 dark window frame mode isn't aligned with Windows 10 native contrast ratio


Lootyhoof:
While looking at this stuff it's appeared to me that the `darkwindowframe` code doesn't appear to use the same contrast ratio as Windows 10 itself.

Using Windows 10 build 15063 ("Creator's Update"), with a titlebar such as the below in Notepad:

![Notepad](http://i.imgur.com/XuCop9c.png)

This is displayed as the following in Pale Moon, as the `darkwindowframe` code is set to `true`:

![Pale Moon](http://i.imgur.com/fzGjo8A.png)

This appears to be the case for several of the pre-set colours available within Windows 10, and only seems to switch to the darker window title and caption buttons when on a very bright titlebar colour.

wolfbeast:
This used to be pretty closely aligned (I tested with many of the preset colors in older versions of Win10) and it's actually using the accessibility formula to calculate. The "flip point" was mostly trial and error.

wolfbeast:
I calculated the flip point used in the code I just added (luminance 110) which is a contrast ration according to the WCAG formula of 9.62 -- that's almost exactly the 9.5 I've put in the darkwindowframe calculation... 

I'm not privy to what formula Win10 uses.

wolfbeast:
I poked [Technet](https://social.technet.microsoft.com/Forums/en-US/d942f9bd-7647-468f-869b-86f966bd42d0/how-is-the-contrast-ratio-calculated-for-deciding-text-should-be-black-or-white?forum=win10itprogeneral) to see if there's anyone who knows how exactly Windows does this contrast calculation.

wolfbeast:
The creator's update won't run in my test environment, I found out (VMWare 11 doesn't like it and goes black). I don't know if they changed this from the anniversary update or not, but i can see about making another more concerted trial run to get the contrast point more fine tunes than 9.5.

FTR, it's in browser.js line 906

wolfbeast:
I think setting the point to WCAG's 7:1 contrast ratio does the trick to get it closer to Windows' own.
See commit 8440d1a7c60d90ad3147759a09229c1d905c4fe9

wolfbeast:
OK, so, according to someone on the dev forum, it's documented as follows:

```Javascript
function accentColorUpdated(elementWithText)
{
    var uiSettings = new Windows.UI.ViewManagement.UISettings();
[...]

Issue #1222, opened 2017-07-08
Use accent color on Windows 10 titlebar more reliably


Lootyhoof:
I've been tracking the related Mozilla bugs for these for a while, and it looks like they're done, so time to put a fork in them.

To work around issues in Windows 10, we've tried multiple methods: the original Mozilla approach (draw everything ourselves but use a light grey window background permanently), as well as attempting to emulate the Windows Vista/7/8 method (let the system render the background and overlay our caption buttons). The former was met with criticism because Windows 10 started to allow people to use accent color in titlebars, while the latter is inconsistent in unusual DPIs and with non-standard system themes.

The new Mozilla approach, though, is to add a few new CSS selectors and media queries. In particular:
`-moz-win-accentcolor` (e.g. `background-color: -moz-win-accentcolor`) to get the system accent color.
`-moz-win-accentcolortext` (e.g. `color: -moz-win-accentcolortext`) to get the text styling associated with this - this could also help with our browser title code...
`-moz-windows-accent-color-applies` (e.g. `@media (-moz-windows-accent-color-applies)`) for when the user has set the accent color to the titlebar (by default Windows 10 just leaves it pure white).

This way, the approach is very much the same as their previous approach, except now with more color involved.

The patches for this are listed in BMO [1344910](https://bugzilla.mozilla.org/show_bug.cgi?id=1344910) and [1344917](https://bugzilla.mozilla.org/show_bug.cgi?id=1344917), with the theme changes in [1196266](https://bugzilla.mozilla.org/show_bug.cgi?id=1196266) (I will take this one, though the former two patches will need completing first).

Please add the labels `Theme/UI`, `CSS` (new label?) and `UXP-task` (or would you prefer this be raised over there instead?).

wolfbeast:
Jeez, annoying, both dependent bug have exactly the same push date and time.
I'll figure out which one was applied first.

Lootyhoof:
Not from what I've seen? According to [the commit](https://hg.mozilla.org/mozilla-central/rev/d07b25fd6533) it looks like 1344917 was pushed first, on 2017-06-08 (though it seems the tests were applied earlier, [here](https://hg.mozilla.org/mozilla-central/rev/d7dca95a5bc8)).

[...]

PR #1229, opened 2017-07-09
Draw a background on Windows 10 using the accent colour

Lootyhoof:
Ref. #1222 

When the accent colour is not set to the titlebar by the user (default), the following is used instead:

Default

When the window is inactive (regardless of accent colour settings), the following is shown:

Inactive

Note that this is affected by #1228, though potentially should be less noticeable with this patch (since previously the window title and caption buttons would have been different colours).

### DIFF
--- a/browser/base/content/browser-title.css
+++ b/browser/base/content/browser-title.css
@@ -164,6 +164,13 @@
       #main-window::after {
         text-align: left;
       }
+      
+      @media (-moz-windows-accent-color-applies: 0) {
+        #main-window:not(:-moz-window-inactive):not(:-moz-lwtheme)::after {
+          /* Default Windows 10 styling is dark - apply a light text styling */
+          color: white;
+        }
+      }
     }
 
     #main-window[sizemode="maximized"]::after {
--- a/browser/themes/windows/browser.css
+++ b/browser/themes/windows/browser.css
@@ -3056,7 +3056,157 @@ toolbar[brighttext] #addonbar-closebutton {
 /* ==== Windows 10 styling ==== */
 
   @media (-moz-os-version: windows-win10) {
-    /* Draw XUL caption buttons on Win10 in lwthemes */
+    /* Draw XUL caption buttons and background on Win10 */
+    @media (-moz-windows-accent-color-applies: 0) {
+      /* Default styling for when no accent color is applied */
+      #main-window:not(:-moz-window-inactive):not(:-moz-lwtheme) {
+        background-color: hsl(239, 33%, 38%);
+      }
+      
+      :root:not(:-moz-window-inactive):not(:-moz-lwtheme) {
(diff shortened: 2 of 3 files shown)


wolfbeast:
How does this work with the pseudo-controls we've been using?
I'd have expected part of this patch to remove them, since we can just draw our own entirely now, but I don't see any of that...?

Lootyhoof:
This makes us use our own buttons fully, whereas before we only used them when hovering over the system controls.

Lootyhoof:
I just realised that this will require changes to themes, too - in particular, the browser title code: themes will either need to work around the window title being white when no accent colour is applied (white on white), or port this patch.

Please can you add the label `Theme changes`?

wolfbeast:
I'm wondering about something:

You said:

> Note that this is affected by #1228, though potentially should be less noticeable with this patch (since previously the window title and caption buttons would have been different colours).

Previously, this would all be determined by "darkwindowframe", but now it isn't (caption buttons depend on darkwindowframe, but text does not). So doesn't this create a discrepancy now while there was none before?
Can't we just use the accentcolortext as stroke color in our SVGs, considering you can use CSS in them?

Lootyhoof:
Hmm...possibly. I did consider this, though my initial concern was that it would affect those without an accent colour applied - with them taking another route for the caption buttons though I guess this shouldn't be a concern.

I'll send a fixup soon to rectify. 👍

wolfbeast:
Also, I noticed we lost some transition effects when going from hovered to non-hovered state.

wolfbeast:
Another note... When the accent color is not set, I think we have a conflict, as I had it switch to the intended color, but also had it switch to a darker grey. Probably because of contention with the browser.js script and the new lookandfeel cpp code... :P

wolfbeast:
I looked at the code involved in Firefox, and it seems they are not using the "darkwindowframe" setup for Windows 10, there, only for Windows 8. I think we need to write down what exactly is used where so we can organize this mess.

Lootyhoof:
The Mozilla approach makes use of [-moz-context-properties](https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-context-properties), a CSS property that wasn't added until Firefox 55 (BMO [1058040](https://bugzilla.mozilla.org/show_bug.cgi?id=1058040)). If this is something you want to add, that would make things easier, though it can be worked around if necessary.

wolfbeast:
`-moz-context-properties` looks like a really dirty hack to me (it seems to just reduce the need for individual SVGs for individual situations... and making it all managed through a single CSS; taking optimization a bit too far) and I'd rather not add it to Pale Moon, tbh. I understand it can be useful in some cases, but I'd prefer to have separate SVG files instead of hacking around what is supposed to be a unidirectional hierarchy.

PR #1230, opened 2017-07-09
Align frontend code with updated BookmarkHTMLUtils.jsm

JustOff:
Based on https://bugzilla.mozilla.org/show_bug.cgi?id=968177.

See [bug report on forum](https://forum.palemoon.org/viewtopic.php?f=37&t=16109).

### DIFF
--- a/browser/components/nsBrowserGlue.js
+++ b/browser/components/nsBrowserGlue.js
@@ -960,10 +960,7 @@ BrowserGlue.prototype = {
         else {
           // We have created a new database but we don't have any backup available
           importBookmarks = true;
-          var dirService = Cc["@mozilla.org/file/directory_service;1"].
-                           getService(Ci.nsIProperties);
-          var bookmarksHTMLFile = dirService.get("BMarks", Ci.nsILocalFile);
-          if (bookmarksHTMLFile.exists()) {
+          if (yield OS.File.exists(BookmarkHTMLUtils.defaultPath)) {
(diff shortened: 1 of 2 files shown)


wolfbeast:
your second commit that had nothing to do with the bug reported caused breakage. Backed out.

JustOff:
Yeah, sorry, that was really very stupid :anguished:

But we still need to take into account `browser.bookmarks.file` and therefore use `BookmarkHTMLUtils.defaultPath` instead of `dirService.get("BMarks", Ci.nsILocalFile)`. 

I'll double-check everything and file a new PR tomorrow.

JustOff:
Today I re-checked everything a few more times, ran tests using Scratchpad and could not find any problems with the code you [reverted](https://github.com/MoonchildProductions/Pale-Moon/commit/f02510df94d71c3bf12c2e03d7fefdf485b7057b).

Could you please be more specific, what kind of breakage did it cause?

wolfbeast:
The "defaultpath" commit was NOT reverted. See now how it gets confusing when you lump multiple things together?

As for the breakage cause by the commit that WAS reverted: launch a new profile (e.g. `mach run`) and watch the gloriously empty bookmarks toolbar and bookmarks organizer.

JustOff:
Thanks, now I found the cause, I was just blind and forgot to add the osfile.js import :flushed:

And as for your remark: these commits were not lumped by accident, they are actually refer to the one issue.

PR #1231, opened 2017-07-10
[minor fix] CSS - user-select: none

janekptacijarabaci:
An example:
https://bugzilla.mozilla.org/attachment.cgi?id=8823010

__Steps to reproduce:__

Copy string "str" to clipboard (Ctrl-C).
Open attached example.
Click in input, press Ctrl+V.
Press Ctrl+A on the whole page.

__Actual results__
`full`: this word is selected

__Expected results__
`full`: this word is omitted

---

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1327902

---

### DIFF
--- a/dom/base/nsRange.cpp
+++ b/dom/base/nsRange.cpp
@@ -3167,7 +3167,18 @@ nsRange::ExcludeNonSelectableNodes(nsTArray<nsRefPtr<nsRange>>* aOutRanges)
           }
 
           // Create a new range for the remainder.
-          rv = CreateRange(node, 0, endParent, endOffset,
+          nsINode* startParent = node;
+          int32_t startOffset = 0;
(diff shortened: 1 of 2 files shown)


wolfbeast:
Is this also dependent on the #1232 prerequisite? because I don't think we split these selections up into ranges as-is now (by design) and as a result this PR won't fix anything.

janekptacijarabaci:
@wolfbeast

You see also: https://github.com/MoonchildProductions/Pale-Moon/pull/1232#issuecomment-314323148

Due to forward compatibility.

But it's just a suggestion.

wolfbeast:
Please keep in mind that this is just a display issue. the copy operation is correct.
so if you prefer to improve unused code.. well I guess it's your time to use...

PR #1214, opened 2017-07-05
DOM - Selection API - implement selection events

janekptacijarabaci:
Ad #1213

An example:
http://www.analizeeconomice.ro/2016/09/exporturile-romaniei-in-paradisurile.html
(PM won't load text in this search box)
See also:
https://forum.palemoon.org/viewtopic.php?f=29&p=116802#p116802

---

## 1)

DOM - implement selection events (base)

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=571294
https://bugzilla.mozilla.org/show_bug.cgi?id=1248148

## 2)

DOM - implement selection events (rename mApplyUserSelectStyle)

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1199288

## 3)

DOM - implement selection events (potentially crash)

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1204291
https://bugzilla.mozilla.org/show_bug.cgi?id=1350972

## 4)

DOM - implement selection events (move cancelBubble to Event)

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1298970

## 5)

DOM - implement selection events
(fire selectstart and selectionchange events on the input node when the selection in that editor changes)

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1196479
https://bugzilla.mozilla.org/show_bug.cgi?id=1231193

## 6)

DOM - implement selection events (enable select events for chrome consumers)

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1231923

## 7)

DOM - implement selection events
(ensure that we don't dispatch the selectstart and selectionchange events for the selections inside input and textarea elements)

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1309628

---

You add the label `Enhancement` / `Web Compatibility` (?), please.

---

I've created the new build (x32, Windows) and tested:

__...but it needs to be tested properly.__

You add the label `Verification Needed`, please.

Issue #1213, opened 2017-07-05
[suggestion] DOM - Selection API (draft) - implement selection events


janekptacijarabaci:
DOM - Selection API - implement selection events (`W3C Working Draft`):

http://w3c.github.io/selection-api/#selectstart-event

> 6.1 selectstart event

> When the user agent is about to associate a new range newRange to the selection in response to a user initiated action, the user agent must fire ([DOM4]) an event with the name selectstart, which bubbles and is cancelable, at the node associated with the boundary point of newRange's start prior to changing the selection if the selection was previously empty or the previously associated range was collapsed.

> If the event is canceled, the user agent must not change the selection.

> The user agent must not fire the event when the user agent sets the selection empty.

> 6.2 selectionchange event

> When the selection is dissociated with its range, associated with a new range or the associated range's boundary point is mutated either by the user or the content script, the user agent must queue a task to fire an event with the name selectionchange, which does not bubble and is not cancelable, at the document associated with the selection.

---

You add the label `Web Compatibility`, please.

wolfbeast:
Can someone please verify this is working in the unstable builds so I can close this issue?

JustOff:
[Search box now works fine for me on OP's site](https://forum.palemoon.org/viewtopic.php?f=29&p=116802#p116889) and I also can confirm this.

janekptacijarabaci:
So there is some problem somewhere. `selectstart` does not work.

h1z1:
Curious little change.  Does this not mean sites are able to further spy on these events as well?

wolfbeast:
Closing as verified.

@h1z1 there is no issue with websites being able to see when a user selects text. The benefits certainly outweigh any potential corner case where you might want to prevent a site knowing when you interact with its content.

PR #1232, opened 2017-07-10
GitHub - Total freeze when dragging selected text; Mouse cursor only shows part of the dragged selection

janekptacijarabaci:
You must first implement (temporarily):
[nsSelection.cpp#L3611](https://github.com/MoonchildProductions/Pale-Moon/blob/27.3.0_Release/layout/generic/nsSelection.cpp#L3611)
`// #if 0`
[nsSelection.cpp#L3641](https://github.com/MoonchildProductions/Pale-Moon/blob/27.3.0_Release/layout/generic/nsSelection.cpp#L3641)
`// #endif`

## 1)
__GitHub - Total freeze when dragging selected text__

Confirmed.

__Steps to reproduce:__

1. View source code on GitHub
2. Select some code
3. Drag it

__Actual results__
Cursor state switches, Firefox freezes completely.

__Expected results__
No freeze (just a short freeze).

---

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1258476
(see also [nsSelection.cpp#L3609](https://github.com/MoonchildProductions/Pale-Moon/blob/27.3.0_Release/layout/generic/nsSelection.cpp#L3609))

## 2)
__Mouse cursor only shows part of the dragged selection__

An example:
https://bugzilla.mozilla.org/attachment.cgi?id=8742129
https://bugzilla.mozilla.org/attachment.cgi?id=8742131

---

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1265249

---

### DIFF
--- a/layout/base/nsPresShell.cpp
+++ b/layout/base/nsPresShell.cpp
@@ -42,6 +42,7 @@
 #include "nsPresShell.h"
 #include "nsPresContext.h"
 #include "nsIContent.h"
+#include "nsIContentIterator.h"
 #include "mozilla/dom/BeforeAfterKeyboardEvent.h"
 #include "mozilla/dom/Element.h"
 #include "mozilla/dom/Event.h" // for Event::GetEventPopupControlState()
@@ -5087,8 +5088,7 @@ PresShell::CreateRangePaintInfo(nsIDOMRange* aRange,
   nsIDocument* doc = startParent->GetCrossShadowCurrentDoc();
   if (startParent == doc || endParent == doc) {
     ancestorFrame = rootFrame;
-  }
-  else {
+  } else {
     nsINode* ancestor = nsContentUtils::GetCommonAncestor(startParent, endParent);
     NS_ASSERTION(!ancestor || ancestor->IsNodeOfType(nsINode::eCONTENT),
                  "common ancestor is not content");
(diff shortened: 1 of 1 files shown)


JustOff:
I can't reproduce neither of the bugs, what am I doing wrong?)

janekptacijarabaci:
@JustOff
Ad 2) - You must first implement: #1231

I'm sorry, I should have added it.

wolfbeast:
The code is disabled for a reason.

wolfbeast:
See code comment:
```
/* XXX: The following code block splits selections up into selection ranges
--
* excluding nodes that are unselectable because of -moz-user-select CSS
* keywords. The resulting array of ranges causes a huge performance hit in
* layout for minor display enhancements for selected text (the text actually
* copied to clipboard etc. is correct).
* For now, this behavior has been disabled until a better/acceptable solution
* is available.
* See also BMO bugs 619273, 739396, 1216001 and 1258476
*/
```

wolfbeast:
Also, once more...
1 issue/PR for 1 bug, please. It's impossible to keep track of which parts of which bug(s) have been addressed by what otherwise.

janekptacijarabaci:
@wolfbeast
> It's disabled for a reason!

Yes. But nonetheless, it has already been modified in PR #1214:
[nsSelection.cpp#L3661-L3746](https://github.com/MoonchildProductions/Pale-Moon/blob/13bd588fbf5a8653b245b87b0e501d1c6e33301c/layout/generic/nsSelection.cpp#L3661-L3746)

And you see: https://github.com/MoonchildProductions/Pale-Moon/issues/1213#issuecomment-314021639
(I suggest adding a comment there - `selectstart`, PR #1214...)

> 1 issue/PR for 1 bug

In this case, there would be a conflict (the same file). I divided it into commits. If it's better for you... I'll try to remember it for next time.

wolfbeast:
Editing the same file in multiple PRs is not a problem as long as they don't touch the same code (in which case the order matters). these two things are clearly 2 different bugs, aren't they?

And it doesn't matter if it's already modified by another of your own changes along with many changes that were ported.

Still, I guess it doesn't matter, and will improve the code, even though the code is currently not in use.

PR #1242, opened 2017-07-15
Rework media IsType Supported + HW accel fixes

trav90:
Followup to #1033.

The PR fixes the following:

- Reworks DecoderTraits:CanHandleMediaType and properly reports the mimetype in MSE. Currently, MSE only checks the file type (be it MP4 or WebM), and if the HTMLMediaElement can play that type of file. With this PR, MSE will now only report MP4 support if the required support is found on the system (primarily applies to Linux here, should FFmpeg not be installed on the system for example).
- Similarly, MediaResource often does not reliably detect the format (and doesn't work at all with the new MSE) when determining if HW acceleration can be used or not. With this PR, we let the PlatformDecoderModule determine this instead based on parsed metadata. This possibly could help with reports on the forum of high CPU usage? (Though it should be noted that I did not see a noticeable difference in CPU usage during my testing of this).
- Fixed a small bug where HW acceleration could not be disabled for video playback when using new MSE.

I've been testing this the past couple hours and have seen no issues. Tested with both MP4 and WebM on YouTube, videos on Vimeo and Twitch, as well as all the tests on [Quirksmode](http://quirksmode.org/html5/tests/video.html) and [dogphilosophy.net](http://hpr.dogphilosophy.net/test/).

This definitely needs thorough testing on Windows to make sure that all supported formats are still playable (due to the CanHandleMediaType and MSE mimetype reporting changes). As such, I'm adding the `Verification Needed` label.

### DIFF
--- a/dom/media/DecoderTraits.cpp
+++ b/dom/media/DecoderTraits.cpp
@@ -536,7 +536,6 @@ DecoderTraits::CanHandleMediaType(const char* aMIMEType,
   if (IsMP3SupportedType(nsDependentCString(aMIMEType))) {
     return CANPLAY_MAYBE;
   }
-#endif
 #ifdef MOZ_OMX_DECODER
   if (IsOmxSupportedType(nsDependentCString(aMIMEType))) {
     return CANPLAY_MAYBE;
--- a/dom/media/DecoderTraits.cpp
+++ b/dom/media/DecoderTraits.cpp
@@ -186,17 +186,20 @@ static char const *const gWebMCodecs[7] = {
   "opus",
   nullptr
 };
+#endif
 
 /* static */ bool
 DecoderTraits::IsWebMType(const nsACString& aType)
 {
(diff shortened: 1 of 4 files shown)


trav90:
I tested this using the latest unstable while I was in temporary possession of a Windows 7 machine and appears to be working as intended on that platform as well. Removing the `Verification Needed` label.

trav90:
Since it appears we will be releasing a point release to address other issues anyway, could 8b926f54cb18b98f6950ed060145a5e347e2565a be uplifted as well (the other commits in this PR shouldn't be needed)?

wolfbeast:
Looks low-risk and self-contained enough, so sure, I'll uplift it to the relbranch.

trav90:
Thanks!

wolfbeast:
Uplifted in eb1ba44fa41598a4de318e414e5e9111bf9a2e85

PR #1253, opened 2017-07-21
Media mirroring/task queue improvements

trav90:
This is a small PR that does the following to help reduce the chances of race conditions in the code:

- Make MediaDecoderReader compatible to use with StateMirroring
- Enable tail dispatch for the decode task queue

Additionally, the PR tweaks usage of MediaTaskQueue to prevent a potential deadlock/hang.

I've been testing this on YouTube (MSE+MP4, MSE+WebM & plain MP4) and Twitch the past few hours. Appears to be working as intended.

### DIFF
--- a/dom/media/MediaDecoderReader.cpp
+++ b/dom/media/MediaDecoderReader.cpp
@@ -342,7 +342,9 @@ MediaDecoderReader::RequestAudioData()
 void
 MediaDecoderReader::BreakCycles()
 {
-  mTaskQueue = nullptr;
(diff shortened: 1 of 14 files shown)


JustOff:
With this PR merged I can't play html5 video at all (checked on youtube & vimeo) and browser does not want to exit after playback attempt until I kill it (PMx86@win7x64) :anguished:

trav90:
Can't reproduce on Linux.

@wolfbeast, go ahead and back this out. I'll try to investigate this once WebM is done.

Issue #1254, opened 2017-07-22
Fix vector leaks and append calls.


wolfbeast:
Our vector code is rather messy at the moment and causes issues on some sites by leaking.
We need to port https://bugzilla.mozilla.org/show_bug.cgi?id=1231224 to clean this up.
See also: https://forum.palemoon.org/viewtopic.php?f=57&t=16216

janekptacijarabaci:
Ad https://forum.palemoon.org/viewtopic.php?f=57&t=16216

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=1225396
AFAIK - this problem is more likely:
https://shared.reliclink.com/d…990ac4794ed.live-shared.min.js
`try{for(var _step,_iterator=componentClass.theme.classes[Symbol.iterator]()...`

wolfbeast:
Split off the iterator issue to #1256 -- that one should be implemented first, looks like.

PR #1257, opened 2017-07-23
Fix vector leaks and append calls

janekptacijarabaci:
Ad https://github.com/MoonchildProductions/Pale-Moon/issues/1254

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1231224

Ad https://hg.mozilla.org/mozilla-central/rev/6d6334c0964f
__Note that this is only partial implementation!__
(https://bugzilla.mozilla.org/show_bug.cgi?id=1234402
\- https://hg.mozilla.org/mozilla-central/rev/a1c4a945741f)
After full implementation - a problem with build on:
https://github.com/MoonchildProductions/Pale-Moon/blob/27.4.0_Release/js/src/irregexp/RegExpEngine.cpp#L1004

---

__Patches must be thoroughly (checked and) tested!__

---

### DIFF
--- a/js/src/asmjs/AsmJSLink.cpp
+++ b/js/src/asmjs/AsmJSLink.cpp
@@ -682,7 +682,7 @@ CallAsmJS(JSContext* cx, unsigned argc, Value* vp)
     // stack since this would leave the module in a state where profiling is
     // enabled but the stack isn't unwindable.
     if (module.profilingEnabled() != cx->runtime()->spsProfiler.enabled() && !module.active())
-        module.setProfilingEnabled(cx->runtime()->spsProfiler.enabled(), cx);
+        module.setProfilingEnabled(cx, cx->runtime()->spsProfiler.enabled());
 
     // The calling convention for an external call into asm.js is to pass an
     // array of 16-byte values where each value contains either a coerced int32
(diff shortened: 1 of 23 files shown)


wolfbeast:
Can you indicate please what you ran into in your full implementation? What exactly didn't you implement of the changeset? What error did you run into?

janekptacijarabaci:
Full implementation = https://hg.mozilla.org/mozilla-central/rev/6d6334c0964f

```
[drive]:\[path]\js\src\irregexp/RegExpEngine.cpp(1004) :
error C3861: 'Move': identifier not found
```
This variable (`new_alternatives`) does not exist
(after: https://github.com/MoonchildProductions/Pale-Moon/blob/27.4.0_Release/js/src/irregexp/RegExpEngine.cpp#L993).

wolfbeast:
Sorry but I don't understand the issue. Those lines aren't touched by the changeset you linked to (only the RAII removal is applicable there which may not be present in our implementation to begin with)

Please indicate exactly what you did NOT implement and why.

janekptacijarabaci:
I didn't implement the following:
https://hg.mozilla.org/mozilla-central/rev/6d6334c0964f

It gives the error already mentioned. And yes, I do not see any sense in it. You can try it out by building it.

wolfbeast:
Well, it's not exactly important. Most likely we are using a different version of irregexp anyway (which is third party code and actually not sticking to regex standards as well, but it's what websites expect) and it's just for suppression of warnings, so we can leave it as is and call it done for the day.

Issue #1266, opened 2017-07-27
[suggestion] DOM - CSSOM View Module (draft) - implement scrollingElement


janekptacijarabaci:
DOM - CSSOM View Module - implement scrollingElement (`W3C Working Draft`):

https://drafts.csswg.org/cssom-view/#dom-document-scrollingelement

> The scrollingElement attribute, on getting, must run these steps:
    - 1. If the Document is in quirks mode, follow these substeps:
      -- 1. If the HTML body element exists, and it is not potentially scrollable, return the HTML body element and abort these steps.
      -- 2. Return null and abort these steps.
    - 2. If there is a root element, return the root element and abort these steps.
    - 3. Return null.

> Note: For non-conforming user agents that always use the quirks mode behavior for scrollTop and scrollLeft, the scrollingElement attribute is expected to also always return the HTML body element (or null if it does not exist). This API exists so that Web developers can use it to get the right element to use for scrolling APIs, without making assumptions about a particular user agent’s behavior or having to invoke a scroll to see which element scrolls the viewport.

> Note: The HTML body element is different from HTML’s document.body in that the latter can return a frameset element.

---

You add the label `Web Compatibility`, please.

wolfbeast:
Current unstable has this implemented.

PR #1267, opened 2017-07-27
DOM - CSSOM View Module - implement scrollingElement

janekptacijarabaci:
Ad #1266

An example:
(any page)

Scratchpad:
```
var scrollElm = document.scrollingElement;
scrollElm.scrollTop = 100;
```

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1153322
https://bugzilla.mozilla.org/show_bug.cgi?id=1265032

See also:
https://developer.mozilla.org/en-US/docs/Web/API/Document/scrollingElement

---

You add the label `Web Compatibility`, please.

---

I've created the new build (x32, Windows) and tested:

### DIFF
--- a/dom/base/nsDocument.cpp
+++ b/dom/base/nsDocument.cpp
@@ -9,7 +9,7 @@
  */
 
 #include "nsDocument.h"
-
+#include "nsIDocumentInlines.h"
 #include "mozilla/ArrayUtils.h"
 #include "mozilla/AutoRestore.h"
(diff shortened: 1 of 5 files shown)


wolfbeast:
More importantly: https://drafts.csswg.org/cssom-view/#dom-document-scrollingelement -- that's the draft we should be looking at, not Mozilla's decisions what to copy from Chrome.

PR #1270, opened 2017-07-29
URL - whitespace characters (an error: 0x00, 0x09, 0x0A, 0x0D => 0x00 (only))

janekptacijarabaci:
Ad https://forum.palemoon.org/viewtopic.php?f=29&p=118429

Can't drag some bookmarks to desktop, e.g.:
https://en.wikipedia.org/wiki/Wikipedia:Templates_for_discussion/Log/2017_July_27#Template:Recent_death_Aboriginal_Aus

---

These characters: 0x09, 0x0A, 0x0D (\t\n\r) are stripped out instead of returning error.

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1189852

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=1272284
https://bugzilla.mozilla.org/show_bug.cgi?id=1271955
https://bugzilla.mozilla.org/show_bug.cgi?id=1190027
etc.
But it is more risky. Meanwhile, there were many fixes.

---

### DIFF
--- a/netwerk/base/nsStandardURL.cpp
+++ b/netwerk/base/nsStandardURL.cpp
@@ -1241,13 +1241,10 @@ nsStandardURL::SetSpec(const nsACString &input)
     if (!spec || !*spec)
         return NS_ERROR_MALFORMED_URI;
 
-    int32_t refPos = input.FindChar('#');
-    if (refPos != kNotFound) {
-        const nsCSubstring& sub = Substring(input, refPos, input.Length());
-        nsresult rv = CheckRefCharacters(sub);
-        if (NS_FAILED(rv)) {
-            return rv;
-        }
+    // NUL characters aren't allowed
+    // \r\n\t are stripped out instead of returning error(see below)
+    if (input.FindChar('\0') != kNotFound) {
+        return NS_ERROR_MALFORMED_URI;
     }
 
     // Make a backup of the curent URL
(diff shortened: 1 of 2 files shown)


wolfbeast:
Aiming for spec compliance here:

https://url.spec.whatwg.org/#fragment-state

For fragments:
> If c is not a URL code point and not U+0025 (%), validation error. 
> The URL code points
are ASCII alphanumeric, U+0021 (!), U+0024 ($), U+0026 (&), U+0027 ('), U+0028 LEFT PARENTHESIS, U+0029 RIGHT PARENTHESIS, U+002A (*), U+002B (+), U+002C (,), U+002D (-), U+002E (.), U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+003F (?), U+0040 (@), U+005F (_), U+007E (~), and code points in the range U+00A0 to U+10FFFD, inclusive, excluding surrogates and noncharacters. 

0x00, 0x09, 0x0A and 0x0D are all invalid URL code points and therefore must throw a validation error (not a valid URL) according to the spec.
I'm not sure if just stripping them (against the spec) is a good idea.

wolfbeast:
The spec does say to strip leading and trailing control characters, so we'll want https://bugzilla.mozilla.org/show_bug.cgi?id=1271955

According to the spec we should also have https://bugzilla.mozilla.org/show_bug.cgi?id=1272284 ported to make sure that path/query/hash is properly encoded in light of control chars.

So, the question is really: web compatibility or spec compliance? Considering stripping trailing is part of the spec and I'm guessing the problem here is trailing newlines (although I don't really understand why the example silently fails) we could take your commit, and the other two I indicated to strike a balance here.

janekptacijarabaci:
I would prefer to leave any changes to someone more qualified...

PR #1284, opened 2017-08-08
Remove BatteryManager API

stroucki:
I suggest that the BatteryManager API be removed. Firefox has already removed it last November.
[Grauniad story](https://www.theguardian.com/technology/2016/nov/01/firefox-disable-battery-status-api-tracking)
[Bruce Schneier's blog](https://www.schneier.com/blog/archives/2016/11/firefox_removin.html)

Pale Moon should too. Sample commit to my fork is linked.


mattatobin:
One, commenting out code instead of removing it is NOT the proper procedure. TWO There is no reason at this stage to do any massive testing and code cleanup in the Tycho codebase.

stroucki:
As I said it was a sample. And it's my fork, so I can do what I want.

mattatobin:
I hope you aren't distributing that "fork" with Pale Moon branding.

stroucki:
I am not.

stroucki:
You are of course correct that removing it is the best course of action, which I have done now. However my changes are Mac only at this time, so Unix and Windoze versions likely need additional modifications.
It is sufficient for my purposes, and it is freely provided.

wolfbeast:
There's no point in accepting a half-done code cleanup patch, freely provided or not. It'll either have to be done completely or not at all; having dead code/broken calling sites in "the OS you don't happen to be using" is not going to fly, sorry.

stroucki:
I don't mind if you don't accept the patch, but I recommend that you follow firefox and remove the battery API.

Issue #921, opened 2017-02-26
Brotli Accept-Encoding/Content-Encoding


wolfbeast:
Using the brotli compressor/decompressor for transfer of HTTP document data can lead to significant bandwidth savings (and resulting faster loads) at low CPU cost.

This enhancement was included in Firefox 44, and we should consider porting it across.
https://bugzilla.mozilla.org/show_bug.cgi?id=366559

Priority: not at all essential, but "nice to have".

wolfbeast:
Base functionality is in with #1289 but likely needs some follow-up fixes.

PR #1289, opened 2017-08-14
Brotli Accept-Encoding/Content-Encoding #921

Gpower2:
Enable Brotli Accept-Encoding/Content-Encoding
Ported bugs: 366559, 1211916, 1211580

Tested and it works like a charm!
Test page: https://www.iisbrotli.com/support

PS. Sorry for the messy PR, I've been working on this for quite some time with a lot of breaks in between.
Fortunately, only the correct files are touched.


janekptacijarabaci:
See also [bug 1224291](https://bugzilla.mozilla.org/show_bug.cgi?id=1224291) ([bug 1243724](https://bugzilla.mozilla.org/show_bug.cgi?id=1243724)).
AFAIK (I don't known if it's needed).

Gpower2:
These do not apply for the version of Brotli that is used with this PR. ;)
I do intend on porting 1340910 which updates Brotli to v0.6 and was completed only 5 days ago,
however I don't currently have the time to carefully port it, so something to wait for in the future! :)

wolfbeast:
Unfortunately Brotli 0.6 will be required to avoid a few sec issues. I'm not comfortable enabling a pre-0.6 version in Pale Moon. So please add the update to 0.6 when you can and then ping me to merge it!

wolfbeast:
Ah wait scratch that. seems that 0.4 is fine for now.

wolfbeast:
Merging this in. Use #1290 and #1291 to address the lib updates needed for WOFF2/Brotli.

wolfbeast:
Unfortunately it doesn't "work like a charm" on real sites. It causes no-content responses on Facebook and Google Search if enabled. There are likely some follow-up bugs that need porting; I've removed the accept-encoding header from the prefs on trunk for now until this is sorted out.

PR #1301, opened 2017-08-19
about:permissions - logins - follow up

janekptacijarabaci:
Ad https://forum.palemoon.org/viewtopic.php?f=57&t=16428

> For example, it will not list forums.mozillazine.org. I have tried reloading the list of sites after visiting and logging into MozillaZine, but so far it hasn't changed anything.

---

### DIFF
--- a/browser/components/preferences/aboutPermissions.js
+++ b/browser/components/preferences/aboutPermissions.js
@@ -458,7 +458,6 @@ let AboutPermissions = {
   /**
    * When adding sites to the dom sites-list, divide workload into intervals.
    */
-  LIST_BUILD_CHUNK: 5, // interval size
   LIST_BUILD_DELAY: 100, // delay between intervals
 
   /**
@@ -846,9 +845,6 @@ let AboutPermissions = {
     try {
       let logins = Services.logins.getAllLogins();
       logins.forEach(function(aLogin) {
-        if (itemCnt % this.LIST_BUILD_CHUNK == 0) {
-          yield true;
-        }
         try {
           // aLogin.hostname is a string in origin URL format
           // (e.g. "http://foo.com").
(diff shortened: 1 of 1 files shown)


wolfbeast:
Can you please describe what exactly you're changing in the behavior here?

janekptacijarabaci:
Ad https://github.com/MoonchildProductions/Pale-Moon/pull/1301/commits/485c4772091337bd4173d348a2546fa167391068
- Ending the loop is not working properly (some of the data - logins - was omitted).
IMHO: It is not needed.

Ad https://github.com/MoonchildProductions/Pale-Moon/pull/1301/commits/ab9c5c2a04d7ce6c37adc3ef3c17175bd0060c40
- `Services.logins.getAllLogins()`:
It works correctly, but throws in Browser Console: `Invalid chrome URI: /`
There is a bug somewhere (see also https://bugzilla.mozilla.org/show_bug.cgi?id=1291346 / https://bugzilla.mozilla.org/show_bug.cgi?id=1283994). But `Services.logins.searchLogins()` with `httpRealm: null` works correctly (without error generation).

janekptacijarabaci:
Ad throws in Browser Console: `Invalid chrome URI: /` (in `NetUtil.newURI`).

An example (Scratchpad, `Environment` - `Browser`):
``` javascript
Components.utils.import("resource://gre/modules/Services.jsm");
let logins = Services.logins.getAllLogins();
logins.forEach(function(aLogin) {
  try {
    let uri = NetUtil.newURI(aLogin.hostname);
    console.log(aLogin.hostname);
  } catch (e) {
  }
});
```
Confirmed for Pale Moon 27+ (i.e. alpha versions).
Firefox 24 / 38, Pale Moon 26.5 - not affected.

PR #1322, opened 2017-08-31
[suggestion] Main menu - Restart normally

janekptacijarabaci:
Ad https://forum.palemoon.org/viewtopic.php?f=13&t=16467

- `Restart normally…` in the Help menu
- Remove unused (AFAIK) `appMenuSafeMode.accesskey`

---

---

### DIFF
--- a/browser/locales/en-US/chrome/browser/browser.dtd
+++ b/browser/locales/en-US/chrome/browser/browser.dtd
@@ -334,7 +334,7 @@ These should match what Safari and other Apple applications use on OS X Lion. --
 <!ENTITY appMenuWebDeveloper.label "Web Developer">
 <!ENTITY appMenuGettingStarted.label "Getting Started">
 <!ENTITY appMenuSafeMode.label "Restart with Add-ons Disabled…">
-<!ENTITY appMenuSafeMode.accesskey "R">
+<!ENTITY appMenuRestart.label "Restart normally…">
 
 <!ENTITY openCmd.commandkey           "l">
 <!ENTITY urlbar.placeholder2          "Search or enter address">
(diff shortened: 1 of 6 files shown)


wolfbeast:
If adding a restart menu entry, it should not be in help; it would be better to call it just "Restart" and put it in the main appmenu/classic file menu right above "Exit".

browser/locales/en-US/chrome/browser/browser.properties
```diff
@@ -348,10 +348,10 @@ safeModeRestartPromptTitle=Restart with Add-ons Disabled
 safeModeRestartPromptMessage=Are you sure you want to disable all add-ons and restart?
 safeModeRestartButton=Restart
 
-# restartNormally
-restartNormalPromptTitle=Restart normally
-restartNormalPromptMessage=Are you sure you want to restart Pale Moon?
-restartNormalButton=Restart
+# restart
+restartPromptTitle=Restart
+restartPromptMessage=Are you sure you want to restart Pale Moon?
```
wolfbeast: Please avoid hard-coding the application name. Use a string variable instead to use `brandShortName`.

browser/locales/en-US/chrome/browser/browser.dtd
```diff
@@ -549,6 +548,9 @@ you can use these alternative items. Otherwise, their values should be empty.  -
 <!ENTITY zoomInButton.tooltip           "Zoom in">
 <!ENTITY zoomControls.label             "Zoom Controls">
 
+<!ENTITY appMenuRestart.label           "Restart…">
```
wolfbeast: Should not have an ellipsis, since it's a single command with a single outcome.

janekptacijarabaci: OK, I also counted the question.

PR #1323, opened 2017-08-31
Main menu - Restart (improvements)

janekptacijarabaci:
Ad https://github.com/MoonchildProductions/Pale-Moon/pull/1322

See:
https://forum.palemoon.org/viewtopic.php?f=13&t=16467#p120690

---

- Added `Ctrl+Alt+R`
- Notify all windows that an application quit has been requested
- about:support - added comments
---

Not added `browser.showQuitWarning`
- It already uses a different question - and I can not do it now. If anyone wants to take it over...

---

---

### DIFF
--- a/browser/base/content/browser-appmenu.inc
+++ b/browser/base/content/browser-appmenu.inc
@@ -188,7 +188,8 @@
       <menuitem id="appmenu_restart"
                 class="menuitem-iconic"
                 label="&appMenuRestart.label;"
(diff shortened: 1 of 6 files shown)


janekptacijarabaci:
@JustOff 
There's a new string:
https://github.com/MoonchildProductions/Pale-Moon/pull/1323/files#diff-38fb3065c40acfe9cd97e7cd67908052R553
Thank you.

PR #1339, opened 2017-09-09
Detect WEBP from content even if it's received with wrong mime type

JustOff:
This resolves ["No images on Coursera"](https://forum.palemoon.org/viewtopic.php?f=37&t=16562) issue.

### DIFF
--- a/image/src/imgLoader.cpp
+++ b/image/src/imgLoader.cpp
@@ -2400,6 +2400,26 @@ nsresult imgLoader::GetMimeTypeFromContent(const char* aContents, uint32_t aLeng
     aContentType.AssignLiteral(IMAGE_JPEG);
   }
 
+  /* maybe a WEBP? */
+  /* WebP files are technically RIFF files. RIFF is a container format like TIFF.
+   * WebP files are RIFF files that contain a single WEBP chunk.
+   *
+   * To detect it, first check that the first 4 bytes are "RIFF",
+   * and then that bytes 8-11 are "WEBP".
+   */
+  else if (aLength >= 12 &&
+     ((unsigned char)aContents[0])==0x52 &&
+     ((unsigned char)aContents[1])==0x49 &&
+     ((unsigned char)aContents[2])==0x46 &&
+     ((unsigned char)aContents[3])==0x46 &&
+     ((unsigned char)aContents[8])==0x57 &&
+     ((unsigned char)aContents[9])==0x45 &&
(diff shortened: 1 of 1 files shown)


wolfbeast:
I see, so cloudfront converts to WebP without changing the MIME type.

Hmm this raises another question though: why is "nosniff" ignored then for these images?

JustOff:
> I see, so cloudfront converts to WebP without changing the MIME type.

Yes, they do. Probably you [were mistaken](https://forum.palemoon.org/viewtopic.php?p=120834#p120834) deciding that it was JPG because of [bug 1347958](https://bugzilla.mozilla.org/show_bug.cgi?id=1347958).

> Hmm this raises another question though: why is "nosniff" ignored then for these images?

I did not dig into the sources too deep, but I guess that after we found that mime type is `image/*` the `nosniff` directive is not taken into account during the following subtype identification. This seems reasonable and does not contradict the nature of the `nosniff` security restriction.

I think the [original topic](https://forum.palemoon.org/viewtopic.php?f=37&t=16562) need to be unlocked or at least updated based on this PR.

PS: And this is UXP task, am I right?

JustOff:
A few more notes:
* It seems that Pale Moon currently does not support `nosniff` at all
* `nosniff` support was landed in Firefox 50 (see [bug 471020](https://bugzilla.mozilla.org/show_bug.cgi?id=471020))
* `nosniff` only applies to `script` and `style` types (see [bug 1302539](https://bugzilla.mozilla.org/show_bug.cgi?id=1302539), [Fetch](https://fetch.spec.whatwg.org/#x-content-type-options-header), [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options))

wolfbeast:
Actually surprised about that -- I was pretty sure `nosniff` support landed a long time ago in the Mozilla platform so I didn't even give it a second thought.

JustOff:
I took a try to backport  `X-Content-Type-Options: nosniff` based on bugs [471020](https://bugzilla.mozilla.org/show_bug.cgi?id=471020) and [1302539](https://bugzilla.mozilla.org/show_bug.cgi?id=1302539), but it seem I can't handle it properly. I can pass some [tests on w3c-test.org](http://w3c-test.org/fetch/nosniff/), but others result to the browser crash. Here is my [current working tree](https://github.com/JustOff/Pale-Moon/tree/PR_nosniff).  Probably we need a someone smarter to move it forward.

PR #1340, opened 2017-09-10
Remove unused pref layout.frame_rate.precise

JustOff:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1356751 and https://forum.palemoon.org/viewtopic.php?p=121476#p121476.

This should be applied to UXP too.


wolfbeast:
Removed from UXP in https://github.com/MoonchildProductions/moebius/commit/5d6198a9e2c44e50cffab1b422aa2154833f909e

PR #1346, opened 2017-09-13
Restore (updated) SSUAO for YouTube

trav90:
Otherwise on live streams, YouTube replaces the live chat box with a "Your browser is out of date" message.

I tested using all three user agent modes on current trunk build, but none worked. Using this SSUAO restores live chat functionality on YouTube.


wolfbeast:
Thanks

Issue #1292, opened 2017-08-15
JS - DOM - implement Object.values / Object.entries


janekptacijarabaci:
https://tc39.github.io/ecma262/2017/#sec-object.values

> 19.1.2.21Object.values ( O )

> When the values function is called with argument O, the following steps are taken:
> Let obj be ? ToObject(O).
> Let nameList be ? EnumerableOwnProperties(obj, "value").
> Return CreateArrayFromList(nameList).

JustOff:
Unfortunately the test failed: https://jsfiddle.net/3pq5pc9q/

janekptacijarabaci:
See #1293 
PropertyDescriptor.enumerable/configurable is not implemented.

wolfbeast:
So, we have a partial implementation then. It should still be OK to ship in 27.5 as-is, but this issue should remain open until the full spec is implemented.

JustOff:
Something really weird is going here. The test running directly in jsfiddle gives me an error from this test:
```
o = { a: 3, b: 2 };
values = Object.values(o);
assertDeepEq(values, [3, 2]);
```

`Error: Assertion failed: got "false", expected "true at _[0].[[Configurable]]`


But when I run it in the scratchpad first chunk is passed and I get completely different error from: 
```
var o = {get a() {delete this.b; return 1}, b: 2, c: 3};
values = Object.values(o);
assertDeepEq(values, [1, 3]);
```
`Error: got own properties ["0", "1", "2", "length"], expected ["0", "1", "length"]`

wolfbeast:
jsfiddle probably interferes.
Are you sure the test is correct?
If this is taken from the marionette/automated tests then you can't just apply it in most cases.

JustOff:
> jsfiddle probably interferes.

It runs well in both Firefox & Chrome.

Unfortunately I'm not so familiar with marionette ...

wolfbeast:
Firefox also converted this entire implementation to C++, so it's not the same code.

JustOff:
I checked Firefox nightlies from the range between [1208464](https://bugzilla.mozilla.org/show_bug.cgi?id=1208464) and [1232639](https://bugzilla.mozilla.org/show_bug.cgi?id=1232639) (js->c++) and they are also pass the test on jsfiddle.

janekptacijarabaci:
``` javascript
if ("values" in Object) {
  var o, values;
  o = { a: 3, b: 2 };
[...]

PR #1357, opened 2017-09-19
Fix Object.values / Object.entries self-hosted implementation

JustOff:
Follow-up to #1293, see issue #1292.

Until we implement the [bug 1170372](https://bugzilla.mozilla.org/show_bug.cgi?id=1170372) in our tree, we need to explicitly specify the attributes in the _DefineDataProperty call when constructing Object.values / Object.entries.

Test: https://jsfiddle.net/3pq5pc9q/

I suggest to uplift this to 27.5_RelBranch.


wolfbeast:
Great! Thank you for your research and the fix.

wolfbeast:
Uplifted in e82579e0985249f414eb90faaef6a121f1d885b0  705c5458d79c3f823acf807fc9df5d5578c1c7b4 and 2f467811368b2ba693be776e6e667e50331c324f

PR #1365, opened 2017-09-25
Expose extra network connection preferences in the UI

### LINKED ISSUE #1364: Expose SOCKS DNS lookup preference in the UI
SOCKS5 offers an option to perform DNS lookups through the proxy. This is used for e.g. TOR browsing and other closed networks that use SOCKS gateways with internal DNS names.

This option has been available for a very long time in Mozilla products, but has never been exposed in the UI. A simple addition to the proxy/connection window should take care of this. Of course it means string changes, so this will not be in any point release but will have to be published in a major release.

JustOff:
* Autologin to proxy if password is saved (`signon.autologin.proxy`), see bugs [#910670](https://bugzilla.mozilla.org/show_bug.cgi?id=910670), [#974394](https://bugzilla.mozilla.org/show_bug.cgi?id=974394)
* Use remote DNS with SOCKS5 (`network.proxy.socks_remote_dns`), see bugs [#665319](https://bugzilla.mozilla.org/show_bug.cgi?id=665319), [#1249494](https://bugzilla.mozilla.org/show_bug.cgi?id=1249494)

palemoon_2017-09-25_17-02-39

Resolves #1364.

### DIFF
--- a/browser/locales/en-US/chrome/browser/preferences/connection.dtd
+++ b/browser/locales/en-US/chrome/browser/preferences/connection.dtd
@@ -32,6 +32,8 @@
 <!ENTITY  socks4.accesskey              "K">
 <!ENTITY  socks5.label                  "SOCKS v5">
 <!ENTITY  socks5.accesskey              "v">
+<!ENTITY  socksRemoteDNS.label          "Send DNS queries through the proxy when using SOCKS v5">
+<!ENTITY  socksRemoteDNS.accesskey      "d">
 <!ENTITY  port.label                    "Port:">
 <!ENTITY  HTTPport.accesskey            "P">
 <!ENTITY  SSLport.accesskey             "o">
@@ -42,3 +44,6 @@
 <!ENTITY  noproxyExplain.label          "Example: .palemoon.org, .net.nz, 192.168.1.0/24">
 <!ENTITY  shareproxy.label              "Use this proxy server for all protocols">
 <!ENTITY  shareproxy.accesskey          "s">
+<!ENTITY  autologinproxy.label          "Do not prompt for authentication if password is saved">
+<!ENTITY  autologinproxy.accesskey      "i">
+<!ENTITY  autologinproxy.tooltip        "This option silently authenticates you to proxies when you have saved credentials for them. You will be prompted if authentication fails.">
--- a/browser/locales/en-US/chrome/browser/preferences/connection.dtd
+++ b/browser/locales/en-US/chrome/browser/preferences/connection.dtd
@@ -32,7 +32,7 @@
 <!ENTITY  socks4.accesskey              "K">
 <!ENTITY  socks5.label                  "SOCKS v5">
 <!ENTITY  socks5.accesskey              "v">
-<!ENTITY  socksRemoteDNS.label          "Send DNS queries through the proxy when using SOCKS v5">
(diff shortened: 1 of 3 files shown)


browser/locales/en-US/chrome/browser/preferences/connection.dtd
```diff
@@ -32,6 +32,8 @@
 <!ENTITY  socks4.accesskey              "K">
 <!ENTITY  socks5.label                  "SOCKS v5">
 <!ENTITY  socks5.accesskey              "v">
+<!ENTITY  socksRemoteDNS.label          "Send DNS queries through the proxy when using SOCKS v5">
+<!ENTITY  socksRemoteDNS.accesskey      "d">
```
wolfbeast: I don't think this is good wording.
Since the option gets disabled when SOCKS v5 isn't in use, it can be simplified to "Resolve hostnames through the proxy"

JustOff: You are right, that's important (btw, the patch itself took me a few minutes, while the invention of this name  - more then an hour). 

Firefox currently uses "Proxy DNS when using SOCKS v5" here, but this leads to problems with both correct understanding and localization (see [bug #1286701](https://bugzilla.mozilla.org/show_bug.cgi?id=1286701)). I checked [German](https://dxr.mozilla.org/l10n-central/source/de/browser/chrome/browser/preferences/connection.dtd), [French](https://dxr.mozilla.org/l10n-central/source/fr/browser/chrome/browser/preferences/connection.dtd), [Spanish](https://dxr.mozilla.org/l10n-central/source/es-ES/browser/chrome/browser/preferences/connection.dtd) and [Russian](https://dxr.mozilla.org/l10n-central/source/ru/browser/chrome/browser/preferences/connection.dtd) language packs and none of them contain direct translation. I'm afraid that similar problem will also arise with the verb "resolve" and, at the same time, I think we should keep mentioning SOCKS5 here to avoid confusion. 

I would like to find an option devoid of all these shortcomings, but I could not do anything other than what I suggested above. Maybe you have a better idea?

wolfbeast: There's many valid ways you can say this, but it's inherently going to be very technical if you're going to be in this dialog box.
I think the only other valid ways of properly saying this are (with some minor variations if needed):
 "Look up domain names via proxy (SOCKS v5 only)"
or "Perform name resolution via proxy (SOCKS v5 only)"
or "Use proxy to perform DNS lookups (SOCKS v5)"

JustOff: "Use proxy to perform DNS queries (SOCKS v5 only)" - is it ok to you?

wolfbeast: Sure

wolfbeast: Just check if it's not too long of a string.

JustOff: It fits well:

palemoon_2017-09-25_20-43-20

wolfbeast: Yup, works for me. Merging.

PR #1370, opened 2017-09-28
Clear site security settings as part of clearing Site Preferences

JustOff:
Clear `SiteSecurityServiceState.txt` when Site Preferences is selected.

Based on [bug 775370](https://bugzilla.mozilla.org/show_bug.cgi?id=775370), see the [report on forum](https://forum.palemoon.org/viewtopic.php?p=123000#p123000).

I think we should uplift this to 27.5.1.

### DIFF
--- a/browser/base/content/sanitize.js
+++ b/browser/base/content/sanitize.js
@@ -365,7 +365,7 @@ Sanitizer.prototype = {
         return true;
       }
     },
-    
+
     siteSettings: {
       clear: function ()
       {
@@ -373,12 +373,12 @@ Sanitizer.prototype = {
         var pm = Components.classes["@mozilla.org/permissionmanager;1"]
                            .getService(Components.interfaces.nsIPermissionManager);
         pm.removeAll();
-        
+
         // Clear site-specific settings like page-zoom level
         var cps = Components.classes["@mozilla.org/content-pref/service;1"]
                             .getService(Components.interfaces.nsIContentPrefService2);
         cps.removeAllDomains(null);
-        
+
         // Clear "Never remember passwords for this site", which is not handled by
         // the permission manager
         var pwmgr = Components.classes["@mozilla.org/login-manager;1"]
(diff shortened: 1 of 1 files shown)


wolfbeast:
This is wrong. nsISiteSecurityServices does not store site preferences or anything else to that effect. It stores HSTS/HPKP data which has nothing to do with the site's content!

JustOff:
Sorry, but no one tells that HSTS/HPKP data has something to do with site's **content** (we have separate Offline Website Data for that). And I think HSTS/HPKP controlled by SiteSecurityServices is exactly part of Site **Preferences**, isn't so?

wolfbeast:
**No, it's not a site preference.** It's not something that is controlled per site or by the site's content, nor is it something controlled by the user. It stores data about the domain's/site's secure connectivity state, not the domain's content. By clearing this you are effectively destroying the effect of using HSTS/HPKP in the first place, especially if you clear it for all sites.

If people want to security/privacy tradeoff for HSTS be in favor of privacy, they should disable HSTS. This should NOT be a side-effect of clearing this per-site data otherwise.

JustOff:
Well, let's take a look at [siteSettings](https://github.com/MoonchildProductions/Pale-Moon/blob/27.5_RelBranch/browser/base/content/sanitize.js#L369) section in `sanitize.js`.  It seem that SiteSecurityServices data cleanup perfectly fits right there. 

However, since you strongly disagree with this (I'm not likely to convince you, although I'm sure of the opposite), we probably should provide a different option to remove HSTS/HPKP data. What do you think about it?

JustOff:
One more thought, this is also how Moebius acts right now. Are you going to remove this call from its sanitize.js?

wolfbeast:
No, it does not fit there. all those things in sanitize.js (zoom level, storing passwords, pop-up settings) are per site settings _as a result of user interaction_, and have absolutely nothing to do with how you connect to the site which is handled by the networking layer and its related specs.

If you use HSTS/HPKP, this MUST be stored in the user's profile. If you do not want this to happen, disable the feature.

If this is how Möbius acts right now then that is also incorrect and needs to be changed.

wolfbeast:
if we need a way to clear this, then it has to be a separate category in the sanitize dialog. It's not that I won't be possible to convince of the utility of this and privacy benefits of it, it's that it's incorrect putting this under site preferences.

JustOff:
Ok, I understand your position now. I'm going to fill a different PR with a separate option to clear HSTS/HPKP data. Do you have suggestions on how to name this category?

wolfbeast:
"Site connectivity data" is the most accurate name for it.

PR #1371, opened 2017-09-28
Add option to clear site connectivity data (HSTS/HPKP)

JustOff:
Based on discussion in #1370.

### DIFF
--- a/browser/locales/en-US/chrome/browser/sanitize.dtd
+++ b/browser/locales/en-US/chrome/browser/sanitize.dtd
@@ -44,6 +44,8 @@ that require it.  -->
 <!ENTITY itemCache.accesskey               "A">
 <!ENTITY itemOfflineApps.label             "Offline Website Data">
 <!ENTITY itemOfflineApps.accesskey         "O">
+<!ENTITY itemConnectivityData.label        "Site Connectivity Data">
+<!ENTITY itemConnectivityData.accesskey    "N">
 <!ENTITY itemActiveLogins.label            "Active Logins">
 <!ENTITY itemActiveLogins.accesskey        "L">
 <!ENTITY itemSitePreferences.label         "Site Preferences">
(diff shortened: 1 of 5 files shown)


browser/locales/en-US/chrome/browser/sanitize.dtd
```diff
@@ -44,6 +44,8 @@ that require it.  -->
 <!ENTITY itemCache.accesskey               "A">
 <!ENTITY itemOfflineApps.label             "Offline Website Data">
 <!ENTITY itemOfflineApps.accesskey         "O">
+<!ENTITY itemConnectivityData.label        "Site Connectivity Data">
+<!ENTITY itemConnectivityData.accesskey    "N">
```
wolfbeast: Should be lower caps n to underline the proper letter.

JustOff: The capital "N" works well too (verified), like "A" for "Cache" four lines above. I just wanted to keep the overall style, but I can change it to lowercase if you insist.

JustOff: See also [XUL/Attribute/accesskey](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/accesskey):

> Although the value is case insensitive, a letter with the case matching the accesskey attribute will be used if both cases exist in the label. Otherwise, if the label has two or more characters which are same as its accesskey, the first of those characters in the label will be underlined.

So, if you don't mind, I'd keep capital "N" to preserve the file style consistent.

wolfbeast: Right. I've just had issues with underlining the wrong letter in the past so I assumed strict case sensitivity.

JustOff:
It seems this should be marked as UXP task (see [discussion](https://github.com/MoonchildProductions/Pale-Moon/pull/1370#issuecomment-332809901)).

wolfbeast:
Um no, because this isn't platform work. This is all front-end.
If you want to change this in Basilisk, then it would be a separate thing, but Basilisk already does this "the firefox way" AIUI

PR #1375, opened 2017-09-28
Revert "Allow disabling HW acceleration even when SharedDecoderManage…

trav90:
…r isn't used"

This reverts commit 67cb802 and restores the behavior of 27.4.2.

With light testing on Youtube this seems to resolve the higher than normal CPU usage when playing 60fps video (https://forum.palemoon.org/viewtopic.php?f=29&t=16797&p=123027)

### DIFF
--- a/dom/media/MediaFormatReader.cpp
+++ b/dom/media/MediaFormatReader.cpp
@@ -448,15 +448,11 @@ void
 MediaFormatReader::DisableHardwareAcceleration()
 {
   MOZ_ASSERT(OnTaskQueue());
-  if (HasVideo()) {
-    mPlatform->DisableHardwareAcceleration();
-    Flush(TrackInfo::kVideoTrack);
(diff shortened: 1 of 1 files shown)


wolfbeast:
So what was the initial commit for, exactly? Just so I know what backing this out has as effect.

wolfbeast:
Uplifted in 6f4513341219057a73f4f3df919b619b6a2e83e6

trav90:
This was originally committed to reduce our dependency on the SharedDecoderManager (which by Mozilla's own admission never should have landed in their tree as it has several issues with it).

I thought I had most of the media subsystem disconnected from the SDM (hence the original commit), but obviously it's still needed in some circumstances.

wolfbeast:
All right. Thanks for the info!

trav90:
I've pushed out a new unstable Linux build and posted in the forum thread linked above asking for verification from users there.

PR #1379, opened 2017-09-29
Restore previous YouTube SSUAO

trav90:
This essentially reverts commit b2fca2a.

This patch, together with commit 487060c results in approx 10-15% less CPU usage when watching 1080p60fps videos on YouTube in a new profile compared to current release on my system. This also restores the "old" YouTube design as well.

So it would seem that something with YouTube's new design is the primary reason for the increased CPU usage after all. :roll_eyes:


mattatobin:
The new youtube site seems to be marked up in some very strange non-html markup with HTML in template snipplets that is used to completely generate the effective rendered html using heavy javascript.. More than ever before. This JS is so heavy and the amount of effective code generated is locking the main thread pretty badly..

However, there isn't anything that can be done about it short of throwing js and layout engine onto other threads and async it all entirely I suspect... Which of course would be terrible to workout and would be crazy to implement.

I do know the huge cpu usage is on the main thread and caused by javascript either by its self processing or the process of generating all these insane google specific tags into effective html and rendering all at once. As I can have one video playing or any piece of media and Youtube's scripting in another tab and it won't affect the video issues with media.. BUT if it is playing and generating effective html on the same tab there is a hiccup but not much more.. I don't know why that is.. but That is what I have observed and could work out.

Frankly I dunno what we can do here except revert back to the older layout via UA as Travis has in this PR.

wolfbeast:
This is missing closing delimiters for the gaming.youtube.com SSUAO.

trav90:
Whoops, my bad. Sorry about that.

PR #1388, opened 2017-10-03
Status bar (s4e) code cleanup

JustOff:
* Remove unused and obsolete code
* ~~Hide~~ Remove debug messages ~~behind `#ifdef MOZ_DEBUG`~~


browser/components/statusbar/Status4Evar.jsm
```diff
@@ -70,11 +76,8 @@ Status4Evar.prototype =
 	destroy: function()
 	{
 		this._window.removeEventListener("unload", this, false);
-		if(Services.vc.compare("28.*", Services.appinfo.version) >= 0)
```
mattatobin: vc will return -1 on Pale Moon 27.* thus the condition is false.. You may want to check and ensure the inner code now exposed doesn't have any consequences.

browser/components/statusbar/Status4Evar.jsm
```diff
@@ -11,7 +11,9 @@ const CI = Components.interfaces;
 const CU = Components.utils;
 
 const s4e_service = CC["@caligonstudios.com/status4evar;1"].getService(CI.nsIStatus4Evar);
+#ifdef MOZ_DEBUG
```
mattatobin: Does anyone build Debug builds? May want not preprocess it out.. Same goes with all MOZ_DEBUG preprocessor conditionals you added.

Could create a preference like `status4evar.debug` and just allow these debug deemed items be based on state of the preference. Also I am not sure all of them are appropriate for debug status.

JustOff: While working on the [Location-4-Evar](https://github.com/JustOff/L4E) add-on, which revives the ability to display status and progress indicators in the location bar, I think I studied the s4e code deep enough to say that these debug messages are completely useless at the moment. Thus, providing an extra preference, which most likely will never be used, does not seem like a good idea. 

Actually, we can safely remove all this code, I put it under `#ifdef`, just in case, if the future browser development will require to make changes to this module and debug them. If you know a better way to keep these pieces of code in the sources, but do not include them in public releases, then I'd be glad to take advantage of it.

wolfbeast: It's usually a bad idea to require a rebuild of a debug build to debug front-end code. So I have to agree with @mattatobin here: reading a (hidden) pref on init, storing it in a var and testing against that var where you've now put MOZ_DEBUG `#ifdef`s is the way to go. Which debug messages are actually useful for development or not can be looked at later on.

JustOff: Well, l will add a hidden preference as you requested, but I'm still absolutely sure that these debug messages are completely useless and this preference most likely will never be used at all, so they could be safely removed right now.

wolfbeast: If you're certain, then just remove the conditional code.

JustOff: Yes, I'm sure. @mattatobin, if you have no objection, I'm going to remove all this debug code.

mattatobin: That works for me.

mattatobin:
Checked all conditions of the Download Status widget and all is a-ok.

I am happy with this.

PR #1393, opened 2017-10-04
Display "not supported" error when trying to install WebExtension

### LINKED ISSUE #1391: XPIProvider detects WebExtensions as corrupt
It would be nice if XPIProvider could detect `manifest.json` (without `install.rdf`) as a WebExtension and display an error other than the add-on is corrupt.

Pinging @wsdfhjxc because he did the original jetpack detection.

The idea is that XPI files with only `manifest.json` but not `install.rdf ` should display an error indicating that [Application] does not support WebExtensions.

However, those with install.rdf, the hybrid/transitional extensions, should still be allowed to install as their compatibility falls into the _maybe works_ category where other Firefox-only targeted extensions live at this point.

JustOff:
This resolves #1391.

### DIFF
--- a/toolkit/mozapps/extensions/internal/XPIProvider.jsm
+++ b/toolkit/mozapps/extensions/internal/XPIProvider.jsm
@@ -1071,6 +1071,7 @@ function loadManifestFromZipReader(aZipReader) {
     zis = aZipReader.getInputStream(FILE_INSTALL_MANIFEST);
   } catch (e) {
     let zws = aZipReader.getInputStream(FILE_WEBEXT_MANIFEST);
+    zws.close();
     throw {
       name: e.name,
       message: e.message,
--- a/toolkit/mozapps/extensions/internal/XPIProvider.jsm
+++ b/toolkit/mozapps/extensions/internal/XPIProvider.jsm
@@ -106,6 +106,7 @@ const DIR_TRASH                       = "trash";
 const FILE_DATABASE                   = "extensions.json";
 const FILE_OLD_CACHE                  = "extensions.cache";
 const FILE_INSTALL_MANIFEST           = "install.rdf";
+const FILE_WEBEXT_MANIFEST            = "manifest.json";
 const FILE_XPI_ADDONS_LIST            = "extensions.ini";
 
 const KEY_PROFILEDIR                  = "ProfD";
@@ -1065,7 +1066,17 @@ function loadManifestFromDir(aDir) {
  * @throws if the XPI file does not contain a valid install manifest
  */
 function loadManifestFromZipReader(aZipReader) {
-  let zis = aZipReader.getInputStream(FILE_INSTALL_MANIFEST);
+  let zis;
+  try {
+    zis = aZipReader.getInputStream(FILE_INSTALL_MANIFEST);
+  } catch (e) {
+    let zws = aZipReader.getInputStream(FILE_WEBEXT_MANIFEST);
+    throw {
+      name: e.name,
+      message: e.message,
+      webext: true
+    };
+  }
   let bis = Cc["@mozilla.org/network/buffered-input-stream;1"].
             createInstance(Ci.nsIBufferedInputStream);
   bis.init(zis, 4096);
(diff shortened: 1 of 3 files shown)


mattatobin:
Does this allow for the compatibility limbo state of a hybrid extension that has both install.rdf and manifest.json?

JustOff:
Yes, hybrid extensions were taken into account.

mattatobin:
Thank you for taking this up, @JustOff .. I'll do a quick build and verify.

mattatobin:
Cursory testing shows no unexpected issues..

r+

toolkit/mozapps/extensions/internal/XPIProvider.jsm
```diff
@@ -1065,7 +1066,17 @@ function loadManifestFromDir(aDir) {
  * @throws if the XPI file does not contain a valid install manifest
  */
 function loadManifestFromZipReader(aZipReader) {
-  let zis = aZipReader.getInputStream(FILE_INSTALL_MANIFEST);
+  let zis;
+  try {
+    zis = aZipReader.getInputStream(FILE_INSTALL_MANIFEST);
+  } catch (e) {
+    let zws = aZipReader.getInputStream(FILE_WEBEXT_MANIFEST);
+    throw {
```
wolfbeast: I find the logic here a little odd. the catch(e) {} block will be aborted if no webextension manifest is found (`getInputStream` will abort the entire code block on failure) and won't throw in that case, while it should, according to the function header. So actually corrupt XPIs would give undefined behavior here, IIUC.
I guess you'd need a nested try/catch here, for when there's no install manifest, to throw either with webext true or false, depending?

wolfbeast: To clarify: the problem here is that a failure on zip reader no longer throws because you're not having a zipReader failure in the top-level of the function any more -- only the catch() block will be aborted and the function itself will not throw.

JustOff: The logic is the following:

* If `install.rdf` exists, then all is ok
* If `install.rdf` does not exists, then `NS_ERROR_FILE_TARGET_DOES_NOT_EXIST` is thrown and we go to the catch block to check for `manifest.json`
* If `manifest.json` does not exist, then `NS_ERROR_FILE_TARGET_DOES_NOT_EXIST` is thrown and `loadManifestFromZipReader` function immediately returns with this error
* If `manifest.json` exist, then we manually throw custom error with extra property `webext` set to `true`

I.e., if `install.rdf` does not exists, then an error is always thrown and we can use `webext` property select appropriate message to display. I verified this works as expected in all cases.

wolfbeast: > If manifest.json does not exist, then NS_ERROR_FILE_TARGET_DOES_NOT_EXIST is thrown and loadManifestFromZipReader function immediately returns with this error

No, it doesn't. It shouldn't! If an error occurs in a scope it will only abort that scope, not its parent. if the scope isn't a function it doesn't throw from the function.
Where the error is thrown you see and it exits is further in when there happens to be another attempted opening of the manifest file: `let uri = buildJarURI(aZipReader.file, FILE_INSTALL_MANIFEST);` so it's only by chance of another attempted read that the function actually throws because that `try{}` block doesn't have a `catch{}` (only has a `finally{}`)

So functionally, yes, it works this way, but it's not actually correct and could do with a rewrite. We can leave it for now and I can re-visit this when I have some time to redo this without the implicit fall-through throwing.

JustOff: You are wrong, javascript errors abort all scopes up to the nearest try/catch block or system handler. I would bet a big box of beer, but it will be unfair, because I just verified this using stack trace.

wolfbeast: I'll take your word for it. :)
I guess the issue for me is that it's probably re-throwing in the catch block, in which case you're absolutely correct, but that that behavior is completely opaque (since zipReader is C++/IDL) so it's not clear in this code whether the flow will be correct or not. If zipReader ever changes behavior and no longer throws, this will break.
So, as said, it's relying on implied behavior and could do with clearer, more explicit code.

Issue #1410, opened 2017-10-13
Implement DOM methods: prepend(), append(), before(), after() and replaceWith()


janekptacijarabaci:
https://dom.spec.whatwg.org/

> The prepend(nodes) method, when invoked, must run these steps:
>
> Let node be the result of converting nodes into a node given nodes and context object’s node document.
> Pre-insert node into context object before the context object’s first child.

> The append(nodes) method, when invoked, must run these steps:
>
> Let node be the result of converting nodes into a node given nodes and context object’s node document.
> Append node to context object.

> The before(nodes) method

> The after(nodes) method

> The replaceWith(nodes) method

wolfbeast:
Actually, this isn't part of the DOM4 spec (which is a W3C spec that does NOT include these functions).
Specs are in conflict, so websites shouldn't implicitly use this.

wolfbeast:
Verified with lolcatz.

wolfbeast:
Re-opened because of build bustage on gcc/linux. See discussion on PR #1411.

Issue #1409, opened 2017-10-13
http://icanhas.cheezburger.com/lolcats won't display certain images.


MegamanXtreme:
Forum thread in question:
https://forum.palemoon.org/viewtopic.php?p=124491

janekptacijarabaci:
Pushlog
https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=9ce31e9f90cb0e534611b0f617c5bbc232ffe748&tochange=ab0044bfa1df858919797bcd6a9aef76a668cd4a

wolfbeast:
Verified fixed with #1410 by merging #1411

PR #1411, opened 2017-10-13
JS - DOM - node - implement DOM4 methods: prepend(), append(), before(), after() and replaceWith() (partially)

janekptacijarabaci:
Ad #1410 (#1409)

See:

https://bugzilla.mozilla.org/show_bug.cgi?id=911477
https://bugzilla.mozilla.org/show_bug.cgi?id=1301777

https://bugzilla.mozilla.org/show_bug.cgi?id=1308922 (partially - follow up (?)):

__Are these changes correct?__
https://github.com/MoonchildProductions/Pale-Moon/commit/afaa122df5a839ccf1e5c6cd8f6ce548c35fc741#diff-9dc8f8c6d5d189b71696d2f9a06188d4L3365
https://github.com/MoonchildProductions/Pale-Moon/commit/afaa122df5a839ccf1e5c6cd8f6ce548c35fc741#diff-9dc8f8c6d5d189b71696d2f9a06188d4L3372

https://bugzilla.mozilla.org/show_bug.cgi?id=1104955 (partially - without `Symbol.unscopables`)

\+`RefPtr` => `nsRefPtr`

---

I've created the new build (x32, Windows) and tested:

https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/prepend#Examples
https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/append#Examples
https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/before#Examples
https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/after#Examples
https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/replaceWith#Examples

### DIFF
--- a/dom/base/Element.cpp
+++ b/dom/base/Element.cpp
@@ -3390,7 +3390,8 @@ Element::InsertAdjacent(const nsAString& aWhere,
     }
     parent->InsertBefore(*aNode, this, aError);
   } else if (aWhere.LowerCaseEqualsLiteral("afterbegin")) {
-    static_cast<nsINode*>(this)->InsertBefore(*aNode, GetFirstChild(), aError);
+    nsCOMPtr<nsINode> refNode = GetFirstChild();
+    static_cast<nsINode*>(this)->InsertBefore(*aNode, refNode, aError);
   } else if (aWhere.LowerCaseEqualsLiteral("beforeend")) {
     static_cast<nsINode*>(this)->AppendChild(*aNode, aError);
   } else if (aWhere.LowerCaseEqualsLiteral("afterend")) {
@@ -3398,7 +3399,8 @@ Element::InsertAdjacent(const nsAString& aWhere,
     if (!parent) {
       return nullptr;
     }
-    parent->InsertBefore(*aNode, GetNextSibling(), aError);
+    nsCOMPtr<nsINode> refNode = GetNextSibling();
+    parent->InsertBefore(*aNode, refNode, aError);
   } else {
     aError.Throw(NS_ERROR_DOM_SYNTAX_ERR);
     return nullptr;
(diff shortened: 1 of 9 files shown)


trav90:
@janekptacijarabaci looks like this causes build bustage on Linux with GCC 4.9:

https://gist.github.com/trav90/917722e79b9df095a5f714214f132742

janekptacijarabaci:
@trav90 
Yes, it is possible. But I do not build on Linux. Can you try to fix it please?

janekptacijarabaci:
@trav90 
Maybe see also: https://github.com/janekptacijarabaci/Pale-Moon/commit/68f397a5677c91d7831994d2d539859cd3c0b5f1
On Windows it works as expected.

wolfbeast:
I'm gathering thet `getAsNode()` doesn't have a provision to return a pointer type and will therefore return an object type. MSVC is obviously smart enough to convert this implicitly, but I guess gcc needs an explicit typecast.

Does gcc like the following?:
```C++
static already_AddRefed<nsINode>
GetNodeFromNodeOrString(const OwningNodeOrString& aNode,
                        nsIDocument* aDocument)
{
  if (aNode.IsNode()) {
    nsCOMPtr<nsINode> node = static_cast<nsCOMPtr<nsINode>>(aNode.GetAsNode());
    return node.forget();
  }

  if (aNode.IsString()){
    nsRefPtr<nsTextNode> textNode =
      aDocument->CreateTextNode(aNode.GetAsString());
    return textNode.forget();
  }

  MOZ_CRASH("Impossible type");
}
```

Specifically, I added a cast: `static_cast<nsCOMPtr<nsINode>>(aNode.GetAsNode());` instead of the direct assignment.

trav90:
@wolfbeast looks like you were right. I've submitted #1415 to resolve this.

Issue #877, opened 2017-02-06
Add RequireInteraction flag to Desktop Notifications


wolfbeast:
This is part of the spec:
https://notifications.spec.whatwg.org/#require-interaction-preference-flag

Currently we don't support this and notifications are always removed after a specific amount of time.

BMO bug: https://bugzilla.mozilla.org/show_bug.cgi?id=862395

wolfbeast:
We'll have this solved in a UXP-based Pale Moon. Closing.

PR #1420, opened 2017-10-21
[suggestion] Notifications - user settings - the immediate duration of the alert

janekptacijarabaci:
See (if not planned in the foreseeable future):
https://github.com/MoonchildProductions/Pale-Moon/issues/877

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=1259108

---


wolfbeast:
Sure, make it configurable works in my book.

PR #1431, opened 2017-10-28
Improve timing of canplay and canplaythrough firing

### LINKED ISSUE #1430: canplay and canplaythrough event are never fired when media element is paused.
This seems to be the cause for some iterations of JWPlayer not working (or with specific configurations, like loading in paused state).

See https://forum.palemoon.org/viewtopic.php?f=3&t=17109

@trav90 could you have a look at implementing this soon? I should push 27.6 out in not too much time now AUS has been moved over to the new infra, and it would be great to have this fixed.

trav90:
This set of commits should resolve #1430.

I have built and tested this on the Gabien site. Videos now play, but the first couple seconds of audio "stutters" and doesn't play correctly. However, after the first couple seconds the stream begins playing fine. So there's still an issue here (likely a race condition as mentioned in #1430), but the streams are at least playable now. Baby steps!

I've also (very) lightly tested this on YouTube with no issues seen. I'll do some more extensive testing of this tomorrow, but in the meantime it should be good to ship out in an unstable build.

### DIFF
--- a/dom/html/HTMLMediaElement.cpp
+++ b/dom/html/HTMLMediaElement.cpp
@@ -3546,7 +3546,12 @@ HTMLMediaElement::UpdateReadyStateInternal()
     MetadataLoaded(&mediaInfo, nsAutoPtr<const MetadataTags>(nullptr));
   }
 
-  if (NextFrameStatus() == MediaDecoderOwner::NEXT_FRAME_UNAVAILABLE_SEEKING) {
+  enum NextFrameStatus nextFrameStatus = NextFrameStatus();
+  if (mDecoder && nextFrameStatus == NEXT_FRAME_UNAVAILABLE) {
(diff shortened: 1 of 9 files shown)


trav90:
I have more properly tested this now on YouTube, Twitch, Grabien, and Vimeo. All looks good!

wolfbeast:
Okay, merging this in for shark-infested waters ;)

janekptacijarabaci:
It works for me, but `Use MSE asynchronously` setting must not be used.

trav90:
And what's the results with async MSE enabled?

The majority of my testing (including the results I listed above) was with default video settings.

janekptacijarabaci:
> what's the results with async MSE enabled?
The situation is the same as in PM 27.5.1 (e.g.).

It is strange, but sometimes it doesn't. Sometimes later I will try...

PR #1440, opened 2017-11-01
Handle uninstalls of restartless addons in XPIProvider

JustOff:
Based on [bug 612168](https://bugzilla.mozilla.org/show_bug.cgi?id=612168). 

This fixes the issue that  restartless add-ons were never received the `ADDON_UNINSTALL` reason on uninstall, but `ADDON_DISABLE` (see [bug 620541](https://bugzilla.mozilla.org/show_bug.cgi?id=620541)).

### DIFF
--- a/toolkit/mozapps/extensions/content/extensions.js
+++ b/toolkit/mozapps/extensions/content/extensions.js
@@ -1651,7 +1651,7 @@ function doPendingUninstalls(aListBox) {
   var listitem = aListBox.firstChild;
   while (listitem) {
     if (listitem.getAttribute("pending") == "uninstall" &&
-        !(listitem.opRequiresRestart("UNINSTALL")))
+        !(listitem.opRequiresRestart("uninstall")))
       items.push(listitem.mAddon);
     listitem = listitem.nextSibling;
   }
--- a/toolkit/mozapps/extensions/content/extensions.js
+++ b/toolkit/mozapps/extensions/content/extensions.js
@@ -1651,7 +1651,7 @@ function doPendingUninstalls(aListBox) {
   var listitem = aListBox.firstChild;
   while (listitem) {
     if (listitem.getAttribute("pending") == "uninstall" &&
-        !listitem.isPending("uninstall"))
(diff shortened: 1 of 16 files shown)


toolkit/mozapps/extensions/content/extensions.js
```diff
@@ -1651,7 +1651,7 @@ function doPendingUninstalls(aListBox) {
   var listitem = aListBox.firstChild;
   while (listitem) {
     if (listitem.getAttribute("pending") == "uninstall" &&
-        !listitem.isPending("uninstall"))
+        !(listitem.opRequiresRestart("UNINSTALL")))
```
wolfbeast: I don't think this should be all-caps, but lowercase instead.

JustOff:
Yes, of course.

toolkit/mozapps/extensions/content/extensions.js
JustOff: Are you sure? See also https://github.com/MoonchildProductions/moebius/blob/master/toolkit/mozapps/extensions/content/extensions.js#L1793 - UNINSTALL is all-caps.

JustOff: Actually it does not matter because of [this](http://xref.palemoon.org/palemoon-trunk/source/toolkit/mozapps/extensions/content/extensions.xml#780):

```
       <method name="opRequiresRestart">
         <parameter name="aOperation"/>
         <body><![CDATA[
           var operation = AddonManager["OP_NEEDS_RESTART_" + aOperation.toUpperCase()];
           return !!(this.mAddon.operationsRequiringRestart & operation);
         ]]></body>
       </method>
```

wolfbeast: If it doesn't matter, then at least be consistent, as in other places it's used lowercase.

As foe Möbius: Mozilla obvious patched it like in the bugzilla bug, so inconsistent. Improvement there would be to also change it to lowercase.

JustOff: Ok, done.

PR #1463, opened 2017-11-13
Ensure we process FIN flags on all-padding final frames.

### LINKED ISSUE #1462: Youtube Live video stops after about 5-10 minutes
This issue by its nature is hard to reproduce as well as verify when it fixed, but I think we should apply the solution from [bug 1381016](https://bugzilla.mozilla.org/show_bug.cgi?id=1381016) to our tree. It's about http2 sessions handling, more details are [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1381016#c36).

The problem was reported [on forum](https://forum.palemoon.org/viewtopic.php?f=12&t=17329), I also experience a similar problem from time to time.

JustOff:
This should prevent http2 sessions freezing and resolve #1462.

### DIFF
--- a/netwerk/protocol/http/Http2Session.cpp
+++ b/netwerk/protocol/http/Http2Session.cpp
@@ -2771,7 +2771,10 @@ Http2Session::WriteSegments(nsAHttpSegmentWriter *writer,
     LOG3(("Http2Session::WriteSegments %p trying to discard %d bytes of data",
           this, count));
 
-    if (!count) {
+    if (!count && mDownstreamState == DISCARDING_DATA_FRAME) {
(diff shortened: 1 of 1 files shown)


janekptacijarabaci:
\+ `UXP task` ?

JustOff:
[Yes](http://xref.palemoon.org/moebius-trunk/source/netwerk/protocol/http/Http2Session.cpp#2891).

JustOff:
With this patch applied, I watch snooker live stream from YouTube for more than hour for now without any freezes. Previously this stream always hang for me within first 15-20 minutes. Probably you should consider to uplift to 27.6 as well.

Issue #1464, opened 2017-11-14
Program returns random data in a function (no-return-in-nonvoid-function)


computersalat:
build on Open Build Service fails cause of:

```
[ 1581s] I: Program returns random data in a function
[ 1581s] E: palemoon no-return-in-nonvoid-function /home/abuild/rpmbuild/BUILD/Pale-Moon-27.6.0_Release/dom/base/nsGlobalWindow.cpp:5696, 5713, 5730, 5747
```

wolfbeast:
OpenBuildService should focus on building and not on static code analysis (which is obviously naive and can't even understand the use of macros)

http://xref.palemoon.org/palemoon-trunk/source/dom/base/nsGlobalWindow.cpp#328

Macro always returns something.

computersalat:
@wolfbeast 
just wondering why someone is preparing a patch and referencing this issue when OBS static code checks produces bullshit warning ...

mattatobin:
My question is why anyone is trying to fix nonbusting nonissue compiler warnings that ONLY obs trips on for a codebase that only has what.. 6 months of life left to it?

wolfbeast:
If OBS treats warnings as errors, then that is OBS's fault. It shouldn't.
Also, it's a warning that has no grounds. If it warns on these functions, then it should warn on all `FORWARD_TO_OUTER_OR_THROW` macro calls where the call is in a non-void function, but it doesn't.

The patch, in fact, also doesn't do anything useful. the return values there are never reached because the macro either throws or returns a value in all cases.

@mattatobin the intended code base life isn't really relevant for a minor fix, but it isn't even a fix.

wolfbeast:
I've merged it in anyway - it doesn't do anything but it also won't hurt in case the macro has unexpected quirks based on compiler in use (e.g. not exiting on a throw properly)

computersalat:
@wolfbeast 
Thank you ... that patch removes mentioned complaints

PR #1362, opened 2017-09-23
JS - DOM - the vertical text - scrolling with middle button

janekptacijarabaci:
Ad https://forum.palemoon.org/viewtopic.php?f=29&t=16769

> Middle Click doesn't show scroll icon on vertical text page.

An example:
http://www.asahi-net.or.jp/~sd5a-ucd/freefonts/Oradano-Mincho/

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=962249
https://bugzilla.mozilla.org/show_bug.cgi?id=1219243

---

You add the label `Verification Needed`, please.

---

I've created the new build (x32, Windows) and tested:
- An horizontal / vertical page
- `DevTools` - `Developer Toolbar` - `screenshot`

PR #1465, opened 2017-11-14
[minor fix] Build - warnings

janekptacijarabaci:
Ad #1464 

Build - warnings:

```
    "[drive]:\\[path]\\dom\\base\\nsglobalwindow.cpp": {
      "hash": "4ff1292eb92f5a6bd9ada15acfbc242033b63cc3", 
      "warnings": [
        {
          "column": null, 
          "filename": "[drive]:\\[path]\\dom\\base\\nsglobalwindow.cpp", 
          "flag": "C4715", 
          "line": 5696, 
          "message": "'nsGlobalWindow::GetScrollMinX' : not all control paths return a value"
        }, 
        {
          "column": null, 
          "filename": "[drive]:\\[path]\\dom\\base\\nsglobalwindow.cpp", 
          "flag": "C4715", 
          "line": 5713, 
          "message": "'nsGlobalWindow::GetScrollMinY' : not all control paths return a value"
        }, 
        {
          "column": null, 
          "filename": "[drive]:\\[path]\\dom\\base\\nsglobalwindow.cpp", 
          "flag": "C4715", 
          "line": 5730, 
          "message": "'nsGlobalWindow::GetScrollMaxX' : not all control paths return a value"
        }, 
        {
          "column": null, 
          "filename": "[drive]:\\[path]\\dom\\base\\nsglobalwindow.cpp", 
          "flag": "C4715", 
          "line": 5747, 
          "message": "'nsGlobalWindow::GetScrollMaxY' : not all control paths return a value"
        }
      ]
    }, 
```

Follow up for #1362

---

### DIFF
--- a/dom/base/nsGlobalWindow.cpp
+++ b/dom/base/nsGlobalWindow.cpp
@@ -5692,7 +5692,9 @@ int32_t
 nsGlobalWindow::GetScrollMinX(ErrorResult& aError)
 {
   MOZ_ASSERT(IsInnerWindow());
-  FORWARD_TO_OUTER_OR_THROW(GetScrollBoundaryOuter, (eSideLeft), aError, 0);
+  int32_t scrollMinX = 0;
+  FORWARD_TO_OUTER_OR_THROW(GetScrollBoundaryOuter, (eSideLeft), aError, scrollMinX);
+  return scrollMinX;
 }
 
 NS_IMETHODIMP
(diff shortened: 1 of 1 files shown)


wolfbeast:
Seriously, this is completely unnecessary, and the "fix" doesn't even do anything useful at all, apart from satisfying broken static analysis... You just wasted your time.
If you "fix" this then you need to "fix" any and all occurrences of `FORWARD_TO_OUTER_OR_THROW` in non-void functions, which magically don't throw this warning.

janekptacijarabaci:
> If you "fix" this then you need to "fix" any and all occurrences of FORWARD_TO_OUTER_OR_THROW in non-void functions, which magically don't emit this warning in the same situation.

IMHO: There are no more.

wolfbeast:
I think you misunderstood.
Check the code. there are plenty of non-void functions that use that macro. Just because they (for some reason) don't trip up the compiler to toss out this warning doesn't mean the structure isn't exactly the same...

http://xref.palemoon.org/palemoon-trunk/search?string=FORWARD_TO_OUTER_OR_THROW

wolfbeast:
And.. I checked again today and there aren't. Okay this is weird. There were several non-void functions that ONLY had the macro call in them, but I can't find them now.

wolfbeast:
I'll just merge it in. dealing with continued discussion is likely going to take more than just adding some code to work around a compiler warning.

Issue #1466, opened 2017-11-14
DevTools - gcli - screenshot - the vertical mode


janekptacijarabaci:
__Steps to reproduce:__
Go to: http://www.asahi-net.or.jp/~sd5a-ucd/freefonts/Oradano-Mincho/
(CSS: `writing-mode: vertical-rl`)
Open: `Tools - Web Developer - Developer Toolbar`
Insert this command: `screenshot --fullpage`

__Actual results:__
The screenshot is cut off (+ the black color).

__Expected results:__
The screenshot is complete.

__Affected versions:__
all (i.e. Firefox trunk; I'm creating an ticket - afterwards...)

wolfbeast:
Merged in.

PR #1467, opened 2017-11-14
DevTools - gcli - screenshot - the vertical mode

janekptacijarabaci:
Ad #1466

I have no better idea...

---

You add the label `Devtools`, please.

---


wolfbeast:
This solution suits me fine.

Issue #1343, opened 2017-09-11
Consider implementing X-Content-Type-Options: nosniff


JustOff:
* Spec: https://fetch.spec.whatwg.org/#x-content-type-options-header.
* Tests: http://w3c-test.org/fetch/nosniff/

This feature has been landed in Firefox 50 (see bugs [471020](https://bugzilla.mozilla.org/show_bug.cgi?id=471020), [1302539](https://bugzilla.mozilla.org/show_bug.cgi?id=1302539) and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options)).

JustOff:
I [have tried](https://github.com/MoonchildProductions/Pale-Moon/pull/1339#issuecomment-328374089) to backport the bugs mentioned above, but have not achieved success. 

With [this patches](https://github.com/JustOff/Pale-Moon/commit/c09653208257eb287a8c35be34fe743fe2491a16) I can pass _image.html_ and _stylesheet.html_ tests, but _parsing-nosniff.html_ and _script.html_ give timeouts, while _importscripts.html_ and _worker.html_ lead to the browser crash. 

It seems the verification part works fine, because when I remove `return rv;` [here](https://github.com/JustOff/Pale-Moon/blob/c09653208257eb287a8c35be34fe743fe2491a16/netwerk/protocol/http/nsHttpChannel.cpp#L962), all errors are displayed in console at the right moment. Something wrong is going when I try to actually exit with `return`.

Does anyone have any idea what's wrong and how to fix it?

JustOff:
I made progress with this task [moving](https://github.com/JustOff/Pale-Moon/commit/b611bb944d449c095b979e6c6d762208906e2cb5) `ProcessXCTO()` call to `ProcessResponse()`, like it was initially proposed [here](https://bugzilla.mozilla.org/attachment.cgi?id=8770524&action=diff#a/netwerk/protocol/http/nsHttpChannel.cpp_sec3). This results to no more timeouts and crashes occur when running [tests](http://w3c-test.org/fetch/nosniff/), but I still don't understand why they happen when `ProcessXCTO()` called from `CallOnStartRequest()`. Probably this is because of some other changes in the platform code that were performed during last two years, but I can't figure exactly which ones were the cause.

Now `XCTO: nosniff` works as it should and all [tests](http://w3c-test.org/fetch/nosniff/) are successful, except for two minor flaws:

[...]

PR #1478, opened 2017-11-17
Change nsIDocumentLoaderFactory and nsIURIContentListener to take MIME types as an XPCOM string, not a char*

JustOff:
This increases overall code consistency and is a prerequisite for adding `text/json` and `text/vtt` to the known MIME types (PR will follow), which is a prerequisite for #1343.

Based on [bug 1136388](https://bugzilla.mozilla.org/show_bug.cgi?id=1136388).

Clobber is required (?) for IDL changes (not included in PR).

### DIFF
--- a/docshell/base/nsDSURIContentListener.cpp
+++ b/docshell/base/nsDSURIContentListener.cpp
@@ -87,7 +87,7 @@ nsDSURIContentListener::OnStartURIOpen(nsIURI* aURI, bool* aAbortOpen)
 }
 
 NS_IMETHODIMP
-nsDSURIContentListener::DoContent(const char* aContentType,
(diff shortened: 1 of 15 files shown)


wolfbeast:
Since you are bumping the IDL GUID, a clobber would normally not be required?

JustOff:
I'm not sure, probably I don't fully understand how clobber works.

wolfbeast:
I'll merge it in and build. If a clobber is needed after all I'll touch `CLOBBER` to force it.

wolfbeast:
Doesn't look like a clobber was needed, indeed. Usually it's only needed if IDL or WebIDL changes in a way that doesn't identify it as a different interface. A GUID change will always mark it as new.

JustOff:
Thanks, I'll try to remember this.

PR #1484, opened 2017-11-19
Handle custom URLs and add "My home page" for the New Tab URL.

wolfbeast:
Instead of dealing with the pref directly in the drop-down, we use an index value.

Upon loading the pref pane, we check against known values in `browser.newtab.url` and set the index accordingly. If no known value is found, the index will be set to `0` and `A custom URL` is shown (normally hidden in the drop-down) to indicate the user has manually set a custom URL for the New Tab page.

When the preferences dialog is closed with OK (accepted), the index value is read, and `browser.newtab.url` will be set to known values for the appropriate index value or remain unchanged in the case of `0`.

Special handling when the user chooses "my home page":
1. Read the browser's home page string
2. In case of multiple entries separated by pipes `|`, only the first URL will be used.
3. The sanitized value will be saved in `browser.newtab.myhome` to check against when determining the proper index.
4. The sanitized value will be saved in `browser.newtab.url`

### DIFF
--- a/browser/components/preferences/preferences.xul
+++ b/browser/components/preferences/preferences.xul
@@ -55,7 +55,9 @@
             style="&prefWinMinSize.styleGNOME;"
 #endif
 #endif
-            onunload="if (typeof gSecurityPane != 'undefined') gSecurityPane.syncAddonSecurityLevel();">
+            onunload="if (typeof gSecurityPane != 'undefined') gSecurityPane.syncAddonSecurityLevel();"
+            ondialogaccept="if (typeof gTabsPane != 'undefined') gTabsPane.writeNewtabUrl();
+                            return true;">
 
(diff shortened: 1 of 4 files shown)


wolfbeast:
@janekptacijarabaci @JustOff -- does this look sane to you? I've already verified proper operation.

janekptacijarabaci:
I do not see a problem - at first sight.

JustOff:
Looks pretty good, but it would be nice to have an interface to edit `browser.newtab.url` directly.

wolfbeast:
@JustOff I think that would go beyond what would generally be needed; most people editing it manually have done so because they wanted their home page (which is not solved) -- and it's in Pale Moon Commander for the people who wish it ;)

wolfbeast:
Thanks for having a check-over. merging.

PR #1489, opened 2017-11-21
Skip `XCTO: nosniff` check in case of 30x response

### LINKED ISSUE #1488: strange error message in the error console
Since the latest unstable i get quite a lot of these here in the error console:

Error: The resource from “whatever” was blocked due to MIME type mismatch (X-Content-Type-Options: nosniff).

For example the page 

http://pamelageller.com/

was shown before that new unstable without issues - now it looks really destroyed.

JustOff:
Addition to #1343, this should resolve #1488.

### DIFF
--- a/netwerk/protocol/http/nsHttpChannel.cpp
+++ b/netwerk/protocol/http/nsHttpChannel.cpp
@@ -163,6 +163,13 @@ WillRedirect(const nsHttpResponseHead * response)
            response->PeekHeader(nsHttp::Location);
 }
 
+// This is required to bypass XCTO: nosniff check
+bool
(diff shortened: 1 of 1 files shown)


JustOff:
I checked this without https://github.com/MoonchildProductions/Pale-Moon/commit/8167b87bc3f0cb68e53dea50a2bd990d6115cab0 applied which I think should be reverted for security reasons.

wolfbeast:
I don't think that commit should be reverted. Like I stated as rationale: if there is no content-type header, you can't perform this check, and it must be allowed because you're missing one of the two arguments to make a comparison between.

PR #1506, opened 2017-11-25
Bug1299753

wolfbeast:
Work to solve issues with `<details>` and `<summary>` if using CSS float on it.

In the case of floating these elements, a block frame needs to be created
(see specific in CSS 2.2), which this set of patches achieves.

nsCSSFrameConstructor::ConstructNonScrollableBlock() has logic to
determine whether to create a block formatting context for a block
frame. The function is refactored to make it reusable by
nsCSSFrameConstructor::ConstructDetailsFrame().

Also, made NS_NewBlockFrame() accept two arguments as other frame
factory functions so that it could be pointed by BlockFrameCreationFunc.
NS_NewBlockFormattingContext is changed accordingly.

The construction for a scrollable DetailsFrame has been revised similarly.

### DIFF
--- a/layout/base/nsCSSFrameConstructor.cpp
+++ b/layout/base/nsCSSFrameConstructor.cpp
@@ -3244,8 +3244,7 @@ nsCSSFrameConstructor::ConstructFieldSetFrame(nsFrameConstructorState& aState,
   }
 
   nsContainerFrame* blockFrame =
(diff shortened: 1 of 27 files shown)


wolfbeast:
Verified passing all tests (including GitHub live site). Merging.

Issue #1523, opened 2017-12-01
ES 2016 ** exponentiation operator


stevenschlansker:
The exponentiation operator `**` (equivalent to `Math.pow`) is in ES2016 and has been out in the wild for longer than that.  Found a site using it, in PM 27.6.1:
```
2 ** 5
  SyntaxError: expected expression, got '*'
```

In Chrome:
```
2 ** 5
  32
```

janekptacijarabaci:
Bugs: [1135708](https://bugzilla.mozilla.org/show_bug.cgi?id=1135708), [1291212](https://bugzilla.mozilla.org/show_bug.cgi?id=1291212), [1243858](https://bugzilla.mozilla.org/show_bug.cgi?id=1243858)

wolfbeast:
[sarcasm]Because using Math.* functions for mathematical operations is _soooo_ 2015[/sarcasm]
Let's at least make sure we implement this correctly in one go, [unlike Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=789161).

wolfbeast:
I'm not going to bother with more than a baseline implementation here. Sorry, benchmark folks -- `**` will be slower than `Math.pow()`! If performance is important, use `Math.pow()` as always, because that's a dedicated math lib that can be expected to be performant and efficient.

wolfbeast:
**Implementation note:** e262 doesn't specify special behavior for a unary expression (I verified), so our implementation will follow **mathematical rules** and allow `-a ** b` without parentheses, prioritizing the unary sign expression over the operation. Firefox throws an error, enforcing parentheses.

PR #1531, opened 2017-12-05
Implement `**` exponentiation operator.

wolfbeast:
This makes `**` available as an exponentiation operator in addition to `Math.pow()`


wolfbeast:
This resolves #1523

Issue #1567, opened 2018-01-04
Crash in MP4Metadata::GetNumberTracks


wolfbeast:
As [reported on the forum](https://forum.palemoon.org/viewtopic.php?f=3&t=17906).

Most likely the website in question has some auto-playing video that is either corrupt or yanked out from under us before the meta-data can be read. Possibly video advertising considering the semi-random nature of it...
The crash address points at the following line of code crashing:
```C++
size_t tracks = mPrivate->mMetadataExtractor->countTracks();
```

Since no stack trace or dump is available, this is all we have to go on for now. Possibly adding some null checks with early exits (on `mPrivate` and `mMetadataExtractor`) would solve this issue?

trav90:
Investigating.

trav90:
I have been unable to reproduce this crash myself, however I have found several Mozilla reports of crashes in MP4Metadata::GetNumberTracks. PR to follow shortly.

wolfbeast:
I'll build an unstable today with this merged in; the original reporter should be able to verify if this solves the problem.

wolfbeast:
Can we assume this is fixed in UXP?

trav90:
Yes it should be.

PR #1569, opened 2018-01-05
Reduce some asyncronocity in MSE code

trav90:
The W3C spec indicates that while everything in MSE is asynchronous, the abort() command is to interrupt the current segment parser loop and have the reset parser loop synchronously completes the frames present in the input buffer. This causes a fundamental issue that abort() will never result in a deterministic outcome as the segment parser loop may be in different condition.

We used to really attempt to abort the current operation, however there could have been a race in the order in which tasks were queued, which would result in a crash (with the most common one being in MP4Metadata::GetNumberTracks). As such, we now simply wait for the current appendBuffer to complete.

This also allows us to cleanup some now unused code and reduce a bit of the complexity of our MSE code.

This hopefully will resolve the crash in #1567, however as I have not been able to reproduce that crash verification is needed. I have lightly tested the changes in this PR on YouTube, Vimeo, and Epicurious.

### DIFF
--- a/dom/media/mediasource/MediaSourceDecoder.cpp
+++ b/dom/media/mediasource/MediaSourceDecoder.cpp
@@ -299,6 +299,12 @@ MediaDecoderOwner::NextFrameStatus
 MediaSourceDecoder::NextFrameBufferedStatus()
 {
   MOZ_ASSERT(NS_IsMainThread());
+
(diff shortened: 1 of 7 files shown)


trav90:
I've tested these changes a bit more thoroughly and no issues seen. Still need verification on whether this fixes #1567.

PR #1576, opened 2018-01-10
Collapse notification bar in fullscreen

SpockMan02:
Remove white bar at the bottom of the screen when watching HTML5 video in fullscreen on Windows
See https://forum.palemoon.org/viewtopic.php?f=3&t=17707, https://bugzilla.mozilla.org/show_bug.cgi?id=1076626, and https://forum.palemoon.org/viewtopic.php?f=29&t=17918

### DIFF
--- a/browser/base/content/browser.css
+++ b/browser/base/content/browser.css
@@ -685,3 +685,9 @@ toolbarbutton[pmkit-button="true"] > .toolbarbutton-badge-container > .toolbarbu
   width: 16px;
   height: 16px;
 }
+
+/* Remove white bar at the bottom of the screen when watching HTML5 video in fullscreen on Windows */
+#main-window[inFullscreen] #global-notificationbox,
+#main-window[inFullscreen] #high-priority-global-notificationbox {
+  visibility: collapse;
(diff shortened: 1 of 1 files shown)


wolfbeast:
@Lootyhoof Sanity check on this, please?

Lootyhoof:
It _seems_ OK at least, as it's only hiding the notification bar while in fullscreen. I'm not sure what kind of notification bar spawns this though (seems the "normal" one by NoScript et al. doesn't use these IDs) so I can't test with client-side CSS applied, nor can I reproduce this myself (do we even have any STR?). The comment in the commit also specifically mentions Windows - but as this is in `/content/browser.css` this will affect other platforms too.

Note that this likely won't discern between fullscreen video and normal fullscreen (F11), which could be annoying for people who want notifications when in fullscreen mode.

As an aside, this is more `Theme/UI` than `Theme changes`, as it shouldn't affect themes.

SpockMan02:
It seems I was mistaken when I said "on Windows"; the Firefox BugZilla bug actually says OS: All.

Issue #1579, opened 2018-01-14
IP leak via WebRTC when built with --enable-webrtc


Ud71p:
PaleMoon has optional WebRTC support, see:
https://github.com/MoonchildProductions/Pale-Moon/commit/d4a4492c6588116d4888510cbb2e1db9ab077a24

I built PaleMoon with --enable-webrtc, because I need it for work.

I often use Tor or VPN. My IP address should be hidden from websites. But with WebRTC it leaks.

Step to reproduce:

Go to 'https://check.torproject.org/'.
Observe " Congratulations. This browser is configured to use Tor. Your IP address appears to be: not-my-IP".
Go to 'about:config'.
Set media.peerconnection.enabled = true.
Go to 'https://diafygi.github.io/webrtc-ips/'.
Observe the problem: "Your public IP addresses: my-real-leaked-IP", that is they can see my real IP address.

The issue is well-known, described here:
https://en.wikipedia.org/wiki/WebRTC#Concerns
And in linked article:
https://torrentfreak.com/huge-security-flaw-leaks-vpn-users-real-ip-addresses-150130/

Version: 27.6.2 (64-bit)

As a work-around, set media.peerconnection.enabled = false.

wolfbeast:
We don't ship Pale Moon with WebRTC. It is well-known that WebRTC exposes the IP through peer connections (because that is by design).

Not sure what you think this issue is supposed to address other than providing information for people using non-standard builds, but it's neither actionable nor in the right place in this issue tracker.

Issue #1573, opened 2018-01-07
IP leak via WebRTC


Ud71p:
I often use Tor or VPN. My IP address should be hidden from websites. But with WebRTC it leaks.

Step to reproduce:

Go to 'https://check.torproject.org/'.
Observe " Congratulations. This browser is configured to use Tor. Your IP address appears to be: *not-my-IP*".
Go to 'about:config'.
Verify that media.peerconnection.enabled = true.
Go to 'https://diafygi.github.io/webrtc-ips/'.
Observe the problem: "Your public IP addresses: *my-real-leaked-IP*".

They can see my real IP address.

The issue is well-known, described here:
https://en.wikipedia.org/wiki/WebRTC#Concerns
And in linked article:
https://torrentfreak.com/huge-security-flaw-leaks-vpn-users-real-ip-addresses-150130/

Version: 27.6.2 (64-bit)

As a work-around, compile PaleMoon without WebRTC, or set media.peerconnection.enabled = false.

mattatobin:
Pale Moon does not compile WebRTC as it is.. In fact.. I am not even sure it builds currently..

So, do you even know anything about our project?

Ud71p:
Yes, PalMoon has (optional) WebRTC support.

Yes, it does compile with --enable-webrtc, I tried myself.

There are even packages for distros making the webrtc flag available for users to flip in package manager UI, see here the packages have the webrtc flag (minus '-' means disabled by default, bug can be turned on by users):
https://gpo.zugaina.org/www-client/palemoon

So, do YOU even know anything about your project?

I opened new issue with more info:
https://github.com/MoonchildProductions/Pale-Moon/issues/1579

mattatobin:
Just because it is in the codebase doesn't mean it should be used. Official configuration does NOT include it. If this option is being exposed to source based distros.. That needs to stop as the code is NOT maintained in the Tycho (Pale Moon 27) codebase and hasn't been a thing that Pale Moon has included for years.

IF there are binary builds out there with webrtc they are in violation of the terms regarding redistribution with offical branding.

Yes, I do know this project and I also know in no circumstance is webrtc enabled Pale Moon no matter if the code builds or not webrtc is unmaintained and building it is NOT ALLOWED.

mattatobin:
But if it requires a patch to the build system to make it explicit.. Guess I will have to make one, eh?

wolfbeast:
official branding + webrtc -> throw error.

PR #1580, opened 2018-01-14
Do not allow WebRTC to be built when MOZ_PHOENIX=1

mattatobin:
In Tycho and older, WebRTC is unsupported and potentially insecure code. While most obey our guidelines and do not build it for binary redistribution it seems that some source based distros expose the configure flag to users for selection. This will prevent WebRTC from building if MOZ_PHOENIX=1 effectively --enable-application=browser.

See: #1573 / #1579


wolfbeast:
I think we should just tie this to official branding. If people use their own branding they can take all the risks with unmaintained code they want.

wolfbeast:
Resolved per discussion by removing the insecure/unmaintained code from Tycho.

PR #1583, opened 2018-01-15
Remove WebRTC

mattatobin:
This pull request rips as much of WebRTC that can reasonably be done without having to dig deep into various bits of DOM Media. It does NOT eliminate conditional code.. I can still do this but it wouldn't change what is left of WebRTC that is required.

This also changes the configure option to give an error when trying to use --enable-webrtc.

I have tested this on Windows and it builds successfully and cursory run through media testing.

I would like a Linux verification on successful build. Though everything that needs to be there for NOT building webrtc should be there.

@trav90 or @khronosschoty


SpockMan02:
Will WebRTC return when Pale Moon moves to UXP?

mattatobin:
Not for Pale Moon

khronosschoty:
Did not test building with --enable-webrtc. 
Did build using standard config. Encountered no issues, everything seems fine.

PR #1593, opened 2018-01-19
Perform an `XCTO: nosniff` check only for 2xx responses

### LINKED ISSUE #1592: `XCTO: nosniff`and Microsoft TFS / SharePoint issues
As [reported on the forum](https://forum.palemoon.org/viewtopic.php?f=3&t=18032), various Microsoft servers set `XCTO: nosniff` header for `401` responses, which leads to incorrect operation in Pale Moon.

It seems that other browsers check `XCTO: nosniff` only for `2xx` responses, so we need to act the same way.

JustOff:
Follow-up to #1489, resolves #1592.


JustOff:
I think it makes sense to uplift this to the release branch.

PR #1599, opened 2018-01-23
Make caret behind CJK char normal width by default

### LINKED ISSUE #1555: Provide an option to disable CJK character caret width auto increasing
https://github.com/MoonchildProductions/Pale-Moon/blob/5214789d035f4e4c08899a6e281116520f605ec6/layout/base/nsCaret.cpp#L191-L193

Demo:
demo1
demo2

It is annoying when I type mixing languages.

SpockMan02:
Add a preference (layout.cjkthickcaret; false by default) for whether or not the caret should be thicker than normal when located behind a Chinese/Japanese/Korean character. Currently this is always the case.

This resolves #1555.

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -4909,4 +4909,4 @@ pref("dom.secureelement.enabled", false);
 pref("plugins.rewrite_youtube_embeds", true);
 
 // Thick caret when behind CJK characters
-pref("layout.cjkthickcaret", false);
+pref("layout.cjkthickcaret", true);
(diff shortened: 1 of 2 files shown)


modules/libpref/init/all.js
```diff
@@ -4907,3 +4907,6 @@ pref("dom.secureelement.enabled", false);
 
 // Turn rewriting of youtube embeds on/off
 pref("plugins.rewrite_youtube_embeds", true);
+
+// Thick caret when behind CJK characters
+pref("layout.cjkthickcaret", false);
```
wolfbeast: Our current behavior should be default -- i.e.: thicken the caret in case of CJK for easier location of it in naturally bolder text due to glyph construction (which is why this code is there to begin with).

PR #1601, opened 2018-01-24
Hook up scheme fixups to the browser's Front-end

### LINKED ISSUE #1600: Hook up scheme fixups to the browser FE
We have the platform code for https://bugzilla.mozilla.org/show_bug.cgi?id=935377 -- but it's not hooked up to our FE so scheme fixups never happen.

This should be easy to fix by simply hooking up the machinery.

janekptacijarabaci:
This resolves #1600


wolfbeast:
Thank you!

PR #1619, opened 2018-02-24
Add Firefox-compat UA override for NationStates

SpockMan02:
NationStates forum Technical board thread: https://forum.nationstates.net/viewtopic.php?f=15&t=435454&p=33415387#p33415387


mattatobin:
Why is this site important?

wolfbeast:
If you are already in a dialogue with the site owners then please try to solve it with them. 
If the problem is with highcharts then the site owners need to get in touch with highcharts to fix their browser detection.

The included list with the browser should remain as short as possible for performance reasons (it is parsed with every web request) so any site that isn't of global importance or at the very least has major national financial/etc. importance should not be included with the browser, but worked around on the user's end by including an override.

SpockMan02:
Ah, okay, gotcha. I'll close this, then.

PR #1636, opened 2018-03-08
Add UA overridde for Dropbox

adeshkp:
Native UA mode added a while back doesn't work anymore


mattatobin:
You're kidding right? IE? Can you not find something closer to a Mozilla UA?

Sa-Ja-Di:
what works:

Mozilla/5.0 (Windows NT 6.3; Win64; x64; rv:10.0) Gecko/20100101 Firefox/10.0


until

Mozilla/5.0 (Windows NT 6.3; Win64; x64; rv:50.0) Gecko/20100101 Firefox/50.0

If taking a user agent above 52, Dropbox page is only partly usable. IE user agent looks like the more stable choice imho ;)

adeshkp:
@mattatobin 
Only after Moonchild.
https://forum.palemoon.org/viewtopic.php?f=19&t=18235&p=136533#p134028

wolfbeast:
Merging this in. We can change it again if they break it again.

sergeevabc:
Thank you.

PR #1640, opened 2018-03-12
Synchronously read the session file if the session need to be restored at startup

### LINKED ISSUE #1310: Prevent loading of default home page when restoring session.
Currently, there is a race between loading the sessionstore file and browser startup. Mozilla's code only opts out of loading the home page when the session store file has been read and information is available, and at that point in time replaces the (partially or completely) loaded home page with whatever should be restored; probably in an attempt to get as much loaded as soon as possible to shave off more time from browser startup (the prominent telemetry property for health report).

See, e.g. https://forum.palemoon.org/viewtopic.php?f=3&t=14216

Almost always, loading sessionstore loses this race, and the home page load is initialized. This needs to be changed.

JustOff:
This resolves #1310 and reduces the overall startup time.

With this patch applied in my test environment with ~25 tabs, the sessionRestored event fires after ~3300 ms instead of ~4000 ms.

### DIFF
--- a/browser/components/sessionstore/nsSessionStartup.js
+++ b/browser/components/sessionstore/nsSessionStartup.js
@@ -77,9 +77,14 @@ SessionStartup.prototype = {
       return;
     }
 
-    _SessionFile.read().then(
-      this._onSessionFileRead.bind(this)
-    );
+    if (Services.prefs.getBoolPref("browser.sessionstore.resume_session_once") ||
+        Services.prefs.getIntPref("browser.startup.page") == 3) {
+      this._ensureInitialized();
+    } else {
+      _SessionFile.read().then(
+        this._onSessionFileRead.bind(this)
+      );
+    }
   },
 
   // Wrap a string as a nsISupports


wolfbeast:
:+1: for reducing async BS.

JustOff:
Unfortunately this can't be easily applied to UXP, there we likely need to use the solution from [bug 1054740](https://bugzilla.mozilla.org/show_bug.cgi?id=1054740).

mattatobin:
I wonder if it will be a problem for Pale Moon on UXP? I assume you are specifically talking about Basilisk tho..

JustOff:
As far as I can see `nsSessionStartup.js` is part of application, so yes, only Basilisk (or whatever) will require a different solution on UXP.

mattatobin:
Understood, thanks @JustOff

Lorienna:
This issue had to do with some changes in `OS.File` and the so-called `worker` modules.

I can provide the pushlogs and BZ bugs if you think they would aid in fixing it in UXP-based applications.

Lorienna:
@JustOff That's when the homepage actually started to appear, and I thought that might be of some aid.
Nevertheless, thank you, and hopefully someone will fix this in Basilisk as well.

mattatobin:
@Lorienna Why don't you do it?

Lorienna:
@mattatobin I won't do it because I have neither necessary skills nor a suitable machine.
Sorry for the late reply.

PR #1652, opened 2018-03-21
[suggestion] "Enter address" if the feature "autocomplete" is disabled

### LINKED ISSUE #1651: URL bar - placeholder - "Search or enter address" even if the feature "autocomplete" and "keyword.enabled" are disabled
Ad https://forum.palemoon.org/viewtopic.php?f=3&t=18634

Preferences:
`browser.urlbar.suggest.history`
`browser.urlbar.suggest.bookmark`
`browser.urlbar.suggest.openpage`
__Edited:__
`keyword.enabled`

> On new tabs I get the following text in the address bar: "Search or enter address". This is incorrect if above prefs are set to "false" since "Search" (autocomplete and keyword) is not allowed.

If you agree - also `UXP-task`?

janekptacijarabaci:
This resolves #1651

---

---

### DIFF
--- a/browser/base/content/browser.js
+++ b/browser/base/content/browser.js
@@ -337,9 +337,18 @@ const gSessionHistoryObserver = {
 };
 
 var gURLBarSettings = {
-  prefSuggestBookmark: "browser.urlbar.suggest.bookmark",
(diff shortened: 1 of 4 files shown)


janekptacijarabaci:
Yes I will do it.

PR #1667, opened 2018-04-23
Improve some button labels in preferences

### LINKED ISSUE #1325: Improve button messages for removing cookies
Options > "Privacy" tab > "remove individual cookies" link

Here are the current Pale Moon messages:
* when there is no search filter:
Remove Cookie - Remove All Cookies
* when there is a search filter:
Remove Cookie[s] - Remove All Cookies

I found it confusing ("a search filter is applied, if I click "Remove All Cookies", will it remove only the shown cookies or all the cookies?")

Then I gave a run with Firefox, and noticed they are using much less ambiguous messages:
* when there is no search filter:
Remove Selected - Remove All
* when there is a search filter:
Remove Selected - Remove All Shown

I think these clearer messages should be ported to Pale Moon.

adeshkp:
This takes care of button labels for handling cookies and passwords.
This resolves #1325 

This does not address "site data settings" as I could not figure out how to invoke site data settings window from Pale Moon UI. Neither could I find it in the codebase.

### DIFF
--- a/browser/components/preferences/cookies.js
+++ b/browser/components/preferences/cookies.js
@@ -5,6 +5,8 @@
 
 const nsICookie = Components.interfaces.nsICookie;
 
+Components.utils.import("resource://gre/modules/PluralForm.jsm");
+
 var gCookiesWindow = {
   _cm               : Components.classes["@mozilla.org/cookiemanager;1"]
                                 .getService(Components.interfaces.nsICookieManager),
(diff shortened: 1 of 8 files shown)


JustOff:
IFAIU, there will be no more major releases of Pale Moon 27, but this PR changes strings, so we could have problem with language packs. Isn't be better to implement this in UXP?

wolfbeast:
We won't be accepting any feature or string changes anymore on this repo. Only critical stability and security issues will be addressed in point releases.
Considering UXP already has these changes in the platform (most likely) there's no real reason to put in extra work.

JustOff:
> Considering UXP already has these changes in the platform (most likely) there's no real reason to put in extra work.

No, it has not.

mattatobin:
Didn't the cookie system on UXP change quite a bit?

wolfbeast:
> No, it has not. Neither application, nor toolkit.

Was this changed post-52 then? I thought it was before.

JustOff:
https://bugzilla.mozilla.org/show_bug.cgi?id=1338036 - RESOLVED FIXED in Firefox 54

wolfbeast:
Well, port away then! :)

wolfbeast:
https://github.com/MoonchildProductions/UXP/issues/253

JustOff:
I leave it to @adeshkp or someone else who thinks this feature is essential ;-)

adeshkp:
Ok, I'll do it. Now I guess, this needs to be done for both Pale Moon and Basilisk. Right?

mattatobin:
That would be nice but continue this in https://github.com/MoonchildProductions/UXP/issues/253

PR #1712, opened 2020-01-27
Simplify HWA selections

wolfbeast:
This should complete the FE&BE work for simplifying the way HWA prefs work.

A Linux build/functionality check would be appreciated prior to merging.


mattatobin:
@JustOff please note string changes when this lands.

palemoon/locales/en-US/chrome/browser/preferences/advanced.dtd
```diff
@@ -21,10 +21,13 @@
 <!ENTITY useAutoScroll.accesskey         "a">
 <!ENTITY useSmoothScrolling.label        "Use smooth scrolling">
 <!ENTITY useSmoothScrolling.accesskey    "m">
-<!ENTITY allowHWAccel.label              "Use hardware acceleration when available">
-<!ENTITY allowHWAccel.accesskey          "r">
 <!ENTITY checkSpelling.label             "Check my spelling as I type">
 <!ENTITY checkSpelling.accesskey         "t">
+<!ENTITY HWAccel.label                   "Hardware Acceleration (HWA)">
```
JustOff: Sorry to be a little late, but do we really need HWA acronym here? It's not referred anywhere near, and it's not entirely clear how this is supposed to be translated (if it should). Maybe we could just get rid of it?

wolfbeast: I included it because that is how we _always_ refer to it (e.g. on the forum, when providing tech support, etc.) -- it shouldn't be translated.

JustOff: > it shouldn't be translated

Ok, I'll add this note for translators.

JustOff:
@mattatobin, thanks for the reminder, however @wolfbeast previously asked me to apply such labels to issues, not pull requests.

PR #1731, opened 2020-02-29
Add 1.25x play speed to the html player context menu

### LINKED ISSUE #1730: Add 1.25x play speed to the html player context menu
As discussed on the forum: https://forum.palemoon.org/viewtopic.php?f=5&t=23918.

JustOff:
This adds `1.25x` play speed option to the html player context menu, ~and also beautifies the names of the play speed menu items, as is done in Basilisk.~

Resolves #1730.

### DIFF
--- a/palemoon/base/content/browser-context.inc
+++ b/palemoon/base/content/browser-context.inc
@@ -85,6 +85,12 @@
                     name="playbackrate"
                     checked="true"
                     oncommand="gContextMenu.mediaCommand('playbackRate', 1.0);"/>
+          <menuitem id="context-media-playbackrate-125x"
+                    label="&mediaPlaybackRate125x.label;"
+                    accesskey="&mediaPlaybackRate125x.accesskey;"
+                    type="radio"
+                    name="playbackrate"
+                    oncommand="gContextMenu.mediaCommand('playbackRate', 1.25);"/>
           <menuitem id="context-media-playbackrate-150x"
                     label="&mediaPlaybackRate150x.label;"
                     accesskey="&mediaPlaybackRate150x.accesskey;"
--- a/palemoon/locales/en-US/chrome/browser/browser.dtd
+++ b/palemoon/locales/en-US/chrome/browser/browser.dtd
@@ -423,6 +423,8 @@ items are mutually exclusive. -->
 <!ENTITY mediaPlaybackRate050x.accesskey "S">
 <!ENTITY mediaPlaybackRate100x.label "Normal Speed">
 <!ENTITY mediaPlaybackRate100x.accesskey "N">
+<!ENTITY mediaPlaybackRate125x.label "Increased Speed (1.25×)">
+<!ENTITY mediaPlaybackRate125x.accesskey "I">
 <!ENTITY mediaPlaybackRate150x.label "High Speed (1.5×)">
 <!ENTITY mediaPlaybackRate150x.accesskey "H">
 <!-- LOCALIZATION NOTE: "Ludicrous Speed" is a reference to the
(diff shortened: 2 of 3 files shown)


palemoon/base/content/browser-context.inc
```diff
@@ -73,27 +73,33 @@
       <menu id="context-media-playbackrate" label="&mediaPlaybackRate2.label;" accesskey="&mediaPlaybackRate2.accesskey;">
         <menupopup>
           <menuitem id="context-media-playbackrate-050x"
-                    label="&mediaPlaybackRate050x.label;"
-                    accesskey="&mediaPlaybackRate050x.accesskey;"
+                    label="&mediaPlaybackRate050x2.label;"
+                    accesskey="&mediaPlaybackRate050x2.accesskey;"
```
wolfbeast: There's no need to change the names of these l10n strings!

palemoon/locales/en-US/chrome/browser/browser.dtd
```diff
@@ -419,17 +419,19 @@ items are mutually exclusive. -->
...
+<!ENTITY mediaPlaybackRate125x2.accesskey "F">
+<!ENTITY mediaPlaybackRate150x2.label "Faster (1.5×)">
+<!ENTITY mediaPlaybackRate150x2.accesskey "a">
+<!-- LOCALIZATION NOTE: "Ludicrous" is a reference to the
 movie "Space Balls" and is meant to say that this speed is very
 fast. -->
-<!ENTITY mediaPlaybackRate200x.label "Ludicrous Speed (2×)">
-<!ENTITY mediaPlaybackRate200x.accesskey "L">
+<!ENTITY mediaPlaybackRate200x2.label "Ludicrous (2×)">
+<!ENTITY mediaPlaybackRate200x2.accesskey "L">
```
wolfbeast: I disagree with changing the names of these options.
If for nothing else, the Spaceballs reference is "Ludicrous speed" not "Ludicrous". (light speed - ridiculous speed - ludicrous speed)

JustOff: What is the reason to use the word speed twice? This does not make things clearer; it only clutters the menu with repetition. Speed ... speed ... speed ...



Do not you think that this is better?

wolfbeast: It makes perfect sense to have full menu item names regardless of what the parent item is called. It's not "beautification" to just shorten everything.

Doing this would make it **inconsistent** with the rest of the application:
`Bookmarks -> Organize Bookmarks`, not `Bookmarks -> Organize`
`History -> Show All History`, not `History -> Show All`
`Print -> Print Preview`, not `Print -> Preview`
etc.
So `Play Speed -> Normal Speed`, not `Play Speed -> Normal`

I suggest using "Increased Speed" instead of "Fast" for 1.25x

JustOff: > Doing this would make it inconsistent with the rest of the application

Well, thanks for the clarification, it's hard to argue with that (although I would prefer to see a shorter version everywhere, but this is of course only my personal preference).

I will now update the PR as you requested.

wolfbeast: Well we can always ask the community about the use of text in the browser. See what they prefer, and if they prefer shorter versions everywhere then we can make that an issue on its own to change it all, consistently, in one go.

palemoon/base/content/browser-context.inc
JustOff: Fixed.

palemoon/locales/en-US/chrome/browser/browser.dtd
JustOff: Fixed.

> Well we can always ask the community about the use of text in the browser. See what they prefer, and if they prefer shorter versions everywhere then we can make that an issue on its own to change it all, consistently, in one go.

Oh, no! Translators will go crazy)

wolfbeast: > Oh, no! Translators will go crazy)

So you're against it after all? ;-)

PR #1741, opened 2020-03-19
Make UI Migration code nss9 file mitigation conditional dependent on MOZ_SECURITY_SQLSTORE

mattatobin:
This should be merged after PR MoonchildProductions/UXP#1487


wolfbeast:
Sorry but this mitigation is only there for the unintentional presence of sql stores from the previous activation of them. Once we decide to enable sql stores (which is unrelated to the linked PR which enables to platform _possibility_ to enable them) then this entire block just needs to go because from that point forward these stores are to be considered intentional.

PR #1749, opened 2020-03-30
Issue #1748 - Remove Windows rule from osx CSS

### LINKED ISSUE #1748: macOS window controls obscured on maximized private windows or windows with personas
When a private browsing window or a window with a persona/lwtheme is maximized (so that it takes up all or almost all space on the screen but is not in fullscreen; accomplished by double-clicking the titlebar or on older OS X versions by clicking the green caption button), the window controls, including the caption buttons and, if present, fullscreen arrows/button, are shifted up and out of the window so that they are completely or mostly hidden and cannot be used. The problem has been recently described on the forum:
https://forum.palemoon.org/viewtopic.php?f=41&t=23935&start=40#p186227
https://forum.palemoon.org/viewtopic.php?f=41&t=24052#p186678
This is what it looks like (the corner of a maximized private browsing window on macOS 10.12): Screenshot 2020-03-29 16 14 49
The issue is caused by misplaced CSS intended to accommodate the Windows appmenu button in the osx theme. Removing one rule fixes the problem, leaving the XUL window controls where they should be and not moving them inconveniently when the window is maximized: Screenshot 2020-03-29 16 18 56
I have the fix (#1747) and will make another pull request shortly.

SpockFan02:
This resolves #1748.

### DIFF
--- a/palemoon/themes/osx/browser.css
+++ b/palemoon/themes/osx/browser.css
@@ -96,10 +96,6 @@
     -moz-appearance: -moz-window-button-box;
 }
 
-#main-window[sizemode="maximized"] #titlebar-buttonbox {
-    -moz-appearance: -moz-window-button-box-maximized;
-}
-
 .titlebar-placeholder[type="appmenu-button"] {
     margin-left: 4px;
 }


wolfbeast:
Perfect. Thank you!

PR #1777, opened 2020-04-13
Put the osx private browsing indicator in its own element

SpockFan02:
This also allows the titlebar of private browsing windows to use the normal titlebar appearance again, and shows the indicator in fullscreen.


wolfbeast:
As discussed this is no longer the desired development path. closing.

PR #1779, opened 2020-04-16
Update to prevent duplicate icons

blipk:
See: https://forum.palemoon.org/viewtopic.php?f=37&t=24090

Does not work on some systems with the quotes.

### DIFF
--- a/palemoon/branding/official/palemoon.desktop
+++ b/palemoon/branding/official/palemoon.desktop
@@ -89,7 +89,7 @@ Categories=Network;WebBrowser;
 MimeType=text/html;text/xml;application/xhtml+xml;application/vnd.mozilla.xul+xml;text/mml;x-scheme-handler/http;x-scheme-handler/https;x-scheme-handler/ftp;
 StartupNotify=false
 Actions=NewTab;NewWindow;NewPrivateWindow;
-StartupWMClass="pale moon"
+StartupWMClass=Pale moon
 
 [Desktop Action NewTab]
 Name=Open new tab


wolfbeast:
PR without an attached Issue.

palemoon/branding/official/palemoon.desktop
```diff
@@ -89,7 +89,7 @@ Categories=Network;WebBrowser;
 MimeType=text/html;text/xml;application/xhtml+xml;application/vnd.mozilla.xul+xml;text/mml;x-scheme-handler/http;x-scheme-handler/https;x-scheme-handler/ftp;
 StartupNotify=false
 Actions=NewTab;NewWindow;NewPrivateWindow;
-StartupWMClass="pale moon"
+StartupWMClass=Pale moon
```
wolfbeast, line 92: Either capitalize both or none -- I'm assuming it's a case sensitive match so make it match?

blipk:
Did not work for me with the lower case.

-------- Original Message --------

wolfbeast:
> Did not work for me with the lower case.

Can someone please figure out why `Pale moon` would work and `pale moon` would not? What about `Pale Moon`? Isn't it just ignoring the "Moon" part in effect without the quotes?

I'd prefer not to break setups to fix others.

wolfbeast:
Tag #1781

blipk:
When I launch the application from the favourite icon on the panel without the StartupWMClass in the desktop file:
```
xprop | grep -i class
WM_CLASS(STRING) = "Navigator", "Pale moon"
```

Pale moon with the capital P is correct, both me and the user in the thread reported it with a capital P.

I have looked at a lot of .desktop files and none of them put the values in quotes.
I'm fairly positive this will work for most users:
`StartupWMClass=Pale moon`

palemoon/branding/official/palemoon.desktop
blipk: Whoever made the commit with the lower case p was incorrect, both me and the user in the forum thread have a capital P.

I have looked at a lot of desktop files and the values have never been in quotes. I'm certain that `StartupWMClass=Pale moon` will work for everyone.

wolfbeast:
@trav90 I'm really in no position to make a call on this; what do you think?

wolfbeast:
I'll be merging this in, assuming it's going to be OK, since I have no way to verify different environments.

PR #1783, opened 2020-04-27
[profile] Copy default userChrome and userContent examples to profile

wolfbeast:
Package `userChrome` and `userContent` examples and copy them to new profiles.


wolfbeast:
> Suggest squash commit.

Already planned doing that ;)

PR #1827, opened 2020-08-13
Restore missing close icon on the undo prompt at the quick dial page

### LINKED ISSUE #1826: Missing close button on undo prompt after removing pages from the quick dial
**Description:**
After removing pages from the quick dial, an undo prompt is shown. The close button used to hide the prompt is missing.

**Affected versions:**
Pale Moon v28.0.0 and above

**Steps to reproduce:**
1. Open the quick dial page (about:newtab)
2. Remove any page from the grid

**Expected result:**
missing-close-icon-expected

**Actual result:**
missing-close-icon

FranklinDM:
This resolves #1826.

While this could've all been placed in the shared new tab style sheet, OSX does not include the SVG close icon used by the other platforms and has its own custom close icon. If desired, I could also do it the other way and include the SVG close icon on OSX.

### DIFF
--- a/palemoon/themes/shared/newtab/newTab.css.inc
+++ b/palemoon/themes/shared/newtab/newTab.css.inc
@@ -43,13 +43,14 @@ body {
   outline: 1px dotted;
 }
 
-
-#newtab-undo-close-button > .toolbarbutton-text {
-  display: none;
-}
-
-#newtab-undo-close-button:-moz-focusring {
-  outline: 1px dotted;
+#newtab-undo-close-button {
+  -moz-appearance: none;
+  padding: 0;
+  border: none;
+  width: 16px;
+  height: 16px;
+  float: right;
+  right: 0;
 }
 
 /* TOGGLE */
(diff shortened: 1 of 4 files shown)


palemoon/themes/shared/newtab/newTab.css.inc
```diff
@@ -43,13 +43,14 @@ body {
   outline: 1px dotted;
 }
 
-
-#newtab-undo-close-button > .toolbarbutton-text {
-  display: none;
-}
-
-#newtab-undo-close-button:-moz-focusring {
```
wolfbeast, line 51: You've removed this rule but not replaced it for accessibility access (keyboard nav).

wolfbeast: Can you please indicate if this was on purpose, and if so, why would it not be needed?

FranklinDM: Sorry if I wasn't able to respond promptly, real life caught on unfortunately.

I intentionally removed the said rule since it is no longer necessary and only creates a double outline (which occurs on the `Undo.` and `Restore All.` buttons on the same prompt. I presumed this is intended behavior since the old new tab design in v27 has this too). This is also the case in Basilisk, from which I patterned the changes on this PR.

Double outline on the buttons on the prompt:
double outline

Focus ring changes:
focus ring

PR #1828, opened 2020-08-13
Padlock Improvements

### LINKED ISSUE #1717: Rewrite the Pale Moon padlock code
The padlock code we're using is relying on very old nsIWebProgressListener code that makes assumptions about old IDL specifics and not using additional pieces of info it provides (e.g. weak cipher info). The platform has also done away with always providing a security level along with "is_secure" so the padlock code needs to get a once-over to modernize it based on our current API.

Specifically, anytime STATE_IS_SECURE it also returns STATE_SECURE_HIGH in most cases (STATE_SECURE_MED and STATE_SECURE_LOW were removed at some point along the way).

This makes STATE_SECURE_HIGH basically worthless and the other two still being bitwise values in the IDL seems to have been kept around only to satisfy consumers of nsIWebProgressListener in a rare case of not breaking extensions but most likely so internal consumers didn't have to be rewritten at the time.

What needs to happen in my view is either STATE_SECURE_HIGH and the idl bitwise definitions that aren't valid anymore need to be removed or they all need restored to working order. Otherwise we need to make notes and keep in mind that if STATE_IS_SECURE we should also set STATE_SECURE_HIGH to maintain consistancy.

As for the padlock component in browser base this simply needs aligned to what is actually being used in cpp and a condition should be added to switch logic for the possibility of being STATE_IS_SECURE but not having STATE_SECURE_HIGH. Which very well might be desirable for the case of local injection on a secure site in the uBO data: uri case.

RealityRipple:
It's time to ignore STATE_SECURE_{HIGH,MED,LOW} from UXP, as they're essentially depricated.

This draft includes the following changes:
- Stricter checks for mixed content in EV situations and active mixed content in secure situations.
- **SSL 3** marked as `Broken`
- **TLS 1.0** and **1.1** marked as `Low` (Gotta wonder, should we bring `Medium` back just for 1.1?)
- **EXPORT** ciphers marked as `Broken`
- **RC2** ciphers marked as `Broken`
- **RC4 + MD5** ciphers marked as `Broken`
- **RC4 + SHA** ciphers marked as `Low`
- **3DES** ciphers marked as `Low`

Note that the detection method for ciphers just uses `indexOf()` on the cipherSuite value, rather than specifically checking for each specific value. This could potentially lead to issues in certain cases.

Resolves #1717 (or heads toward a resolution).

### DIFF
--- a/palemoon/base/content/padlock.js
+++ b/palemoon/base/content/padlock.js
@@ -84,7 +84,9 @@ var padlock_PadLock =
             } else if (aCipher.indexOf("_SHA") > -1) {
               level = "low";
             }
-          } else if (aCipher.indexOf("_DES_EDE3_") > -1) {
+          } else if (aCipher == "TLS_RSA_WITH_3DES_EDE_CBC_SHA") {
+            level = "low";
+          } else if (aCipher == "TLS_DHE_RSA_WITH_3DES_EDE_CBC_SHA") {
             level = "low";
           }
         }
--- a/palemoon/base/content/padlock.js
+++ b/palemoon/base/content/padlock.js
@@ -21,36 +21,15 @@ var padlock_PadLock =
     const wpl = Ci.nsIWebProgressListener;
     const wpl_security_bits = wpl.STATE_IS_SECURE |
                               wpl.STATE_IS_BROKEN |
-                              wpl.STATE_IS_INSECURE |
-                              wpl.STATE_LOADED_MIXED_ACTIVE_CONTENT |
-                              wpl.STATE_LOADED_MIXED_DISPLAY_CONTENT |
-                              wpl.STATE_IDENTITY_EV_TOPLEVEL;
+                              wpl.STATE_IS_INSECURE;
     var level;
     var highlight_urlbar = false;
 
     switch (aState & wpl_security_bits) {
-      case wpl.STATE_IS_SECURE | wpl.STATE_IDENTITY_EV_TOPLEVEL:
-        level = "ev";
-        highlight_urlbar = true;
-        break;
       case wpl.STATE_IS_SECURE:
-      case wpl.STATE_IS_BROKEN |
-           wpl.STATE_LOADED_MIXED_DISPLAY_CONTENT:
         level = "high";
         highlight_urlbar = true;
         break;
-      case wpl.STATE_IS_BROKEN |
-           wpl.STATE_LOADED_MIXED_ACTIVE_CONTENT:
-        level = "low";
-        highlight_urlbar = true;
-        break;
-      case wpl.STATE_IS_BROKEN | wpl.STATE_IDENTITY_EV_TOPLEVEL |
-           wpl.STATE_LOADED_MIXED_ACTIVE_CONTENT |
-           wpl.STATE_LOADED_MIXED_DISPLAY_CONTENT:
-      case wpl.STATE_IS_BROKEN | wpl.STATE_IDENTITY_EV_TOPLEVEL |
-           wpl.STATE_LOADED_MIXED_ACTIVE_CONTENT:
-      case wpl.STATE_IS_BROKEN | wpl.STATE_IDENTITY_EV_TOPLEVEL |
-           wpl.STATE_LOADED_MIXED_DISPLAY_CONTENT:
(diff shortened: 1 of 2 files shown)


wolfbeast:
> Gotta wonder, should we bring Medium back just for 1.1?

Considering the low adoption rate of TLS 1.1, no.

> DHE-RSA-AES ciphers marked as Low

And where have I told you to go ahead and decide that this is supposed to be the case? Has this been discussed? Don't log-roll things. Just don't; it will make me respect you a hell of a lot less.
DHE-RSA key exchange may be slow, but there's nothing bad about it. We already block too small DH keys with an error.

palemoon/base/content/padlock.js
```diff
@@ -22,50 +22,84 @@ var padlock_PadLock =
...
+          } else if (aCipher.indexOf("_RC4_") > -1) {
+            if (aCipher.indexOf("_MD5") > -1) {
+              level = "broken";
+            } else if (aCipher.indexOf("_SHA") > -1) {
+              level = "low";
+            }
+          } else if (aCipher.indexOf("TLS_DHE_RSA_WITH_AES") > -1) {
+            level = "low";
+          } else if (aCipher.indexOf("TLS_RSA_WITH_AES_128_") > -1) {
+            level = "low";
```
wolfbeast, line 90: Log-rolled undiscussed restriction.

palemoon/base/content/padlock.js
```diff
@@ -22,50 +22,84 @@ var padlock_PadLock =
...
+          } else if (aCipher.indexOf("_3DES_") > -1) {
+            level = "low";
```
wolfbeast, line 92: This should be a check for `_DES_EDE3_` IIRC (3DES = Encrypt-Decrypt-Encrypt DES with 3 keys)

RealityRipple:
> > DHE-RSA-AES ciphers marked as Low
> 
> And where have I told you to go ahead and decide that this is supposed to be the case? Has this been discussed? Don't log-roll things. Just don't; it will make me respect you a hell of a lot less.
> DHE-RSA key exchange may be slow, but there's nothing bad about it. We already block too small DH keys with an error. In fact, if you want to mark something in key exchanges low integrity, then it should be straight up RSA key exchange, but even _that_ should be _discussed_ first.

Shit, that was supposed to be a specific check for TLS_RSA_WITH_AES_128_GCM_SHA256. That is the one listed in PMC's Deprecated/Weak/Broken list, isn't it? Sorry, I really should have slept before I wrote this.

palemoon/base/content/padlock.js
RealityRipple: Oh, yeah, sorry. `_3DES_` finds ciphers "WITH_3DES_EDE_CBC". Bad read on my part.

RealityRipple:
Should the security level tooltips down in the `setPadlockLevel` function be localized? I noticed they're hardcoded in English in this JS.

palemoon/base/content/padlock.js
RealityRipple: Here's the list of 3DES-related ciphers and which protocols use them as far as I can tell:
- `SSL_EN_DES_192_EDE3_CBC_WITH_MD5` **SSL 2**
- `SSL_CK_DES_192_EDE3_CBC_WITH_MD5` **SSL 3**
- `SSL_RSA_FIPS_WITH_3DES_EDE_CBC_SHA` **SSL 3**
- `SSL_RSA_OLDFIPS_WITH_3DES_EDE_CBC_SHA` **SSL 3**
- `TLS_RSA_WITH_3DES_EDE_CBC_SHA` **SSL 3**/**TLS**
- `TLS_DH_DSS_WITH_3DES_EDE_CBC_SHA` **SSL 3**/**TLS**
- `

Note that the ones that start with `TLS_` but are **SSL 3*** compatible have equal consts starting with `SSL_` instead. If the cipherSuite string happens to return `SSL_` for SSL protocols still, we'll definitely want to account for that. If we end up going back to checking for `_3DES_`, though, this potential issue gets swept under the rug.

wolfbeast:
Let's please move in-depth discussion about decisions to be made to the issue, and not in the PR.

wolfbeast (approved):
Looks good as far as I can tell.

PR #1838, opened 2020-09-17
Fix vendor name in about dialog for unofficial builds.

adeshkp:
For some reason name is picked from `branding/unofficial/locales/en-US/brand.dtd` and not `branding/unofficial/locales/en-US/brand.properties` but both of them have `vendorShortName` defined.

How it looks currently:
Screenshot 2020-09-17 at 6 16 08 PM


wolfbeast:
I kind of like the "." ;-)
But sure.

